PR #253, opened 2014-12-16
`searchControl` has changed to `search` in createVis options hash

pichot:
Searchbox options argument appears to have changed. Also appears to default to 'true'.


fdansv:
Sorry this took so long, merged, thanks!

PR #277, opened 2015-01-16
named map tutorial example

ohasselblad:
Files for named maps tutorial:
+ template used
+ live example 

Both need to be accessible through a link like:  http://cartodb.github.io/cartodb.js/examples/tutorials/named-maps-example.html


matallo:
LGTM

PR #280, opened 2015-01-21
fix not showing time slider if no steps

Xatpy:
Tests added.
Also, in the example given in the issue, the time steps are: 0.5
Because of that, the condition is greather than 1 (instead of !== 1)

Please, @javisantana , could you review it?


javisantana:
update with develop

Xatpy:
I close this PR because I'll continue in this other [PR#283](https://github.com/CartoDB/cartodb.js/pull/283) which is updated from develop.

PR #283, opened 2015-01-22
Fix not showing time slider if steps aren't > 1

Xatpy:
From an updated branch (develop). (It came from PR 280)

Tests added.
Also, in the example given in the issue, the time steps are: 0.5
Because of that, the condition is greather than 1 (instead of !== 1)


Xatpy:
Please @javisantana , could you check it?
I 'started' again from an updated branch (develop)

javisantana:
good, merging

PR #297, opened 2015-01-28
Updates carto and torque for heatmap functionality

fdansv:
@javisantana 
Needs fixes, merge when:

* [x] Marker size for marker-files is correct (https://github.com/CartoDB/torque/issues/106)
* [x] Map style invalidates when attributes change in wizard.


javisantana:
:es:

PR #298, opened 2015-01-29
docfix: pos and latlng order were switched in event callbacks in examples

wilsaj:
Looks like the correct order is (e, latlong, pos, ...): http://docs.cartodb.com/cartodb-platform/cartodb-js.html#layerfeatureoverevent-latlng-pos-data-layerindex


javisantana:
thanks

PR #302, opened 2015-01-30
Fixes loader removal

javierarce:
@javisantana: can you review this?


javisantana:
I don't like repeating code (and bindings in the layers) but let's keep as is and think about how to move refactor mobile and Vis classes

PR #304, opened 2015-02-02
Added missing entries to vizjson spec

ohasselblad:
Fixed typo, added missing entries to named map portion


javisantana:
thnks!!

PR #308, opened 2015-02-02
Adds latest fixes from torque

fdansv:
This updates cartodb.js with:

- Problem with the icon preloader's counter
- Smarter way of correcting out of bounds coordinates

cc @javisantana


javisantana:
:es:

PR #310, opened 2015-02-03
Defines the API for the static maps

javierarce:
This branch adds a new object (`cdb.Image`) defined in [src/core/image.js](https://github.com/CartoDB/cartodb.js/blob/259-medusa/src/core/image.js) that uses the new Static Api to generate images from visualizations.

## Changed/added files:

- [src/cartodb.js](https://github.com/CartoDB/cartodb.js/blob/259-medusa/src/cartodb.js)
- [src/core/image.js](https://github.com/CartoDB/cartodb.js/blob/259-medusa/src/core/image.js)
- [core/image.spec.js](https://github.com/CartoDB/cartodb.js/blob/259-medusa/test/spec/core/image.spec.js)
- [examples/image.html](https://github.com/CartoDB/cartodb.js/blob/259-medusa/examples/image.html)


src/core/image.js
```diff
@@ -0,0 +1,539 @@
...
+
+    this.supported_formats = ["png", "jpg"];
+
+    this.defaults = {
+      tiler_domain: "cartodb.com",
+      tiler_port: "80"
+    };
+
+    this.available_basemaps = ["light_all", "light_nolabels", "dark_all", "dark_nolabels"];
+    this.options = _.defaults({
```
javisantana: this constructor is almost a copy for Map constructor, shouldn't we call Map constructor instead?

src/core/image.js
```diff
@@ -0,0 +1,539 @@
...
+
+      var ld = {
+        layers: layers
+      };
+
+      return ld;
+
+    },
+
+    _requestPOST: function(params, callback) {
```
javisantana: this should not be reimplemented here

src/core/image.js
```diff
@@ -0,0 +1,539 @@
...
+          callback(data);
+        },
+        error: function(xhr) {
+          callback(null);
+        }
+      });
+
+    },
+
+    _tilerHost: function() {
```
javisantana: should use Map one

src/core/image.js
```diff
@@ -0,0 +1,539 @@
...
+      var opts = this.options;
+      return opts.tiler_protocol +
+        "://" + ((opts.user_name) ? opts.user_name+".":"")  +
+      opts.tiler_domain +
+        ((opts.tiler_port != "") ? (":" + opts.tiler_port) : "");
+    },
+
+    _host: function(subhost) {
```
javisantana: should use Map one

src/core/image.js
```diff
@@ -0,0 +1,539 @@
...
+      var file = "cartodb.mod." + modName + (cartodb.DEBUG ? ".uncompressed.js" : ".js");
+      var src = this.getPath(file);
+      if (!src) {
+        cartodb.log.error("can't find cartodb.js file");
+      }
+      ImageLoader.loadScript(src);
+    }
+  };
+
+  window.imgjson = function(data) {
```
javisantana: this should not be duplicated from vizjson loader

src/core/image.js
javierarce: :ok_hand: removed

src/core/image.js
javierarce: :ok_hand: removed

src/core/image.js
javierarce: :ok_hand: I changed the constructor to use the Map one.

src/core/image.js
javierarce: :ok_hand: removed

src/core/image.js
javierarce: :ok_hand: we're now using the vizjson loader

javisantana:
:es:

PR #322, opened 2015-02-06
Image core

javierarce:
This branch:

- Removes the backbone dependency from the Image library.
- Extracts the Loader into its own file (cdb.vis.Loader → cdb.core.Loader)
- Adds a new example that only uses cartodb.core.js
- Update specs.

Extra stuff:

- Allows to specify a custom basemap (see [example/images/basemap.html](https://github.com/CartoDB/cartodb.js/blob/fd15b4c597985f14a94b5936cb20a11962ad594e/examples/images/basemap.html))
- Adds support for Plain basemaps.
- Adds vis namespace to Loader: cdb.vis.Loader = cdb.core.Loader

/cc @javisantana

### DIFF
--- a/examples/images/basemap.html
+++ b/examples/images/basemap.html
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <title>Single image example | CartoDB.js</title>
+    <title>Basemap | CartoDB.js</title>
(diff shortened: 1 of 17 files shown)


javierarce:
:es: by @javisantana.

PR #328, opened 2015-02-10
It won't overwrite images in production when version is prerelease

xavijam:
It basically uses the same way we do with JS or CSS. It checks if the version is prerelease and in that case, minor files won't be uploaded.

Example:

http://libs.cartocdn.com.s3.amazonaws.com/cartodb.js/v5/5.0.0-beta/themes/img/dark_loader.gif
http://libs.cartocdn.com.s3.amazonaws.com/cartodb.js/v5/5.0/themes/img/dark_loader.gif

cc @CartoDB/frontend

### DIFF
--- a/grunt/tasks/s3.js
+++ b/grunt/tasks/s3.js
@@ -111,7 +111,7 @@ module.exports = {
         ]
       },
 
-      'png-dist': {
+      'png-bugfixing': {
         options: {
           overwrite: true,
           cache: false,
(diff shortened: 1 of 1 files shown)


xavijam:
Reminder: delete v5 folder from CartoDB.js.

javisantana:
I dont know what was the problem so can't check this will work

xavijam:
We were uploading png and gif files into two routes/folders.

1) .../v5/5.0/... (minor)
2) .../v5/5.0.0-beta/... (bugfixing)

In that case, dev files will overwrite production files in 5.0 folder. With that commit we will prevent that behaviour. If the version is "prerelease", it won't be uploaded for minor version.

xavijam:
Last call

xavijam:
Merging after 6 days.

PR #330, opened 2015-02-11
Add generated css to .gitignore

rafatower:
At least when building CartoDB with `grunt` this is a bit of a pain for switching branches and some other git operations.

@javierarce can you please review and merge? thanks!


javierarce:
:+1: I guess it's safe to ignore this file since we already have:

`themes/css/all.css`
`themes/css/cartodb.ie.css`

Issue #313, opened 2015-02-03
Pan/Zoom events don't work as appropriate on touch-screen Windows 8


iriberri:
From SB: 4390616

iriberri:
PS: Waiting for specific details about browsers/versions

Xatpy:
I can zoom and pan. But I'm having problems with clicks and infowindows. 

The problem with Internet Explorer in touch mobiles (tested with Nokia Lumia 635) is that the click event is being triggered twice. So, because of that: no infowindows is appearing. 
I've tested it with the following code in `cartodb.uncompressed.js : 11427`:
```
nativeHandler = function (element, fn, args) {
        return function (event) {
          event = fixEvent(event || ((this.ownerDocument || this.document || this).parentWindow || win).event, true)
          debugger
          if ([event].concat(args)[0].type === "click") {
            alert(nativeHandler --> click!!!');
          }
          return fn.apply(element, [event].concat(args))
        }
      }
```
I check the same code with Android 4.4 (bq aquaris 4.5) and iOS7 (iPhone 4S) with Chrome and Safari and worked ('click' event triggered only once)
I've recorded a video with it--> [Link](https://dl.dropboxusercontent.com/u/22131567/ie_click.mp4) (sorry, video was uploaded flipped)

Xatpy:
With [this PR](https://github.com/CartoDB/wax/pull/6) and [this](https://github.com/CartoDB/cartodb.js/pull/334) it should be resolved, except drag and infowindow with Internet Explorer 11 (explanation [here](https://github.com/CartoDB/wax/pull/6) )

Xatpy:
Fixed with **3.12.6** and SupportBee replied.

Videos about the issue:
- [Infowindow](https://dl.dropboxusercontent.com/u/22131567/VID_20150223_124757.mp4)
- [Pan / Zoom](https://dl.dropboxusercontent.com/u/22131567/VID_20150223_175931.mp4)

PR #334, opened 2015-02-11
Fix GMaps and IEMobile click event:

Xatpy:
Related ticket: https://github.com/CartoDB/cartodb.js/issues/313

Click event and its infowindows wasn’t being triggered with Internet Explore
Mobile and GMaps.
 It was working with Leaflet maps, and the reason was that in
leaflet_cartodb_leaflet.js file it has the ‘case:’ like in this commit.

Also, I've added a condition in the other *if* because it was returning some errors when I was debugging with MS Visual Studio. It's protecting if the properties clientX and clientY aren't defined in o.e.changedTouches[0]


Xatpy:
Please, @javisantana, could you check it?

PR #335, opened 2015-02-12
Grunt checks package version before running any task

xavijam:
Checking if:
- version is empty
- version is not semver valid.

cc @fdansv


fdansv:
:shipit:

PR #337, opened 2015-02-13
Allows to force the https protocol via param

javierarce:
We now support to force the `https` protocol via param: 

```var image = cartodb.Image(vizjson, { https: true }).size(400, 300);```

@javisantana or @CartoDB/frontend: can you review it, please?


javisantana:
:es:

PR #339, opened 2015-02-13
Fixes generating a static map with a plain layer 

javierarce:
We weren't getting the correct type from the basemap, and as a result the "plain static maps" (as in a map with one plain color as the basemap) weren't getting generated.

Could you please review it, @javisantana, @CartoDB/frontend?


xavijam:
:+1:

PR #347, opened 2015-02-16
Makes sure values named zero are not legended as NULL

fdansv:
Ref. CartoDB/cartodb#1944

Basically the template checks for falsy values for name when making the legends' html. As 0 is a falsy value (```0 == false``` is true, god bless javascript ;) ), is was adding its label as ```NULL```. This makes sure that non-string values (in this case, numbers) are turned into strings before passing them to the template (```"0" == false``` is false).

screen shot 2015-02-16 at 17 46 59

@xavijam CR please? :) 
cc @ohasselblad


javisantana:
test please :tmcw:

Issue #333, opened 2015-02-11
Offset bug on touch devices when map is position fixed


turban:
We're having trouble with the _findPos method in leaflet_cartodb_layergroup.js: 
https://github.com/CartoDB/cartodb.js/blob/develop/src/geo/leaflet/leaflet_cartodb_layergroup.js#L267

The bug occurs when we have our map in a position fixed div (we use this as pseudo fullscreen on our site). The position calculation fails on touch devices, and the popup is not shown (tested on Andorid and iPhone). 

Try this example on your mobile:
http://thematicmapping.org/playground/cartodb/bugs/fixedpos.html

Scroll down to the map and click the fullscreen button. You'll see that popups on the CartoDB layer are not showing, while the Leaflet marker popup works. Everything works if the map div is not position fixed. 

Related to this bug: https://github.com/CartoDB/cartodb.js/issues/114

turban:
This is a serious bug for our site - is it worked on?

fdansv:
Hi @turban, have you tried using the current version of Leaflet's fullscreen plugin, from Mapbox's CDN? Seems to work fine when I use it instead of what you currently have:

```
<script src='//api.tiles.mapbox.com/mapbox.js/plugins/leaflet-fullscreen/v0.0.2/Leaflet.fullscreen.min.js'></script>
<link href='//api.tiles.mapbox.com/mapbox.js/plugins/leaflet-fullscreen/v0.0.2/leaflet.fullscreen.css' rel='stylesheet' />
```

turban:
We're not using any of these plugins, our map don't work because of a bug in cartodb.js with maps in fixed position. It used to be a bug on all devices, but after https://github.com/CartoDB/cartodb.js/issues/114 was solved , it's now only on touch devices. The reason the mapbox plugin works is that they use the Fullscreen API, but we find it too obtrusive for our users. 

My example above is just a minimal test case - you get the same error without the fullscreen pluging and just a fixed positioned div.

Example maps on our site:
1. http://www.nrk.no/norge/sa-syk-er-din-kommune-1.12194283
2. http://www.nrk.no/norge/to-av-tre-politimestere-er-positive-til-det-nye-politikartet-1.12214918

In the first map the popup are shown but with an offset. In the second map the offset is far off the map.  We don't experience the same problems with Leaflet popups.

fdansv:
@turban thanks for clarifying that. This has now been fixed at #355

turban:
[...]

PR #355, opened 2015-02-18
Properly generates leaflet point for touch events

fdansv:
Ref. #333

Coordinates of the return value of ```findPos()``` were ```NaN``` when passing a touch event on a map in fixed position. Using x and y as a fallback when clientX/clientY are undefined solves this.

@xavijam


xavijam:
:+1:

javisantana:
cc @Xatpy maybe this is interesting for you

PR #356, opened 2015-02-19
Fixes tiler request when the vizjson contains a named map and a torque layer with a named map

javierarce:
We [check](https://github.com/CartoDB/cartodb.js/pull/356/files#diff-e14eff99e7c7b57af717eedd1f3cc9d2R153) if the vizjson contains a named map and torque layer with a named map inside. If that's the case, we don't send the torque layer in the request.

### DIFF
--- a/src/vis/image.js
+++ b/src/vis/image.js
@@ -276,11 +276,14 @@
 
       var layerDefinition = new LayerDefinition(layer_definition, layer_definition.options);
 
+      var query    = layerDefinition.options.query || "SELECT * FROM " + layerDefinition.options.table_name;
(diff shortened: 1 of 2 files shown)


javierarce:
retest this please

javierarce:
@CartoDB/frontend / @javisantana: can you please have a look at this PR, please?

javisantana:
I'd add a test, this is an edge case that cen be missed in the future pretty easily

javierarce:
@javisantana: spec added. The code was fine for you?

javisantana:
yes yes :es:

javierarce:
Thanks!

Issue #357, opened 2015-02-20
Time slider should show times with ranges higher than 24h


fdansv:
Continuing from https://github.com/CartoDB/cartodb/issues/2371

At the moment we're only showing times in the time slider if the time range is of less than 24 hours ([source](https://github.com/CartoDB/cartodb.js/blob/develop/src/geo/ui/time_slider.js#L116)). We should probably increase that to 72 hours since right now it seems a bit too short.

@javisantana 
@iriberri

PR #358, opened 2015-02-20
Allowing to show time in time slider when range is less than 72h

fdansv:
Ref #357 
@javisantana 
@iriberri


iriberri:
This would make me happy :-)

javisantana:
:us:

PR #360, opened 2015-02-20
Update source code link for Leaflet example

mnquintana:


javisantana:
hey, that link was ok, it points to the "easy" example. Why do you think is wrong?

mnquintana:
The easy example isn't adding a CartoDB layer to a pre-existing Leaflet map - it's using the built-in `createVis` method.

javisantana:
you are right, I was looking at the wrong link, thanks !

PR #363, opened 2015-02-23
Cdb-336_IE10 fails to load infowindow data thats already been viewed

Xatpy:
The explanation is in this [StackOverflow post](http://stackoverflow.com/questions/779379/why-is-settimeoutfn-0-so
metimes-useful) : 
Recaping: (there's a more large explanation in that post)
> what setTimeout does is add a new event to the browser event queue and the rendering engine is already in that queue (not entirely true, but close enough) so it gets executed before the setTimeout event. . 

Also, it was 'hard' to reproduce. I was using Windows 8.1 and InternetExplorer 11 **emulating IE 10**, but it was working correctly. So, I had to use a virtual machine from [modern.ie](https://www.modern.ie/) with Win8/IE10. There, I could reproduce the problem, but it wasn't happening always. I think that the Internet connection and speed in events are related.

Also, I've modified  the test 'should call promise on error' to allow this fix (adding more time to setTimeout)

### DIFF
--- a/src/api/sql.js
+++ b/src/api/sql.js
@@ -155,9 +155,15 @@
         xhr = resp;
         resp = JSON.parse(resp.response);
       }
-      promise.trigger('done', resp, status, xhr);
-      if(success) success(resp, status, xhr);
-      if(callback) callback(resp);
+      //Timeout explanation. CartoDB.js ticket #336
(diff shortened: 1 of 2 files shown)


javisantana:
:es:

Issue #276, opened 2015-01-15
Support for retina tiles


rochoa:
Add support to retrieve retina tiles when it makes sense.

rochoa:
@javisantana I assigned it to you as I don't know who is gonna take care of it.

rochoa:
Check: https://github.com/CartoDB/Windshaft/blob/master/doc/Multilayer-API.md#retina-support

saleiva:
hey @javisantana can we think if we should prioritize this?

javisantana:
this issue depends more on api side and there is a bunch of things pending so I'd say let's finish started things first

rochoa:
Backend issue with pending work: https://github.com/CartoDB/Windshaft/issues/243

makella:
+1!

nobuti:
There has been no activity on this issue for several months and that's why are closing it (automatically). If you think this still needs to be addressed please reopen it. The good news is that we are now working on the next version of CARTO.js, stay tuned.

PR #364, opened 2015-02-23
Checks if screen is retina or superior, in which case adds @2x 

fdansv:
Ref. #276 

@rochoa

### DIFF
--- a/src/vis/vis.js
+++ b/src/vis/vis.js
@@ -319,6 +319,15 @@ var Vis = cdb.core.View.extend({
       this.https = true;
     }
 
+    // this checks that the screen is retina to add @2x to the url
+    if (window.devicePixelRatio > 1) {
+      data.layers.forEach(function(layer){
+        if(layer.type === "tiled"){
+          layer.options.urlTemplate = layer.options.urlTemplate.replace("{y}", "{y}@2x");
+        }
+      })
+    }
+
     if (data.https) {
       this.https = data.https;
     }


fdansv:
Will add a test so don't merge just yet, just CR ;)

src/vis/vis.js
```diff
@@ -319,6 +319,15 @@ var Vis = cdb.core.View.extend({
       this.https = true;
     }
 
+    // this checks that the screen is retina to add @2x to the url
+    if (window.devicePixelRatio > 1) {
+      data.layers.forEach(function(layer){
+        if(layer.type === "tiled"){
```
rochoa: I see this is only for ***basemaps***, but ... do all our supported basemaps follow that pattern?

We use that in our basemaps but I'm not sure if other providers follow that approach: Here, NASA and the like.

fdansv: @rochoa Yeah, you're right. This makes me think... is this something we want to add automatically, or let the user activate it with options? Many tile servers won't support an @2x flag.

Leaflet has a ```detectRetina``` option. We can add a similar one to our options and then apply this check.

Regarding the pattern, we can use a regex to add the flag just before the extension dot.

rochoa: My proposal is having different URLs for *normal* and retina resolution, if we detect we are in a retina display and we have a retina resolution URL for the basemap then we use it, otherwise fallback to the *normal* URL.

javisantana:
the way to do this is to have a retina flag for our layers so if it's enabled we use @2x for tiles we generate (mainly torque, png and grid)

We never can do that replacement on layers because is not standard. 

@rochoa solution is a good compromise I think, we do right now something like that for https:

https://github.com/CartoDB/cartodb.js/blob/develop/src/vis/layers.js#L11

saleiva:
bump!

src/vis/vis.js
javisantana: the place should be this: https://github.com/CartoDB/cartodb.js/blob/develop/src/vis/layers.js#L39

PR #368, opened 2015-02-23
Remove cdbui bower dependency

knownasilya:
The repo no longer exists..


javisantana:
@matallo could you review and merge?

matallo:
actually it is being used for CartoDB.js home page https://github.com/CartoDB/cartodb.js/blob/a3b001f6513e6fc93be6582c5886dae0f3d7c556/www/index.html#L33 but that's another discussion

no problem to merge

PR #369, opened 2015-02-23
Uses the CDN to retrieve the images

javierarce:
This PR:

- Enables the CDN.
- Allows to use the `no_cdn` param (see specs)
- Updates specs.
- Adds a couple of new specs to test the `no_cdn` param.


javisantana:
:es:

PR #370, opened 2015-02-23
Don't assume mapType is defined

knownasilya:
When using Google Map's `map.overlayMapTypes.setAt(index, mapType)`, the in betweens are set to `undefined`, which causes an error.


knownasilya:
Wrong branch...

knownasilya:
Actually, since 7.0.2 of wax doesn't even have this code, will this change even be merged?

javisantana:
we are using our own version of wax https://github.com/cartodb/wax so maybe it's still there

knownasilya:
No, that's where I looked.

Issue #371, opened 2015-02-23
Image documentation inside api.md


javisantana:
We lack of cartodb.js documentation for static maps api in api.md, it's here:

https://github.com/CartoDB/cartodb.js/wiki/CartoDB-Map-API

javierarce:
And if you need help clarifying something from that page, feel free to ask me, @ohasselblad.

javierarce:
Moved to the docs repo.

javierarce:
bump :dog:

ohasselblad:
on it! will try to get it done this week

PR #373, opened 2015-02-24
Allows to override the bbox

javierarce:
This PR adds a new user option (`override_bbox`) that allows to override the default use of the bounding box to generate the image, using the center instead.

Plus: [wiki updated](https://github.com/CartoDB/cartodb.js/wiki/CartoDB-Map-API#override_bbox).

@CartoDB/frontend / @javisantana: review this, please :)

### DIFF
--- a/src/vis/image.js
+++ b/src/vis/image.js
@@ -366,7 +366,7 @@
 
       var url = this._host() + this.endPoint;
 
-      if (bbox && bbox.length) {
+      if (bbox && bbox.length && !this.userOptions.override_bbox) {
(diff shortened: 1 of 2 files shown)


viddo:
:+1:

javierarce:
Thanks!

javisantana:
as a side comment I would store test viz.json locally, we had problems with that in the past

saleiva:
Should we update the docs too?

javierarce:
Yeah, I did that too, look at the description :)

saleiva:
Description mentions wiki page not official docs

javierarce:
Ah, ok, there's a ticket for that: https://github.com/CartoDB/cartodb.js/issues/371

PR #376, opened 2015-02-24
Don't assume mapType is defined

knownasilya:
Because when using `overlayMapTypes.setAt(index, mapType)`, gmaps inserts `undefined` in the in-betweens.

_Note: The current version of wax under the cartodb org doesn't even have this code, but who knows when that will be used here, so this is the fix in the short term._


javisantana:
thanks!

PR #378, opened 2015-02-24
Extract the CDN host configuration from the vizjson

javierarce:
Right now we are using the default one, which works in production, but we need to use the one specified in the the vizjson so it works locally and in staging.

### DIFF
--- a/test/spec/core/image.spec.js
+++ b/test/spec/core/image.spec.js
@@ -70,6 +70,20 @@ describe("Image", function() {
 
   });
 
+  it("should extract the cdn_url from the vizjson", function(done) {
+
+    var vizjson = "http://documentation.cartodb.com/api/v2/viz/e7b04b62-b901-11e4-b0d7-0e018d66dc29/viz.json";
+
+    var image = cartodb.Image(vizjson);
+
+    image.getUrl(function(err, url) {
+      expect(image.options.cdn_url.http).toEqual("api.cartocdn.com");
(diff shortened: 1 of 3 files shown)


javierarce:
@CartoDB/frontend / @javisantana: would you mind having a look, please?

javierarce:
@CartoDB/frontend: would you mind having a look, please?

test/spec/core/image.spec.js
```diff
@@ -70,6 +70,20 @@ describe("Image", function() {
...
   });
 
+  it("should extract the cdn_url from the vizjson", function(done) {
+
+    var vizjson = "http://documentation.cartodb.com/api/v2/viz/e7b04b62-b901-11e4-b0d7-0e018d66dc29/viz.json";
+
+    var image = cartodb.Image(vizjson);
+
+    image.getUrl(function(err, url) {
+      expect(image.options.cdn_url.http).toEqual("api.cartocdn.com");
```
xavijam: If the cdn url is different? is it customizable?

javierarce: That's not the objective of this ticket, but I'll create a new one to allow using a custom CDN :)

javisantana:
we use the cdb_url is set in the layergroup call, we never use viz.json one. Why we decided to do this?

javierarce:
Because it wasn't working and I thought I had to explicitly set it using the one coming from the viz.json.

javierarce:
I'm creating a ticket to fix this.

PR #379, opened 2015-02-24
Don't use the hidden layers to generate images

javierarce:
This PR fixes a problem where images were being generated using all the layers. We now only use the visible ones.


javierarce:
@CartoDB/frontend: can you look at this, please?

javierarce:
Thanks, @xavijam.

PR #385, opened 2015-02-25
Uses the CDN from the layegroup

javierarce:
- Updates specs.
- Replaces `_requestPOST`  with `getLayerToken`.


javisantana:
:us:

PR #389, opened 2015-02-26
Don't send the urlTemplate if we don't have it

javierarce:
This will prevent an error generating images for Google Maps, for example. /cc @rochoa 

@CartoDB/frontend / @javisantana: can you :eyes: :pray:?

### DIFF
--- a/src/vis/image.js
+++ b/src/vis/image.js
@@ -175,7 +175,12 @@
 
         }
 
-        var layers = [ this._getBasemapLayer() ];
+        var layers = [];
+        var basemap = this._getBasemapLayer();
+
+        if (basemap) {
+          layers.push(basemap);
+        }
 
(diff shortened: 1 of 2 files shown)


javisantana:
sounds ok, works with plain urls (maybe it has nothing to do)

rochoa:
@javisantana plain urls? I don't follow you.

The [urlTemplate is a required option for http layers](https://github.com/CartoDB/Windshaft/blob/master/doc/MapConfig-1.3.0.md#23-http-layers-options) so not sending it means it will fail.

javisantana:
sorry, it was a question about plain layers, I don't know if the code manages plain layers as basemaps.

Looks ok anyways

javierarce:
Thanks! I'll merge as soon as the build turns green.

PR #392, opened 2015-02-27
Disables the CDN if the server doesn't send us the configuration

javierarce:
Related to: https://github.com/CartoDB/cartodb-onpremise/issues/180

In some cases, the server could send an empty CDN config:

screen shot 2015-02-27 at 11 27 46

In this case, we should set the `no_cdn` flag to `false`, to avoid an [error in layer_definition.js](https://github.com/CartoDB/cartodb.js/blob/develop/src/geo/layer_definition.js#L611)

### DIFF
--- a/src/geo/layer_definition.js
+++ b/src/geo/layer_definition.js
@@ -598,18 +598,26 @@ Map.prototype = {
   },
 
   _host: function(subhost) {
+
     var opts = this.options;
-    if (opts.no_cdn) {
+    var has_empty_cdn = opts.cdn_url && (!opts.cdn_url.http && !opts.cdn_url.https);
+
+    if (opts.no_cdn || has_empty_cdn) {
       return this._tilerHost();
     } else {
+
       var h = opts.tiler_protocol + "://";
+
       if (subhost) {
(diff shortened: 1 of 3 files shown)


javierarce:
@javisantana: can you review it, please?

javisantana:
In this cases I think the behavior is to use the default one. That should be managed in layergroup level

javierarce:
I'm afraid not: 

https://github.com/CartoDB/cartodb.js/blob/develop/src/geo/layer_definition.js#L609

The problem is that, without changing anything, if `cdn_url = { http: "", https: "" }` evaluates to `true`. And if I send a falsy value, then the layerdefinition will use the default `CDN_URL`.

javisantana:
Ok, but then let's fix it at that line in layergroup so is fixed not only for images

javierarce:
Allright!

javierarce:
Moved logic to the layer definition file and added spec. Ready for review.

javisantana:
nice :es:

javierarce:
Thanks! I'll release it on Monday an coordinate with @santisaez.

PR #394, opened 2015-03-04
Stop adding a new image in infowindow-with-image-header template [do not merge]

xavijam:
We have found several problems here:

- We shouldn't get first field value as url. Value should be the one that appears as src in the image tag.
- We have to stop adding magically a new  element within the infowindow element, it causes problems with custom templates.

cc @javisantana @javierarce

### DIFF
--- a/test/spec/geo/infowindow.spec.js
+++ b/test/spec/geo/infowindow.spec.js
@@ -439,4 +439,6 @@ describe("cdb.geo.ui.infowindow", function() {
       model.set("content", { fields: fields });
       model.set('template', '<div class="cover"><img src="http://hello.png" /></div>');
       expect(view.$(".cover img").attr('style')).toBe(undefined);
-   
+    });
+  });
+});
--- a/test/spec/geo/infowindow.spec.js
+++ b/test/spec/geo/infowindow.spec.js
@@ -435,7 +435,7 @@ describe("cdb.geo.ui.infowindow", function() {
       expect(view.$(".cover img").attr('style')).toBe(undefined);
     });
 
-    it("if the theme doesn't have cover don't edit the image", function() {
+    it("if the theme doesn't have a cover don't style the image", function() {
       model.set("content", { fields: fields });
       model.set('template', '<div class="cover"><img src="http://hello.png" /></div>');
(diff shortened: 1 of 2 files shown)


javisantana:
is this change backwards compatible?

xavijam:
We have to test it in staging as we talked yesterday

javisantana:
backwards compatible with the cartodb.js API i meant

test/spec/geo/infowindow.spec.js
```diff
@@ -425,23 +425,18 @@ describe("cdb.geo.ui.infowindow", function() {
...
-      model.set("content", { fields: fieldsWithoutURL });
-      model.set('template', '<div class="cartodb-popup header" data-cover="true"><div class="cover"></div></div>');
-      expect(view.$el.find("img").length).toEqual(0);
+    it("should not edit <img> if it is not valid", function() {
+      model.set('template', '<div class="cartodb-popup header" data-cover="true"><div class="cover"><img src="hello"/></div></div>');
+      expect(view.$(".cover img").attr('style')).toBe(undefined);
     });
 
-    it("if the theme doesn't have cover don't append the image", function() {
+    it("if the theme doesn't have cover don't edit the image", function() {
```
javierarce: What does "don't edit" the image mean in this spec? I'm unsure of what this spec is trying to test.

xavijam: I can't change the name of the function, due to backward compatibility, but we don't add any  element anymore, and in fact, the function only loads and "positionates" correctly the image, that means it edits the style of the image if it is present, of course.

xavijam:
@javisantana, I have to make more tests, but yes, it does not change or remove any method, so "API" is not affected. It only changes the behaviour of the ```_loadCover```. In any case, as I said, I'm still testing.

javisantana:
I'm affraid of people with custom templates (not custom from the editor, i mean real handmade)

test/spec/geo/infowindow.spec.js
javierarce: Ah, ok. So why don't we say instead: "if the theme doesn't have a cover don't style the image"?

xavijam: :+1:

PR #396, opened 2015-03-04
Removes default CDN host

javierarce:
We'll now trust the tiler and use whatever CDN configuration we get (or nothing, if `cdn_url` is empty).


javisantana:
sounds good!

javierarce:
Thanks!

Issue #397, opened 2015-03-11
Torque layer with named map must request the named map


rochoa:
## STR
1. Create a points private table
1. Create a visualisation from it
1. Use a torque wizard

## Current result
The viz.json does not return a `namedmap` layer but only a torque layer with an associated named map.
cartodb.js tries to instantiate a layergroup through the anonymous endpoint with the torque layer.

## Expected result
As the torque layer has an associated namedmap cartodb.js should instantiate the layergroup with its associated named map instead of using the anonymous way.

Ref: https://github.com/CartoDB/cartodb.js/blob/develop/src/vis/image.js#L162

See https://github.com/CartoDB/cartodb/issues/2705

rochoa:
Thanks @javisantana and @Kartones for helping me understanding better the viz.json :+1:

rochoa:
cc @saleiva

saleiva:
@javierarce this is important since is one of the things that are more visible on the new dashboard.

javierarce:
I know, I'm waiting for a review of this: https://github.com/CartoDB/cartodb.js/pull/398

saleiva:
ah ok!

javierarce:
I've released a new version of cartodb.js with the fix. I'll update CartoDB so the change appears in the editor.

rochoa:
:heart_eyes:

PR #398, opened 2015-03-11
Torque layer with named map must request the named map 

javierarce:
Please, see the description of the original issue: https://github.com/CartoDB/cartodb.js/issues/397

Basically, in the cases where we have a vizjson of a torque with a named map inside (= a torque that was created with a private table, for example) we need to use the named map to do the request.

### DIFF
--- a/src/vis/image.js
+++ b/src/vis/image.js
@@ -186,11 +186,11 @@
 
           var layer = data.layers[i];
 
-          if (layer.type === "torque" && !ignoreTorqueLayer) {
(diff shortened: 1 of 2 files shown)


javierarce:
@CartoDB/frontend: can you review this pull request, please?

javisantana:
sounds good but take into account this is a hack. The correct way to do this would be to do it inside _getTorqueLayerDefinition and that returns the named map inside torque layer.

I'd add that as a comment

javierarce:
@javisantana: you're totally right. I've refactored the code instead. Have a look again, please.

javisantana:
:es:

PR #399, opened 2015-03-12
Upgraded mustache.js to v1.1.0

alonsogarciapablo:
@CartoDB/frontend., this PR bumps the version of mustache.js from v0.5.2 to v1.1.0.

We (@javisantana and I) discovered a bug in mustache.js when investigating cartodb/cartodb#2420 and this bump fixes it.

PTAL @javisantana, @xavijam

### DIFF
--- a/src/core/template.js
+++ b/src/core/template.js
@@ -67,7 +67,15 @@ cdb.core.Template = Backbone.Model.extend({
 }, {
   compilers: {
     'underscore': _.template,
-    'mustache': typeof(Mustache) === 'undefined' ? null: Mustache.compile
+    'mustache': typeof(Mustache) === 'undefined' ?
+      null :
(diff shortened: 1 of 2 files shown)


xavijam:
New mustache! :dancer:

xavijam:
We should check mustache templates with different mustache libraries, shouldn't we?

javisantana:
basically the things that can break are the infowindows but if you tried every template we have in cartodb we are pretty safe

alonsogarciapablo:
When I run `grunt dist`, the new version of mustache is not being included in the dist files :-(. It turns out, we are not including the file under `vendor/mustache.js`. Instead, we're relaying on the mustache.js that comes with `wax.cartodb.js` (v0.5.5).

Issue #361, opened 2015-02-20
Full screen button not working on Safari 8.0.3


iriberri:
I also tested in Version 8.0.2 (10600.2.5) over Mac OSX 10.10.1.

Notify: SB/4576792

javierarce:
Does it throw a JavaScript error?

Xatpy:
@javierarce No, it doesn't throw a Javascript error.

I've been researching, and it seems to be a **Bug with Safari**. 
References:
- [Stackoverflow I](http://stackoverflow.com/questions/8427413/webkitrequestfullscreen-fails-when-passing-element-allow-keyboard-input-in-safar) : *This is known Safari bug.* Sol-> **...(/Safari/.test(navigator.userAgent) ? undefined : Element.ALLOW_KEYBOARD_INPUT)**
- [Stackoverflow II](http://stackoverflow.com/questions/16150601/substitute-for-allow-keyboard-input-javascript-full-screen) : *Unfortunately, there isn't a good way to even do feature detection, since element.ALLOW_KEYBOARD_INPUT is defined in Safari, *
- [HTML5gamedevs](http://www.html5gamedevs.com/topic/2602-fullscreeninput-issue/) : *"For security reasons, most keyboard inputs have been blocked in the fullscreen mode. However, in Google Chrome you can request keyboard support by calling the method with a flag:
docElm.webkitRequestFullScreen(Element.ALLOW_KEYBOARD_INPUT);
This does not work in Safari, and the method won’t be called.*
- [Hack.mozilla.org](https://hacks.mozilla.org/2012/01/using-the-fullscreen-api-in-web-browsers/)

I've done a patch and I'm "pull requesting" it.

javierarce:
:ok_hand:

Xatpy:
merged.

PR #400, opened 2015-03-12
Fix fullscreen in Safari

Xatpy:
Ref. [ticket #361](https://github.com/CartoDB/cartodb.js/issues/361) : **Full screen button not working on Safari 8.0.3**

This seems to be a Safari bug. In the ticket I pointed several references.

And, I've written this patch using [this reference](http://stackoverflow.com/questions/8427413/webkitrequestfull
screen-fails-when-passing-element-allow-keyboard-input-in-safar)  --> **...(/Safari/.test(navigator.userAgent) ? undefined : Element.ALLOW_KEYBOARD_INPUT)**

Because of that, I used *undefined* as parameter.

I tested it with Safari 8.0 (10600.1.25), Chrome (41.0.2272.76) and Firefox (35.0.1) in MacOS, and Chrome (Android-4.4) and it works ok.

Please, @javierarce , could you check it? Thanks!

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,5 @@
+- Fixes a bug with fullscreen in Safari #361
+
 3.12.12 (12//03//2015)
 - Fixes a bug that prevented generating previews of torque layers with named maps
 
--- a/NEWS
+++ b/NEWS
@@ -1,4 +1,4 @@
-- Fixes a bug with fullscreen in Safari #361
+- Fixes a bug with fullscreen in Safari #361 
 
 3.12.12 (12//03//2015)
 - Fixes a bug that prevented generating previews of torque layers with named maps
--- a/src/geo/ui/fullscreen.js
+++ b/src/geo/ui/fullscreen.js
@@ -71,7 +71,10 @@ cdb.ui.common.FullScreen = cdb.core.View.extend({
     if (!doc.fullscreenElement && !doc.mozFullScreenElement && !doc.webkitFullscreenElement && !doc.msFullscreenElement) {
 
       if (docEl.webkitRequestFullScreen) {
-        requestFullScreen.call(docEl, Element.ALLOW_KEYBOARD_INPUT);
+        // Cartodb.js #361 :: Full screen button not working on Safari 8.0.3 #361
+        // Safari has a bug that fullScreen doestn't work with Element.ALLOW_KEYBOARD_INPUT);
+        // Reference: Ehttp://stackoverflow.com/questions/8427413/webkitrequestfullscreen-fails-when-passing-element-allow-keyboard-input-in-safar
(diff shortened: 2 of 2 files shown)


Xatpy:
Oh, what a pity! :confused: 
I've tested it, and I can type (in fullscreen) with Firefox and Chrome, but not with Safari.

Xatpy:
Although, related to this, I've tested that search box isn't working right now (with normal (no fullscreen) window too).
[Example](http://cartodb.github.io/cartodb.js/examples/easy.html) Searching: 'Barcelona'
I mean, it's throwing the query to nokia (see Network tab), but the map isn't being 'redraw'.

javierarce:
We are using cartodb.js 3.11.36 there, maybe you should update the example
to 3.12?

-- 
Javier Arce

Xatpy:
I tested it with 3.12 too (in local) , and it doesn't work :S

Xatpy:
@javierarce , I've been researching, and it seems that *Safari doesn't support keyboard input in fullscreen for security issues*.
References:
- [Caniuse.com](http://caniuse.com/#feat=fullscreen) : Known issues tab : `Safari blocks access to keyboard events in fullscreen mode (as a security measure).`
- [Screenfull.js](https://github.com/sindresorhus/screenfull.js/) : Safari doesn't support use of the keyboard in fullscreen. 

I tested again this fix, and input keyboard works in Firefox and Chrome.

javierarce:
LOL. Ok, well, green light to the changes :)

Xatpy:
The mystery of search is solved:

The problem was that an old [viz.json](http://team.cartodb.com/api/v2/viz/dbd18712-c972-11e4-8199-0e018d66dc29/viz.json) was being used in that [example](http://cartodb.github.io/cartodb.js/examples/easy.html). And that viz.json had a basemap with maxZoom : 10. This basemap was [removed time ago](https://github.com/CartoDB/cartodb/commit/a29d7cf8240d9807f6f097b9185e17621709004a#diff-5b8fbc0db4bda84e2a4891a453e8e328L109). 
But, the viz.json was being used, and it generated a problem when it was looking for the zoom.  I discovered it in in [this line](https://github.com/CartoDB/cartodb.js/blob/develop/src/geo/map.js#L540). Because if I used the search [with this method](https://github.com/CartoDB/cartodb.js/blob/develop/src/geo/ui/search.js#L70) (setting the zoom and center directly), it worked correctly (with zoom 13). The problem was that this maxZoom wasn't correct, because that basemap had more zoom levels (21, instead of 10).

I solved this problem changing the basemap in that vizjson. I've used the actual Positron. I tested searching with fullscreen too, and it worked. :D 

So @javierarce ,with your ok, I'll merge it. thanks! ^^

PR #401, opened 2015-03-13
Url rewrite

javisantana:
this is a pretty major change, not because it's a lot of code, it's because it changes the way we send urls from backend (both in cartodb and viz.json)

We are using right now configuration options tile "tiler_host", "tiler_protocol" and so on so we build the final url composing those urls frontend side. 

We changed to use url templates, basically the urls to reach a cartodb service are not composed from different attributes, they are a single attribute, like:

```
sql_api_template: 'http://{user}.cartodb.com/api/v2/sql
```

We accept a special ``{user}`` parameter, that's used to be able to compose those urls given an username (for example, from a javascript api call)

I added code to maintain backwards compatibility so tiler_* and sql_api_* are still valid.

You should also check this PR in cartodb which removes every tiler_* and sql_api_* from frontend code and start to use ``maps_api_template`` and ``sql_api_template`` https://github.com/CartoDB/cartodb/pull/1933

This is really important so everyone needs to review this: @viddo @xavijam @javierarce @Xatpy @matallo @alonsogarciapablo


viddo:
:+1: and thx for the thorough explanation

Xatpy:
Read :ok_hand:

matallo:
understood

javierarce:
The `Image.js` part looks fine :dog:

javisantana:
retest this please

xavijam:
:clap: :clap:

PR #403, opened 2015-03-13
Upgrade mustache.js and wax.cartodb.js

alonsogarciapablo:


alonsogarciapablo:
This is ready for review an will be merged as soon as we make sure that all tests pass (test suite is currently broken for other reasons).

@javisantana, can you please take a look?

javisantana:
:es:

this reminds me that at some point we should move file list from cartodb.js file to grunt build

PR #406, opened 2015-03-17
Handle null values in Infowindows and Tooltips

alonsogarciapablo:
This PR fixes cartodb/cartodb#2420.

I have modified `cdb.geo.ui.infowindow.contentForFields`, which is the function that generates the fields that will get rendered in tooltips and infowindows. After these changes, this function will only return empty fields, when `options.empty_fields` is set to true (something that only happens for tooltips in the editor).

I added some tests for it. 

@xavijam @javisantana PTAL

### DIFF
--- a/src/geo/ui/infowindow.js
+++ b/src/geo/ui/infowindow.js
@@ -185,13 +185,13 @@ cdb.geo.ui.InfowindowModel = Backbone.Model.extend({
     options = options || {};
     var render_fields = [];
     for(var j = 0; j < fields.length; ++j) {
-      var f = fields[j];
-      var value = String(attributes[f.name]);
-      if(options.empty_fields || (attributes[f.name] !== undefined && value != "")) {
+      var field = fields[j];
+      var value = attributes[field.name];
+      if(options.empty_fields || (value !== undefined && value !== null)) {
         render_fields.push({
-          title: f.title ? f.name : null,
-          value: attributes[f.name],
-          index: j ? j : null
+          title: field.title ? field.name : null,
+          value: attributes[field.name],
+          index: j
         });
       }
     }
@@ -201,7 +201,7 @@ cdb.geo.ui.InfowindowModel = Backbone.Model.extend({
       render_fields.push({
         title: null,
         value: 'No data available',
-        index: j ? j : null,
+        index: 0,
         type: 'empty'
       });
     }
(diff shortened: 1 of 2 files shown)


src/geo/ui/infowindow.js
```diff
@@ -185,13 +185,13 @@ cdb.geo.ui.InfowindowModel = Backbone.Model.extend({
     options = options || {};
     var render_fields = [];
     for(var j = 0; j < fields.length; ++j) {
-      var f = fields[j];
-      var value = String(attributes[f.name]);
```
xavijam: @javisantana, we have checked this parameter, but we don't know if there is something behind the shadows, do you have any idea about it?

src/geo/ui/infowindow.js
```diff
@@ -185,13 +185,13 @@ cdb.geo.ui.InfowindowModel = Backbone.Model.extend({
...
-      if(options.empty_fields || (attributes[f.name] !== undefined && value != "")) {
+      var field = fields[j];
+      var value = attributes[field.name];
+      if(options.empty_fields || (value !== undefined && value !== null)) {
         render_fields.push({
-          title: f.title ? f.name : null,
-          value: attributes[f.name],
-          index: j ? j : null
```
xavijam: Same for this @javisantana, any clue?

src/geo/ui/infowindow.js
javisantana: no idea, but If I did, i did it because of something. My guess is: as we used this code in cartodb.js and the editor and we wanted to render null values as "null" in the editor I did that hack.

I think is ok to remove it

src/geo/ui/infowindow.js
javisantana: yes, that's mustache problem, zero values were not managed properly so I needed to change the 0 index by a null. I guess the new mustache fixed this

xavijam: Yeah, I know this is on purpose, but i don't remember either.

src/geo/ui/infowindow.js
xavijam: Yeah, I know this is on purpose, but i don't remember either.

alonsogarciapablo:
Is that a :+1: guys?

xavijam:
yep :)

Issue #412, opened 2015-03-24
Fullscreen button is throwing errors


xavijam:
[js error](https://my.trackjs.com/details/9af2df365821487993272effc3c308c0)

```
_toggleFullScreen@http://cartodb-libs.global.ssl.fastly.net/cartodbui/assets/3.7.31/javascripts/cdb.js 18:28814  Source
[native code]
dispatch@http://cartodb-libs.global.ssl.fastly.net/cartodbui/assets/3.7.31/javascripts/cdb.js 2:9602  Source
h@http://cartodb-libs.global.ssl.fastly.net/cartodbui/assets/3.7.31/javascripts/cdb.js 2:5517  Source
http@//d2zah9y47r7bi2.cloudfront.net/releases/current/tracker.js 3:497
dispatchEvent@[native code]
_simulateEvent@http://cartodb-libs.global.ssl.fastly.net/cartodbui/assets/3.7.31/javascripts/cdb.js 10:14476  Source
_onUp@http://cartodb-libs.global.ssl.fastly.net/cartodbui/assets/3.7.31/javascripts/cdb.js 10:14054  Source
f@http://cartodb-libs.global.ssl.fastly.net/cartodbui/assets/3.7.31/javascripts/cdb.js 8:29305
```

xavijam:
@alonsogarciapablo is you can check it with him or in your own, perfect too :).

Xatpy:
Sorry guys, but I was ill yesterday and I couldn't check it.

@alonsogarciapablo did you check anything about it?

@xavijam , I've checked that those error is happening with iPhone (Safari Mobile / iOS 8.2). 

Also, could be [this fix](https://github.com/CartoDB/cartodb.js/pull/400) related? I did it last week, but I couldn't test it with iPhone.

alonsogarciapablo:
Hello Xatpy. I missed this yesterday too. Do you need some help?

Xatpy:
I've been trying to replicate it, but I couldn't. 

I've tried with Chrome 41.0.2272.101 / Safari 8.0.2 (10600.2.5) / Firefox (36.0.1) on OSX Yosemite 10.10.1; and Chrome/Android. Also, I did a quick test with iOS7/Safari, but it was quick because it was my girlfriends' phone and she has to go earlier.

@alonsogarciapablo Do you have an iOS 8? If so, could you test it?

[...]

PR #413, opened 2015-03-25
Cdb-412-Fullscreen button is throwing errors

Xatpy:
Ref. ticket #412 : https://github.com/CartoDB/cartodb.js/issues/412

Track.js was reporting that the error was throw with the following info:
- Browser: Mobile Safari 8.0
- Operating System: iOS 8.2

A fullscreen action was being executed from a mobile device, so the option *force_mobile=false* would be necessary (thanks @xavijam ).

Indeed, I tried it with desktop (OS X 10.10.1) browsers (Chrome/Fixefox/Safari) and it worked. But, I [emulated](http://www.macinstruct.com/node/494) an iOS 8.1 and I was able to reproduce the error.

So, the issue is:
**Nowadays (2015/03/25), fullscreen is not supported in iOS Safari. [Reference](http://caniuse.com/#feat=fullscreen)**

With this patch, I'm avoiding the error messages. Also, I think that it would be better if we don't allow to put on screen the fullscreen button in cases like that, but I think that it would be another (and more complex) issue. 

Please, @xavijam @alonsogarciapablo , could you check it? Thanks!

### DIFF
--- a/src/geo/ui/fullscreen.js
+++ b/src/geo/ui/fullscreen.js
@@ -76,8 +76,8 @@ cdb.ui.common.FullScreen = cdb.core.View.extend({
         // Reference: Ehttp://stackoverflow.com/questions/8427413/webkitrequestfullscreen-fails-when-passing-element-allow-keyboard-input-in-safar
         requestFullScreen.call(docEl, undefined);
       } else {
-        // CartoDB.js #412 : 
-        // Nowadays (2015/03/25), fullscreen is not supported in iOS Mobile. Reference: http://caniuse.com/#feat=fullscreen
+        // CartoDB.js #412 :: Fullscreen button is throwing errors
+        // Nowadays (2015/03/25), fullscreen is not supported in iOS Safari. Reference: http://caniuse.com/#feat=fullscreen
         if (requestFullScreen !== undefined) {
           requestFullScreen.call(docEl);
         }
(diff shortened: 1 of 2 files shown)


alonsogarciapablo:
I will take a look.

src/geo/ui/fullscreen.js
```diff
@@ -76,7 +76,11 @@ cdb.ui.common.FullScreen = cdb.core.View.extend({
         // Reference: Ehttp://stackoverflow.com/questions/8427413/webkitrequestfullscreen-fails-when-passing-element-allow-keyboard-input-in-safar
         requestFullScreen.call(docEl, undefined);
       } else {
-        requestFullScreen.call(docEl);
+        // CartoDB.js #412 :: Fullscreen button is throwing errors
+        // Nowadays (2015/03/25), fullscreen is not supported in iOS Safari. Reference: http://caniuse.com/#feat=fullscreen
+        if (requestFullScreen !== undefined) {
```
alonsogarciapablo: Not sure if, in JS, this is the same as:

```
if (requestFullScreen) {
  ...
}
```

alonsogarciapablo:
I haven't emulated iOS 8.1 but I think your solution is :+1: . I left a minor question/comment. Other than that, great job @Xatpy!

xavijam:
If you agree, for me perfect, pablo is the reviewer this time :).

Xatpy:
thanks @alonsogarciapablo I've updated the branch with your comment fixed.

alonsogarciapablo:
:+1:

PR #417, opened 2015-03-30
Updates torque with version 2.11

fdansv:
@xavijam 
News since last update:
```
    - Do not fix values in the edge (#147)
    - Windshaft provider accepts an optional data type for coordinates (#149)
    - Acceptance tests
    - renderTile accepts a callback to be called when rendering finishes
    - tile rendering deferred until all assets are loaded
    - sprite rendering now scales source to marker dimensions
    - Filters accept a canvas class
    - Adjusted point position taking resolution into account
    - Changed loop: false to pause animation at last frame
    - On torque-frame-count: 1, always pause for better performance
    - Fixes getTimeSpan
    - Adds timetostep method to torque layer
    - Function qualifyURL can be provided to not rely on document's one
```


xavijam:
Go go go !

PR #418, opened 2015-03-30
Reverts windows newline messup

fdansv:
Replaces CRLF to unix newline in three files after accidental conversion.

@viddo I think this should do it?


viddo:
LGTM :shipit:

Issue #404, opened 2015-03-13
do not use remote viz.json in tests


javisantana:
image tests need to fetch remote viz.json. I'd create some local fixtures with them

javisantana:
deplicated https://github.com/CartoDB/cartodb.js/issues/375

PR #420, opened 2015-03-30
Shows password protected visualizations

javierarce:
Password protected visualizations require to get an array with the `auth_tokens`. This PR solves that.

@javisantana: can you :eyes:, please?

### DIFF
--- a/test/spec/core/image.spec.js
+++ b/test/spec/core/image.spec.js
@@ -435,7 +435,7 @@ describe("Image", function() {
 
   it("should send the auth_tokens", function(done) {
 
-    var vizjson = "http://documentation.cartodb.com/api/v2/viz/8639778a-d6ea-11e4-861c-0e018d66dc29/viz.json"
+    var vizjson = "http://documentation.cartodb.com/api/v2/viz/e11db0aa-d77e-11e4-9039-0e853d047bba/viz.json"
 
     var image = cartodb.Image(vizjson).size(400, 300);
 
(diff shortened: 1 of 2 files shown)


javisantana:
looks good but I would not use viz.json online, that is something that will fail at some point :(

javierarce:
retest this please

javierarce:
You're right. I keep postponing a fix for that. 

I'm assigning myself this ticket: https://github.com/CartoDB/cartodb.js/issues/404

javierarce:
@javisantana: can you review [my last change](https://github.com/CartoDB/cartodb.js/commit/aee06ea9c1963228b99f1d495328b80066f69157), please?

test/spec/core/image.spec.js
```diff
@@ -433,4 +433,35 @@ describe("Image", function() {
...
+
+    StaticImage.prototype.load = function(vizjson, options) {
```
javisantana: why is this needed?

javierarce: To test the vizjson response without making a real call. Do you prefer to use other method?

javisantana: you could test using the layergroup version of contrcutor, right?

javierarce: Yes, that wouldn't test the regular method which has the majority of the code.

javierarce: bump

javisantana:
:es:

PR #424, opened 2015-04-01
Updates torque to torque 2.11.1

fdansv:
This update includes the method ```getValueFromBBox```, which we need for Torque infowindows.

cc @xavijam


javisantana:
I think we can wait to the new version with error fixed

PR #426, opened 2015-04-03
adding google heatmap tutorial code

michellechandra:


michellechandra:
@andrewxhill @javisantana 

Adding Google Heatmap tutorial code.

tutorial-google-heatmap.html - example code people will refer to for the tutorial
tutorial-google-heatmap-2.html - map code to be hosted on gh pages so can be included at the top of the tutorial page, uses a Google for Work account

javisantana:
the code looks good. Do we want to have examples of google heatmaps instead of using our heatmaps?

andrewxhill:
So, this ticket predated our own heatmaps. It cannot hurt though, but I think we should add a paragraph to the top saying that you can also do heatmaps with cartocss and that they will scale a lot better. 

let me take a look

javisantana:
it's ok.

FYI: using geojson format is less optimal than user json with "select st_x(the_geom) as lon, st_y(the_geom) as lat from table". For this case I'd not change it since geojson is a well known format

PR #428, opened 2015-04-06
Fix <a target="_blank" attribute

viddo:
Change the policy to allow target attribute


viddo:
@javisantana ok?

iriberri:
Subscribing to contact users

javisantana:
we try to not change third party libraries. If we do so, we fork in cartodb account so we can track upstream changes.

Otherwise when we update third party libraries we forget what we changed and introduce regressions.

viddo:
Fair enough, that makes sense. The only issue I see is that the code is currently residing on google code, is a SVN repo, and requires a built-step to generate the bundle file either way (it's not provided a pre-built/dist file). I'll look into it.

viddo:
The tool they provide to export to Github doesn't work, not sure what the best way would be to set this up. For now, I'd suggest we merge this as it is, to solve the issue the immediate bug for our users and since there are already changes to this 3rd party lib (as documented in the head of the file). Dealing with setting up a mirror repo and doing proper patches etc. can be done as a separate step.

PR #429, opened 2015-04-06
Sanitize infowindow output on render only

viddo:
While debugging the [issue with broken  that contained `&`s](https://github.com/CartoDB/cartodb/issues/3010) more thoroughly I found the real issue, infowindows are rendered differently in editor vs. using cartodb.js (why the images/URLs break only on embeds). 

The culprit is that [infowindow model's content fields are sanitized on update](https://github.com/CartoDB/cartodb.js/blob/develop/src/geo/ui/infowindow.js#L193), so these values will be double-escaped upon render (due to the default mustache template, i.e. `{{ img_url }}`). For the editor the code don't use this way of rendering so was just injected as-is.

With these changes the render is consistent for both the editor and cartodb.js usage by default, i.e. what the user sees in the editor is how it will look in the embed too.

### DIFF
--- a/src/geo/ui/infowindow.js
+++ b/src/geo/ui/infowindow.js
@@ -190,7 +190,7 @@ cdb.geo.ui.InfowindowModel = Backbone.Model.extend({
       if(options.empty_fields || (value !== undefined && value !== null)) {
         render_fields.push({
           title: field.title ? field.name : null,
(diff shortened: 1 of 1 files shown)


viddo:
@javisantana please review
cc @iriberri

javisantana:
:es:

PR #436, opened 2015-04-10
We don't allow to force the https anymore. Removed unnecesary specs

javierarce:
This PR removes the option to force the `https` protocol of a static map and fixes the specs accordingly (as in removing them completely)

### DIFF
--- a/src/vis/image.js
+++ b/src/vis/image.js
@@ -239,11 +239,6 @@
       this.options.tiler_protocol = protocol;
       this.options.tiler_port     = port;
 
-      if (this.imageOptions.vizjson.indexOf("https") === 0) {
-        this.options.tiler_protocol = "https";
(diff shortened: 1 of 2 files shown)


javierarce:

src/vis/image.js
```diff
@@ -239,7 +239,7 @@
       this.options.tiler_protocol = protocol;
       this.options.tiler_port     = port;
 
-      if (this.userOptions.https || this.imageOptions.vizjson.indexOf("https") === 0) {
```
javisantana: this is not needed since if the viz.json is requested by https the url it uses for the tiler is https (since 3.14 version)

javisantana:
:es:

src/vis/image.js
javierarce: Ok, I'll remove it. Thanks!

PR #440, opened 2015-04-15
Order legend items alphabetically

alonsogarciapablo:
Refs cartodb/cartodb#2808.

@javisantana PTAL.


javisantana:
I'm not sure this is the right place, cartodb.js should render whatever you tell to render, it should not be smart about that

alonsogarciapablo:
Closing this in favor of cartodb/cartodb#3218.

PR #447, opened 2015-04-20
Doesn't add timeslider with Torque static maps in mobile

fdansv:
Ref. #446 

Makes sure the timeslider is not templated in when the Torque map is static.

@ohasselblad @javisantana


javisantana:
sounds good but I'm afraid of this breaking because data.torqueLayer is changed. I think there are tests for this, right?

PR #448, opened 2015-04-21
Reorder infowindow layers

mbarrenechea:
This will allow us to reorder layers with infowindow in GFW


xavijam:
Could you change it first in https://github.com/CartoDB/wax, please :).

PR #450, opened 2015-04-21
Shows date ranges on sliders with few steps

fdansv:
Ref. https://github.com/CartoDB/torque/issues/135

...instead of concrete dates:

@javisantana @ohasselblad

### DIFF
--- a/test/spec/geo/ui/time_slider.spec.js
+++ b/test/spec/geo/ui/time_slider.spec.js
@@ -73,7 +73,7 @@ describe('cdb.geo.ui.TimeSlider', function() {
       });
       it("should return a formatter function that defines a two-date range", function() {
         var moment = new Date(1424269402400);
-        expect(formatter.call(view,moment)).toEqual("Feb 18 - Feb 20");
+        expect(formatter(moment, view.torqueLayer)).toEqual("Feb 18 - Feb 20");
       });
     });
 
--- a/test/spec/geo/ui/time_slider.spec.js
+++ b/test/spec/geo/ui/time_slider.spec.js
@@ -63,6 +63,20 @@ describe('cdb.geo.ui.TimeSlider', function() {
       });
     });
 
+    describe("given a step that spans more than 48 hours", function() {
+      beforeEach(function() {
+        var start = 1423699205000;
+        var end = 1424649534000;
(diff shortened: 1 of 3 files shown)


javisantana:
I think @viddo should review this since he was the last one in changing that :*

javisantana:
(check some failing tests and add some more please)

viddo:
Looks legit, +1 for adding new tests for the new use-cases to have it covered though.

fdansv:
Altered a check slightly for tests to pass, added one test for this case.

test/spec/geo/ui/time_slider.spec.js
```diff
@@ -63,6 +63,20 @@ describe('cdb.geo.ui.TimeSlider', function() {
       });
     });
 
+    describe("given a step that spans more than 48 hours", function() {
+      beforeEach(function() {
+        var start = 1423699205000;
+        var end = 1424649534000;
+        view.torqueLayer.getTimeBounds = function(){return {start: start, end: end}};
+        view.torqueLayer.options = {steps: 5};
+        formatter = view.formatterForRange.bind(view)(start, end);
```
viddo: why doing the bind here? since you're calling the `formatterFoRange` on the view object the `this` context will be the view

fdansv: Yes, sorry, that was a lapsus, I was adding it on L76 and confused both lines ;)

test/spec/geo/ui/time_slider.spec.js
```diff
@@ -63,6 +63,20 @@ describe('cdb.geo.ui.TimeSlider', function() {
...
+      });
+      it("should return a formatter function that defines a two-date range", function() {
+        var moment = new Date(1424269402400);
+        expect(formatter.bind(view)(moment)).toEqual("Feb 18 - Feb 20");
```
viddo: .bind is best used when you want to make a partial application of the function (predefined params), if you just want to call the function with a certain context it's better to use .call/.apply (depending on what kind of arguments you want to give it). So to make it clearer I'd recommend to just use:
``formatter.call(view, moment)`

fdansv: That's great to know, thanks, I'm still getting used to Javascript's context model. Will correct it now.

Issue #351, opened 2015-02-17
Static Map API: Allow to specify a step for torque layers 


javierarce:
It would be nice to have: 

* cdb.Image(`vizjson`, { step: 15 }).getUrl(…)
* cdb.Image(`vizjson`, { percentage: 50 }).getUrl(…)

Issue #439, opened 2015-04-15
image.bbox raises exception


javisantana:
in this line:

https://github.com/CartoDB/cartodb.js/blob/develop/src/vis/image.js#L415

when you use it with the API (not using vizjson), it fails. 
 
you can see here a workaround:

```
 var i = cartodb.Image(layer_definition)
      .size(width, height)
      i.userOptions = {}
      i.bbox([bounds[1][1], bounds[1][0], bounds[0][1], bounds[0][0]]) 
      .getUrl(function(error, url) {
       ...
      });
```

PR #451, opened 2015-04-22
Allows to specify a step for torque layers

javierarce:
This PR adds the possibility to specify a `step` param to render a specific frame of a torque layer.

`cartodb.Image(vizjson, { step: 100 }).size(500, 500)`

Original issue: https://github.com/CartoDB/cartodb.js/issues/351. Related to: https://github.com/CartoDB/cartodb/issues/3268

### DIFF
--- a/src/vis/image.js
+++ b/src/vis/image.js
@@ -348,6 +348,7 @@
       return {
         type: "torque",
         options: {
+          step: this.userOptions.step || 0,
           sql: query,
           cartocss: cartocss
         }
--- a/test/spec/core/image.spec.js
+++ b/test/spec/core/image.spec.js
@@ -86,6 +86,22 @@ describe("Image", function() {
 
   });
 
+  it("should allow to use a step for a torque layer", function(done) {
+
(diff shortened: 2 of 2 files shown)


javierarce:
@javisantana: can you review this, please?

javierarce:
@javisantana: can you review this, please?

javierarce:
@fdansv: can you have a look, please?

javisantana:
two comments:

- there is a bug related to userOptions and I think this would have the same problem: https://github.com/CartoDB/cartodb.js/issues/439
- why not to do the same interface that we have: cartodb.Image(vizjson).size(500, 500).step(10)

javierarce:
@javisantana: because all the chained methods are executed in a queue after the layergroup request is made and we need to send the `step` option before. 

As I see it, the chained methods control the general look of the image (zoom, center, format) + allows to render the image (into, getURL, etc.), while the params are used to control "internal" settings of the static map.

I'll take care of the issue you mentioned.

javierarce:
Ok, #439 is fixed :)

javisantana:
:es:

PR #453, opened 2015-04-22
Defines userOptions for the layer_definition loader

javierarce:
`userOptions` wasn't defined for the `loadLayerDefinition`, so in some cases an exception was triggered.

Fixes: https://github.com/CartoDB/cartodb.js/issues/439

### DIFF
--- a/src/vis/image.js
+++ b/src/vis/image.js
@@ -97,7 +97,7 @@
 
     },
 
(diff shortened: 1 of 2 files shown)


javierarce:

javisantana:
:es:

PR #454, opened 2015-04-23
Fix vis.addInfowindow() method name in doc/API.md

fullofcaffeine:
I was searching for the actual definition of ```vis.addInfoWindow``` in the cartodb.js source code after reading about it in the API docs but couldn't find anything with that name. Happens that the method itself is defined as ```addInfowindow``` and not ```addInfoWindow``` (notice the lower case w).


javisantana:
thansk!

Issue #455, opened 2015-04-24
Display date ranges in timeslider only if whole range is less than a year


fdansv:
The change we introduced the other day to the timeslider --showing date ranges when steps span more than a day-- should only affect visualizations that span less than a year. Otherwise visualizations that span several years become very confusing.

cc @javisantana, @ohasselblad

alonsogarciapablo:
@fdansv I think this issue can be closed, right?

fdansv:
Oh, indeed! Weird that I missed this the other day, when I was doing some issue cleanup.

PR #456, opened 2015-04-24
Only ranges dates if the total range is less than a year

fdansv:
Ref #455 

@javisantana


javisantana:
sounds right but I'd try to add some tests here, looks like is becoming kind of hard to follow logic

fdansv:
@javisantana added, all good?

javisantana:
fdescribe ? what does it do?

javisantana:
apart from that comment, all good

fdansv:
Ah damn, good that you mentioned it!!! That's just for Jasmine to perform only that test, instead of the whole suite. Shouldn't have committed that, eeeesh.

Issue #460, opened 2015-04-28
Create public methods for Text overlays


iriberri:
In the same way that we can do [this](https://github.com/CartoDB/cartodb.js/blob/develop/examples/annotations.html) with annotations, we've been asked for the same regarding Text boxes.

nobuti:
There has been no activity on this issue for several months and that's why are closing it (automatically). If you think this still needs to be addressed please reopen it. The good news is that we are now working on the next version of CARTO.js, stay tuned.

PR #462, opened 2015-04-28
Adds methods to create text overlays

javierarce:
Reference: https://github.com/CartoDB/cartodb.js/issues/460

### DIFF
--- a/examples/text.html
+++ b/examples/text.html
@@ -1,0 +1,77 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>Easy example | CartoDB.js</title>
+    <meta name="viewport" content="initial-scale=1.0, user-scalable=no" />
+    <meta http-equiv="content-type" content="text/html; charset=UTF-8"/>
+    <link rel="shortcut icon" href="http://cartodb.com/assets/favicon.ico" />
+    <style>
+      html, body, #map {
+        height: 100%;
+        padding: 0;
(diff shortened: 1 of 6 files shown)


examples/text.html
```diff
@@ -0,0 +1,77 @@
...
+              "font-family-name": "Helvetica",
+              "box-color": "#F84F40",
+              "box-opacity": 0.7,
+              "box-padding": 10,
+              "line-color": "#F84F40",
+              "line-width": 50
+            }
+          });
+
+          this.mapView.$el.append(text.render().$el);
```
javisantana: wouldn't vis.addOverlay work here?

javierarce: Yes, but I think it make sense to have a simple way to create text overlays, similar to the annotations.

javisantana: ok, so why instead of exposing internal things we don't expose something in visualization to add the annontations//texts and so on that tracks the removal and so on. ``addOverlay`` creates and adds an overlay, we need something that only adds it (because it is already created).

Exposing mapView means that if we change that implementation it will fail

javierarce: You're right. I'll work on that. Thanks!

nobuti:
There has been no activity on this issue for several months and that's why are closing it (automatically). If you think this still needs to be addressed please reopen it. The good news is that we are now working on the next version of CARTO.js, stay tuned.

PR #463, opened 2015-04-28
Fixes problems setting maxZoom and minZoom

javierarce:
This PR attempts to fix a problem that occurs when changing between basemaps with different zoom configurations.

See related issue: https://github.com/CartoDB/cartodb/issues/2250

`maxZoom` and `minZoom` get returned as a string (except in development, where they are returned as an integer), so the zoom is never correctly updated.


javierarce:

javierarce:
:dog: bump

javisantana:
:es:

PR #464, opened 2015-04-30
Updates version of Torque to 2.11.2

fdansv:
Changes since last replacement:

```
- Added error handling to Torque
- Fixed 'remove' event triggering when removing a Torque layer.
- Value is now capped at 255 instead of performing modulo on higher values.
```

@javierarce ;)


javierarce:
:flags:

Issue #458, opened 2015-04-24
make sure all the default basemaps are switch to http


javisantana:
https://github.com/CartoDB/cartodb-platform/pull/457

we have a mapping to convert https to http in case https is not active, be sure all the urls are converted right

https://github.com/CartoDB/cartodb.js/blob/develop/src/vis/layers.js#L11

santisaez:
@fdansv: we've just merged https://github.com/CartoDB/cartodb-platform/pull/457, so now we're forcing all basemap HTTP requests to always use `https://`.

zenitraM:
That mapping is missing our recent basemaps:
`{s}.basemaps.cartocdn.com` should become `cartodb-basemaps-{s}.global.ssl.fastly.net` on SSL

fdansv:
ACK :+1:

fdansv:
Fixed/added URL templates in #465.

PR #465, opened 2015-04-30
Adds new basemap url scheme to https to http converter

fdansv:
Ref. #458 

Not sure if there is anything else you want to add.

@javisantana @zenitraM


zenitraM:
:+1:

Issue #466, opened 2015-05-04
add sandwich support 


javisantana:
maps api supports not only rendering cartodb layers but http and plain layers (see 2.3 here https://github.com/CartoDB/Windshaft/blob/master/doc/MapConfig-1.3.0.md)

layergroup in cartodb only supports cartodb layers, we should add api to support those layers.

The api should look like

```
 cartodb.createLayer(map, {
          user_name: 'examples',
          type: 'cartodb',
          sublayers: [
          {
                type: "http",
                urlTemplate: "http://{s}.basemaps.cartocdn.com/light_nolabels/{z}/{x}/{y}.png"
                //subdomains: [ "a", "b", "c" ]
          },
          {
             sql: 'select * from country_boundaries',
             cartocss: '#layer { polygon-fill: #F00; polygon-opacity: 0.3; line-color: #F00; }'
             }, {
               type: 'http',
               //urlTemplate: 'http://a.basemaps.cartocdn.com/light_only_labels/{z}/{x}/{y}.png',
               urlTemplate: 'http://{s}.tile.stamen.com/toner-labels/{z}/{x}/{y}.png',
               subdomains: [ "a", "b", "c" ]
             }]
        })
        .addTo(map)
        .done(function(layer) {
          setTimeout(function() {
            layer.getSubLayer(0).setUrlTemplate(
               'http://{s}.tile.stamen.com/toner/{z}/{x}/{y}.png'
            )
            layer.getSubLayer(2).hide();
          }, 3000)

```

I created a draft here to test the implementation but it does need to be used: https://github.com/CartoDB/cartodb.js/pull/445

javisantana:
@rochoa will point here the latest documentation so @alonsogarciapablo can finish the integration (and the documentation, to be reviewed by @ohasselblad )

saleiva:
yay!

ohasselblad:
:+1: i'm excited to play with sandwiches

rochoa:
Initial draft: https://github.com/CartoDB/Windshaft-cartodb/pull/301

PR #470, opened 2015-05-05
Cleanup

alonsogarciapablo:
In preparation for #466, in this PR:
- I've created `cdb.core.util` and moved some utility methods that were included in `layer_definition.js`.
- I've also re-organized some methods in `layer_definition.js` and organized tests a bit.
- I've moved 'Sublayer' to it's own file.
- I've moved some specs to the right folder.

@javisantana PTAL.

PS: I still need to add `util.js` to grunt tasks.


javisantana:
don't forget to add util.js to cartodb.js file list (it's the way to add files to the build, probably we need to change that)

Apart from that It's perfect

alonsogarciapablo:
retest this please

alonsogarciapablo:
retest this please

alonsogarciapablo:
@javisantana I [added util.js to the list of files in src/cartodb.js](https://github.com/CartoDB/cartodb.js/commit/b6ddb85fa3e472038d144fbe4731723091c613ea). Do I need to add it anywhere else? Otherwise I will merge this.

alonsogarciapablo:
@javisantana I've added some more commits with some more cleanup. PTAL. I've updated the description with the changes included in this PR. I'd like to merge this before continuing working on #466,  so that we don't end up having a massive PR. Thanks!

src/geo/layer_definition.js
```diff
@@ -661,242 +456,169 @@ Map.prototype = {
...
+  var layer_definition = {
+    version: '1.0.0',
+    stat_tag: 'API',
+    layers: []
+  };
 
+  for (var i = 0; i < sublayers.length; ++i) {
+    layer_definition.layers.push({
+      type: 'cartodb',
+      options: sublayers[i]
```
alonsogarciapablo: Hey @javisantana, I think there's something wrong here. We're adding adding everything in `sublayers[i]` to the `options` attribute. That's generating something like:
```
{
  type: 'cartodb',
  options: {
    sql: '...',
    cartocss: '...',
  }
}
```

But according to the spec, it should be something like:

```
{
  type: 'cartodb',
  sql: '...',
  cartocss: '...',
  options: {
    ...
  }
}
```

Also, methods like `Sublayer#setSQL` are setting those attribute inside of the `options` attribute, which seems wrong to me.

What do you think?

javisantana: what spec?

alonsogarciapablo: I was referring to MapConfig, but it's actually right :)

Issue #414, opened 2015-03-27
Cartodb.js doesn't render torque map if jquery is loaded


martinburch:
I am using the following example code, which is the same as [the easy.html example in this repo](https://github.com/CartoDB/cartodb.js/blob/develop/examples/easy.html) except I am loading jQuery separately.

According to [this documentation](http://docs.cartodb.com/tutorials/toggle_map_view.html), that's allowed:
> CartoDB.js uses jQuery already, and when it detects that you haven’t loaded it it makes it available for your use. That way there is no need to duplicate library imports. But don’t worry, it will respect any JQuery version you optionally load elsewhere.

Current behavior: only a static map is loaded.
Desired behavior: torque animations display and interactivity is enabled through a time_slider control.

Note: the viz.json in the example below is a public viz from Andrew Hill's demo. I first encountered this bug using my own freshly-generated viz.json but I don't want to make my table public yet as part of this bug report.

```
<!DOCTYPE html>
<html>
  <head>
    <title>Easy example | CartoDB.js</title>
    <meta name="viewport" content="initial-scale=1.0, user-scalable=no" />
    <meta http-equiv="content-type" content="text/html; charset=UTF-8"/>
    <link rel="shortcut icon" href="http://cartodb.com/assets/favicon.ico" />
    <style>
      html, body, #map {
        height: 100%;
        padding: 0;
        margin: 0;
      }
    </style>

    <link rel="stylesheet" href="http://libs.cartocdn.com/cartodb.js/v3/3.12/themes/css/cartodb.css" />
  </head>
  <body>
    <div id="map"></div>

	<script src="https://ajax.googleapis.com/ajax/libs/jquery/1.10.2/jquery.min.js"></script>
    <script src="http://libs.cartocdn.com/cartodb.js/v3/3.12/cartodb.js"></script>

    <script>
      function createmap() {
        cartodb.createVis('map', 'http://andrew.cartodb.com/api/v2/viz/b5cece38-4af4-11e3-bfb4-3085a9a9563c/viz.json');
      }

      window.onload = createmap;
    </script>
  </body>
</html>
```

martinburch:
Via support email, @iriberri suggests I try cartodb_nojquery.js instead of relying on cartodb.js to detect jQuery has been loaded.

That partially works, but cartodb.js relies on an ancient version of jQuery (version 1.8.1 released 3 years ago) with browser-detection features. Not having these features causes trouble dragging the map.

[...]

PR #471, opened 2015-05-07
Uses internal version of jQuery in maps with Torque

fdansv:
Ref. #414 

A header in the Torque module of Cartodb.js that makes sure that the bundled version of jquery (1.7.2) is used, and not one specified in the DOM. Also, as in the visualisations without torque, it restores the user-specified version if there is any.

@javisantana


fdansv:
retest this please

javisantana:
some news about this?

fdansv:
As discussed, this solution is good to go!

PR #474, opened 2015-05-08
Add note about SSL URLs

pramsey:
Added a note at the end of the docs about SSL URLs to the API. Better fix would be to make `https://libs.cartocdn.com/` equal to `https://cartodb-libs.global.ssl.fastly.net`


javisantana:
how is this different that https://github.com/CartoDB/cartodb.js/blob/develop/doc/API.md#https-support

pramsey:
It isn't.

PR #475, opened 2015-05-11
Not using common folder for templates, now it is old_common

xavijam:


javierarce:
:ok_hand:

xavijam:
thanks @javierarce

PR #479, opened 2015-05-18
Hide <img> tags in infowindow covers if url is invalid

### LINKED ISSUE #461: Hide or remove `<img>` tag of the infowindows with image header
If [the URL of an infowindow cover is invalid](https://github.com/CartoDB/cartodb.js/blob/develop/src/geo/ui/infowindow.js#L568) (it might not even be a URL for example), we should remove or hide the `` tag that is included in the cover, as we did in cartodb to prevent displaying a broken `` tag.

NOTE: Nobody has complaint about this, so it's not a big deal.

alonsogarciapablo:
Fixes #461 (as [we did in CartoDB](https://github.com/CartoDB/cartodb/commit/331058ffdb4dbd5aa4a3b958504d0195c1769bf6)).

@xavijam please take a look.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,6 @@
+3.14.X ()
+* Hide <img> tag of infowindow covers when the url is invalid
+
 3.14.2 (06//05//2015)
 * Allow to specify a template for the items of a custom legend.
 * The NOKIA geocoder doesn't encode the whitespaces anymore.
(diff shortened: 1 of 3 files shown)


xavijam:
Looks good, don't forget to do the same in the infowindow.js file in the editor.

alonsogarciapablo:
@xavijam Thanks! It's already been [changed](https://github.com/CartoDB/cartodb/commit/331058ffdb4dbd5aa4a3b958504d0195c1769bf6) in the editor.

PR #486, opened 2015-05-21
Prevent tooltips from being shown when no fields have been selected

alonsogarciapablo:
This fixes cartodb/cartodb#3617. The problems were:

1. There was a layer filled with features (timezones) on top of the map. So we were hovering a feature all the time.
2. We were adding the tooltip cause the other layer had a tooltip (with fields) and we were not checking if layers have fields in the tooltip when we enable.

This PR fixes that by preventing a tooltip from being shown if there are no fields.

cc// @javisantana


javisantana:
:es:

PR #488, opened 2015-05-22
Only show legends and torque controls for visible layers

alonsogarciapablo:
Fixes cartodb/cartodb#3351.

With this PR, cartodb.js will only:

- Display legends for visible layers.
- Display torque controls if torque layer is visible.

@fdansv can you please take a look? thanks!


fdansv:
Looks good to me!

PR #490, opened 2015-05-25
Fix issue with infowindow scrolling due to slow image loading:

alonsogarciapablo:
Fixes cartodb/cartodb#3515:

- Enables jscrollpane's autoReinitialise option so that the plugin
  correctly scrolls after the image has been loaded.
- Enables jscrollpane's maintainPosition option.

More info about these options [here](http://jscrollpane.kelvinluck.com/settings.html).

Before:

before

After:

after

@xavijam PTAL. Thanks!


xavijam:
yay! good fix, jscrollpane is one of the hardest.

PR #493, opened 2015-05-26
Unbind dropdown bound events on clean

### LINKED ISSUE #491: Dropdown target handler not unbound when view is cleaned
[This bind's callback](https://github.com/CartoDB/cartodb.js/blob/b71d2223633db21d07c0ecea35326e29ccd7a201/src/ui/common/dropdown.js#L53) is never unbound, which means that if this view is instantiated with a target the target's default action won't work after the view is destroyed. Should be unbound on `.clean()`.

viddo:
Fixes #491, by overriding clean to unbind the event bindings done on none-view elements in the constructor.

### DIFF
--- a/src/ui/common/dropdown.js
+++ b/src/ui/common/dropdown.js
@@ -138,6 +138,12 @@ cdb.ui.common.Dropdown = cdb.core.View.extend({
     this.isOpen = true;
   },
 
+  clean: function() {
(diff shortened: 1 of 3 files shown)


viddo:
@javisantana @xavijam can you review, please?

xavijam:
:jp: finally taking care of the bug, thanks!

PR #494, opened 2015-05-26
Use http version of google maps api in tests

alonsogarciapablo:
cc/ @viddo


viddo:
:ok_hand: works on my machine™ (too) :dancer:

Issue #121, opened 2014-06-03
Create new example using Leaflet layer control


iriberri:
It would be useful to have an example showing how to include several CartoDB layers with Leaflet.

javisantana:
assign it whoever you want

namessanti:
@andrewxhill

iriberri:
Actually: http://bl.ocks.org/iriberri/074a71940dc606d98079

On Tuesday, May 26, 2015, Santiago Giraldo Anduaga <notifications@github.com>
wrote:

> @andrewxhill <https://github.com/andrewxhill>
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/CartoDB/cartodb.js/issues/121#issuecomment-105661358>.
>


-- 

Carla Iriberri
*CartoDB*

andrewxhill:
boom

On Tue, May 26, 2015 at 5:09 PM, Carla <notifications@github.com> wrote:

> Actually: http://bl.ocks.org/iriberri/074a71940dc606d98079
>
> On Tuesday, May 26, 2015, Santiago Giraldo Anduaga <
> notifications@github.com>
> wrote:
>
> > @andrewxhill <https://github.com/andrewxhill>
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <https://github.com/CartoDB/cartodb.js/issues/121#issuecomment-105661358
> >.
> >
>
>
> --
>
> Carla Iriberri
> *CartoDB*
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/CartoDB/cartodb.js/issues/121#issuecomment-105665388>.
>



-- 

andrewxhill <http://twitter.com/andrewxhill>
http://cartodb.com

javisantana:
@iriberri why don't you add that example to examples folder?

PR #497, opened 2015-05-27
121 new control layers example

iriberri:
Closes https://github.com/CartoDB/cartodb.js/issues/121#issuecomment-105776740

There's already a similar example that uses createLayer sublayers instead of two different createLayers, but I think having several approaches with different possibilities is useful for the community.

All the viz.jsons called in this example are based in the `documentation` account.

cc @javisantana


iriberri:
huh, I got other commits here :S

javisantana:
I think it only adds one file so it's safe :es:

PR #503, opened 2015-06-02
Forces https basemaps if https property has been explicitly defined

### LINKED ISSUE #500: "The page was loaded over HTTPS, but requested an insecure image 'http://d.basemaps.cartocdn.com/dark_nolabels/12/656/1583.png'."
"This content should also be served over HTTPS."

STR

- go to cartodb.com 
- `createVis` is being called with `https` options

**expected behaviour:**

- from documentation:

> https: if true forces tiles to be fetched using https. If false it uses the predefined method

**actual behaviour:**

- tiles are requestd via HTTP

example:

https://bl.ocks.org/matallo/0ffb28ccafe1ded67574

/cc @fdansv

fdansv:
Fixes #500 

@matallo

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -198,7 +198,7 @@ cartodb.createVis('map', url)
   - **time_slider**: show time slider with torque layers (enabled by default)
   - **layer_selector**: show layer selector (default: false)
   - **legends**: if it's true legends are shown in the map.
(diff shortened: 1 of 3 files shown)


matallo:
looks good to me, thanks for the https support, what about adding a small case breakdown in the docs?

fdansv:
Sure thing, I'll add that and merge.

PR #504, opened 2015-06-02
Adds support for multiple types of layers (aka Sandwich)

alonsogarciapablo:
image

Implements #466.

@javisantana please take a look when you have a sec. [This commit](https://github.com/CartoDB/cartodb.js/commit/54443acbebb174754227ee6c1a8a3f8da633120d) will give you a good idea of how we're storing and using the metadata we're receiving from the tiler. I've left a couple of TODOs about some possible renamings that would make (IMHO) layer_definition.js easier to understand. Thanks for your feedback!


examples/leaflet_sandwich.html
```diff
@@ -0,0 +1,82 @@
...
+      }
+    </style>
+
+    <link rel="stylesheet" href="http://libs.cartocdn.com/cartodb.js/v3/3.13/themes/css/cartodb.css" />
+  </head>
+  <body>
+    <div id="map"></div>
+
+    <!-- include cartodb.js library -->
+    <script src="http://libs.cartocdn.com/cartodb.js/v3/3.14/cartodb.core.js"></script>
```
javisantana: how is possible than using core version you can use createLayer ? it should not be possible

alonsogarciapablo: You're totally right. I was only testing this locally. I will fix it.

PR #513, opened 2015-06-09
Do not enable layer interaction if tooltip is empty

alonsogarciapablo:
Fixes cartodb/cartodb#3949.

We're enabling interaction on every sublayer created using the editor and that was causing some issues. We should only enable interaction when tooltips have 

@javierarce can you PTAL? Thanks!


javierarce:
:+1: code looks good to me.

PR #522, opened 2015-06-12
Use the right indexes when fetching grids and attributes

### LINKED ISSUE #518: interactivity layer index is wrong with named maps where cartodb layer is not the first one
- create a named map with these layers:
  - http
  - cartodb (with some interaction)
- using createLayer instance that named map and enable interactivity

check the requests done by the client, it tries to fetch grid.json tiles for layer 0 when it should be layer 1

THis needs to be fixed in 3.11, 3.12, 3.13, 3.14 and 3.15 branches

alonsogarciapablo:
Fixes #518. We know use the metadata returned by the tiler to determine the right indexes of mapnik layers in the map (anonymous or named map). This is important in when generating the urls for grids and the urls to fetch the attributes of mapnik layers.

@javisantana, what do you think?

### DIFF
--- a/src/geo/layer_definition.js
+++ b/src/geo/layer_definition.js
@@ -1,3 +1,33 @@
+/**
+ * Wrapper for map properties returned by the tiler
+ */
+function MapProperties(mapProperties) {
+  this.mapProperties = mapProperties;
+}
+
+MapProperties.prototype.getMapId = function() {
+  return this.mapProperties.layergroupid;
+}
+
+/**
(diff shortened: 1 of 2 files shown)


javisantana:
sounds good, it's going to hard to backport as is, right?

alonsogarciapablo:
@javisantana yes :(

alonsogarciapablo:
I've squashed all commits into one so it's easy to apply the patch to the rest of versions of cartodb.js. `layer_definition.js` has changed a bit since 3.11, so will have to do some manual merging.

PR #524, opened 2015-06-16
Makes the slider have the correct range every time it changes

fdansv:
Fixes https://github.com/CartoDB/cartodb/issues/3951

... by using always the number of steps given by the torque provider, as opposed to the one in the cartocss properties.

@javisantana


fdansv:
retest this please

PR #534, opened 2015-06-22
Uses local time in timeslider instead of UTC

fdansv:
nogmt

Fixes https://github.com/CartoDB/cartodb/issues/4135
@alonsogarciapablo mind a quick CR? :)

cc @saleiva


alonsogarciapablo:
:+1:

Issue #519, opened 2015-06-11
Allow viz.json URLs that don't start with "http"


alonsogarciapablo:
Right now, [we only support URLs that start with "http"](https://github.com/CartoDB/cartodb.js/blob/sandwich/src/api/layers.js#L45).

We should also be able to do things like:
```
cartodb.createLayer(map, './viz.json').addTo(map);
```

cc// @javisantana

javisantana:
:D

fdansv:
OK so we want to use relative addresses. Here's the thing: are we assuming that the file will always be called via jsonp? (adding a `<script>` tag to the DOM). If the file is in the same domain, as it is in this use case, there is no need for the use of jsonp, right? What do you think of this:

https://github.com/CartoDB/cartodb.js/blob/03688591d2ec70f9e66cfc77988a03f537662c42/src/core/loader.js#L24-L37

alonsogarciapablo:
Sounds reasonable to me.

fdansv:
Fixed in #536

PR #535, opened 2015-06-22
Allows the use of relative addresses to load vizjsons

fdansv:
@alonsogarciapablo what do you think?

Fixes https://github.com/CartoDB/cartodb.js/issues/519


fdansv:
ah crap need to change a couple things

src/core/loader.js
```diff
@@ -30,6 +34,12 @@ var Loader = cdb.vis.Loader = cdb.core.Loader = {
     }
   },
 
+  getNoJSONP: function(url, callback) {
+    $.get(url).done(function(data){
```
javisantana: no jquery is allowed here, it could not be loaded

javisantana:
hey, this thing is as easy as remove this check https://github.com/CartoDB/cartodb.js/blob/develop/src/api/layers.js#L45

fdansv:
damn, it seems i've overcomplicated it a bit... ok changing

test/spec/api/layers.spec.js
```diff
@@ -222,6 +222,17 @@ describe('api.layers', function() {
 
       });
 
+      it("should load local viz.json correctly", function(done) {
+        sinon.spy($, "get");
+        cartodb.createLayer(map, "../../demos/viz.json")
+            .addTo(map);
+        setTimeout(function(){
+          expect($.get.getCall(0)); // If jquery get has been called, means it's loading it as a local, non jsonp vizjson
```
alonsogarciapablo: Did you try `expect($.get).toHaveBeenCalled()`?

fdansv: I was using sinon, not jasmine in this case. I'm closing this PR and opening a new one with the more simple change :)

PR #536, opened 2015-06-22
Removes check for http-beginng vizjson addresses

fdansv:
@javisantana @alonsogarciapablo 
Fixes #535


src/api/layers.js
```diff
@@ -42,7 +42,7 @@
     } else if(layer.table !== undefined && layer.user !== undefined) {
       // layer object points to cartodbjson
       url = cartodbUrl(layer);
-    } else if(layer.indexOf && layer.indexOf('http') === 0) {
```
javisantana: I'd just leave the else

PR #539, opened 2015-06-23
fixes 538 adding the isVisible function

jsanz:
I hope everything is fine :smile:

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -653,6 +653,10 @@ Removes the sublayer from the layer temporarily. The layer is refreshed after ca
 
 Toggles the visibility of the sublayer and returns a boolean that indicates the new status (true if the sublayer is visible, false if it is hidden)
 
+#### sublayer.isVisible()
+
+Gets the visibility of the sublayer.
+
 ### cartodb.CartoDBLayer.CartoDBSubLayer
 
 #### sublayer.getSQL()
--- a/doc/API.md
+++ b/doc/API.md
@@ -655,7 +655,7 @@ Toggles the visibility of the sublayer and returns a boolean that indicates the 
 
 #### sublayer.isVisible()
 
-Gets the visibility of the sublayer.
+It returns `true`  if the sublayer is visible.
 
 ### cartodb.CartoDBLayer.CartoDBSubLayer
 
--- a/test/spec/geo/sublayer.spec.js
+++ b/test/spec/geo/sublayer.spec.js
@@ -141,6 +141,17 @@ describe('Sublayers', function() {
       })
     });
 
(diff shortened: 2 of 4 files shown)


jsanz:
Should I also document this, right?

test/spec/geo/sublayer.spec.js
```diff
@@ -141,6 +141,17 @@ describe('Sublayers', function() {
       })
     });
 
+    describe('.isVisible', function() {
```
alonsogarciapablo: What about this:

```
it('should return true if sublayer is visible', function() {
  sublayer.set('hidden', undefined);

  expect(sublayer.isVisible()).toBeTruthy();

  sublayer.set('hidden', false);

  expect(sublayer.isVisible()).toBeTruthy();
})

it('should return false if sublayer is hidden', function() {
  sublayer.set('hidden', true);

  expect(sublayer.isVisible()).toBeFalsy();
})

```

jsanz: Better, thanks!

doc/API.md
```diff
@@ -653,6 +653,10 @@ Removes the sublayer from the layer temporarily. The layer is refreshed after ca
 
 Toggles the visibility of the sublayer and returns a boolean that indicates the new status (true if the sublayer is visible, false if it is hidden)
 
+#### sublayer.isVisible()
+
+Gets the visibility of the sublayer.
```
alonsogarciapablo: :+1: for updating the docs.

I'd be a bit more explicit about what this method returns. For example:
```
It returns `true` if the sublayer is visible.
```

test/spec/geo/sublayer.spec.js
jsanz: Weird this works to test the falsy case

```javascript
        sublayer.hide();
        expect(sublayer.isVisible()).toBeFalsy();
```

but this doesn't

```javascript
        sublayer.set('hidden', true);
        expect(sublayer.isVisible()).toBeFalsy();
```

It's OK?

alonsogarciapablo: Sorry, perhaps it's:

```
sublayer.set({
  hidden: true
});
```
That's what `.hide` does internally.

jsanz: Thanks @alonsogarciapablo I think now it's more complete and working both tests

alonsogarciapablo:
retest this please!

jsanz:
Sorry :disappointed:, now it's passing the tests I think

alonsogarciapablo:
:+1: Good job!

PR #544, opened 2015-06-24
Fix #512, mousewheel error

### LINKED ISSUE #512: mousewheel events error
Basically anywhere I scroll I get this error:

```no-highlight
Uncaught TypeError: Cannot read property 'apply' of undefined cartodb.js:16885 

$.event.special.mwheelIntent.handler	@	cartodb.js:16885
jQuery.event.dispatch	@	jquery.js:4670
handler	@	cartodb.js:16825
```

Should https://github.com/CartoDB/cartodb.js/blob/master/vendor/mwheelIntent.js#L57 be `($.event.dispatch || $.event.handle)` like here: https://github.com/CartoDB/cartodb.js/blob/b71d2223633db21d07c0ecea35326e29ccd7a201/vendor/mousewheel.js#L81

knownasilya:
Not sure why this is happening, maybe I'm doing something wrong elsewhere, but this should fix it. See the original issue for details.

### DIFF
--- a/vendor/mwheelIntent.js
+++ b/vendor/mwheelIntent.js
@@ -54,7 +54,7 @@ $.event.special.mwheelIntent = {
         minDif = 3;
       }, 1500);
       e = $.extend({}, e, {type: 'mwheelIntent'});
-            return $.event.handle.apply(this, arguments);
+            return ($.event.dispatch || $.event.handle).apply(this, arguments);
     }
     }
 };


knownasilya:
Ping :shipit:

knownasilya:
Anyone? :+1:

knownasilya:
Feels kind of stale in here :wink:

vhamer:
Sorry for the delay, all! @knownasilya is already a contributor, feel free to merge when ready!

javisantana:
ideally we should fix this in the original library https://github.com/jquery/jquery-mousewheel, I think they are pretty ahead of this right now. 

Not sure if is still compatible, if so we should replace the old library

xavijam:
Take into account that library was included due to scrollPane dependency.

javisantana:
but looks like a different module

javisantana:
ok, let's merge it, we will update the full stuff

PR #546, opened 2015-06-26
Take `visible` attribute into account when serialising maps 

### LINKED ISSUE #541: createLayer doesn't follow vizjson layers visible setting
Adding a visualization using ``createLayer`` doesn't follow the settings from the editor about layers visibility so all layers are shown. I've tried this both on GMaps and Leaflet so I'm not sure if this is  something intentional or a bug (??)

alonsogarciapablo:
Fixes #541. We were using `options.hidden` to determine if layers in "LayerDefinition"s or "NamedMap"s were visible.  But that is something that only gets set after invoking `sublayer.hide()` or `sublayer.show()`. With this PR, we also take into account the `visible` attribute that (sub)layers have in viz.json files so that layers are correctly displayed after loading a viz.json file.

@javisantana, can you please take a look at this? Thanks!

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,6 @@
+3.X ()
+* Take `visible` attribute into account when determining visibility of layers and serializing maps (#546)
+
 3.15 (24//06//2015)
 * cartodb.js knows how to work with multiple types of sublayers (#508):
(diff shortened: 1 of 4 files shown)


javisantana:
:es:

PR #548, opened 2015-06-30
Describe [WIP]

javisantana:
- It lacks of specs and documentation (that's fine for merging but should be done)
- I'd move html files to examples


src/geo/ui/search.js
```diff
@@ -11,7 +11,9 @@ cdb.geo.ui.Search = cdb.core.View.extend({
     "mousedown":                  '_stopPropagation'
   },
 
-  initialize: function() {},
+  initialize: function() {
```
javisantana: does this change belong here?

javierarce: I don't understand what's going on here :|

Issue #367, opened 2015-02-23
Bower version not built


knownasilya:
Installing cartodb.js from bower just provides a copy of this repo, which is pretty much useless since the point is to automate asset retrieval. I'd expect the version on bower to be a built version (the files produced by the grunt build step).

rowanwins:
+1 please

ElMesa:
Yep, plz solve this, Grunt+Bower haves it's audience. Me included

matallo:
you can do now a `bower install cartodb.js`and it will install the dist files from https://github.com/CartoDB/cartodb.js-bower

PR #551, opened 2015-07-01
prepare for bower

matallo:
as long as we check-in the dist files this issue https://github.com/CartoDB/cartodb.js/issues/367 would be solved

this is for example what Backbone or Bootstrap do
https://github.com/jashkenas/backbone
https://github.com/twbs/bootstrap

or, in order to not to pollute the history the dist files could be pushed to a different repo, like Foundation does after a successful publish

https://github.com/zurb/foundation/blob/master/.travis.yml
https://github.com/zurb/bower-foundation

your call @javisantana @alonsogarciapablo, what do you guys think should be the best approach here?

on the long term, it would be great to pull the dependencies from bower, too

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,8 +1,8 @@
 {
   "name": "cartodb.js",
   "main": [
-    "dist/cartodb.js",
-    "dist/themes/css/cartodb.css"
+    "cartodb.js",
+    "themes/css/cartodb.css"
   ],
   "version": "3.15.1",
   "homepage": "https://github.com/CartoDB/cartodb.js",
--- a/bower.json
+++ b/bower.json
@@ -1,10 +1,20 @@
 {
   "name": "cartodb.js",
-  "version": "0.0.1",
-  "dependencies": {},
-  "devDependencies": {
-    "backbone": "~1.1.2",
-    "modernizr": "~2.8.3",
-    "jquery": "~2.1.1"
-  }
+  "main": [
+    "dist/cartodb.js",
+    "dist/themes/css/cartodb.css"
+  ],
+  "version": "3.15.1",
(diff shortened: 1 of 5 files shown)


alonsogarciapablo:
I'm okay with this solution. We should update `RELEASING.md`, change the `.gitignore` so that those files are not ignored and commit/push them.

matallo:
we do have now a cartodb.js-bower repo https://github.com/CartoDB/cartodb.js-bower

to publish to that repo, each time we release a new version remember to

```sh
./bower.sh
```

ideally we should move this to the `.travis.yml` but this will work for now

CR @javisantana @alonsogarciapablo

matallo:
ping

bower.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
```
alonsogarciapablo: :+1: Nice script!

bower.json
```diff
@@ -1,10 +1,20 @@
...
-  "devDependencies": {
-    "backbone": "~1.1.2",
-    "modernizr": "~2.8.3",
-    "jquery": "~2.1.1"
-  }
+  "main": [
+    "cartodb.js",
+    "themes/css/cartodb.css"
+  ],
+  "version": "3.15.1",
```
alonsogarciapablo: It would be nice if this number got automagically updated when someone runs `grunt build`. Right now [we update some versions in examples, README, etc.](https://github.com/CartoDB/cartodb.js/blob/develop/grunt/tasks/replace.js).

alonsogarciapablo:
Great work @matallo. I left a comment with a nice-to-have :)

bower.json
matallo: oh, I add it as a TODO :smile:

PR #552, opened 2015-07-02
adds a way to build the current dist without user input

javisantana:
so now we have:

- dist task to build the current version
- default task is dist
- the release task is the one that ask about new version

there were several problems:
- imagemin was failing silently when prompt task was being run so when it's not run everything failed because a bug in imagemin. Does make sense to use imagemin?
- the build was working by chance, some of the needed variables were being added to the config when they should be part of the config from the start
- grunt was not reporting any error and it took 3 hours to realize then the process freeze

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -19,6 +19,11 @@ module.exports = function(grunt) {
   }
 
   var version = pkg.version.split('.');
+  var VERSION_OBJ =  {
+      major:      version[0],
+      minor:      version[0] + '.' + version[1],
+      bugfixing:  pkg.version
+   }
 
   var config = {
     dist: 'dist',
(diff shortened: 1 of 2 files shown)


javisantana:
cc @xavijam

xavijam:
retest this please

xavijam:
cc @CartoDB/frontend

javisantana:
cc @azamorano you can execute just grunt to build the current version

javisantana:
cc @CartoDB/frontend please read since the default task has changed

xavijam:
Further to ```imagemin```, we should keep it because it is nice to have a task that optimizes production images. Of course, if we have it, we should make it work properly :D.

xavijam:
In any case :es:.

azamorano:
thanks a lot!! :+1:

PR #555, opened 2015-07-03
[WIP] Labels on top

alonsogarciapablo:
This PR contains all the necessary changes to support a tiled layer with labels on top of a map.

### DIFF
--- a/src/geo/layer_definition.js
+++ b/src/geo/layer_definition.js
@@ -598,7 +598,7 @@ MapBase.prototype = {
     var layers = [];
     for(var i = 0; i < this.layers.length; ++i) {
       var layer = this.layers[i];
-      if(layer.options && !layer.options.hidden) {
+      if(this._isLayerVisible(layer)) {
         layers.push(i);
       }
     }
--- a/src/geo/layer_definition.js
+++ b/src/geo/layer_definition.js
@@ -911,7 +911,7 @@ NamedMap.prototype = _.extend({}, MapBase.prototype, {
     this.layers = _.clone(named_map.layers) || [];
     for(var i = 0; i < this.layers.length; ++i) {
       var layer = this.layers[i];
-      layer.options = layer.options || { hidden: false };
+      layer.options = layer.options || { 'hidden': layer.visible === false };
       layer.options.layer_name = layer.layer_name;
     }
     this.named_map = named_map;
--- a/src/geo/map.js
+++ b/src/geo/map.js
@@ -226,6 +226,7 @@ cdb.geo.Layers = Backbone.Collection.extend({
       return parseInt(m.get('order'), 10);
     };
     this.bind('add', this._assignIndexes);
+    this.bind('remove', this._assignIndexes);
   },
 
   /**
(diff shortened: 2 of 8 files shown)


test/spec/geo/map.spec.js
```diff
@@ -47,7 +47,7 @@ describe("geo.map", function() {
     });
 
     it("should compare equal layers correctly", function() {
-      var layer1 = new cdb.geo.PlainLayer({color: '#zipote'});
```
javisantana: XDD

src/geo/layer_definition.js
```diff
@@ -911,7 +911,7 @@ NamedMap.prototype = _.extend({}, MapBase.prototype, {
     this.layers = _.clone(named_map.layers) || [];
     for(var i = 0; i < this.layers.length; ++i) {
       var layer = this.layers[i];
-      layer.options = layer.options || { hidden: false };
+      layer.options = layer.options || { 'hidden': layer.visible === false };
```
alonsogarciapablo: Layers inside of the `named_map` key has a `visible` attribute which should determine the visibility of layers. It was broken until now.

src/geo/map.js
```diff
@@ -39,7 +39,10 @@ cdb.geo.MapLayer = cdb.core.Model.extend({
       if(myType === 'Tiled') {
         var myTemplate  = me.urlTemplate? me.urlTemplate : me.options.urlTemplate
           , itsTemplate = other.urlTemplate? other.urlTemplate : other.options.urlTemplate;
-        return myTemplate === itsTemplate;
+        var myName = me.name? me.name : me.options.name;
+        var itsName = other.name? other.name : other.options.name;
+
+        return myTemplate === itsTemplate && myName === itsName;
```
alonsogarciapablo: Two `Tiled` layers could potentially have the same `urlTemplate` and a different layer with labels. We now also use the name to determine if two layers are equal.

src/geo/map.js
```diff
@@ -223,6 +226,7 @@ cdb.geo.Layers = Backbone.Collection.extend({
       return parseInt(m.get('order'), 10);
     };
     this.bind('add', this._assignIndexes);
+    this.bind('remove', this._assignIndexes);
```
alonsogarciapablo: We should reassign order of layers when a layer is removed from the collection.

alonsogarciapablo:
@xavijam I would like you to take a look at this PR since it has some stuff related to the order of layers... I've included some comments to explain what each change is about. Thanks!

src/geo/map.js
```diff
@@ -230,25 +234,37 @@ cdb.geo.Layers = Backbone.Collection.extend({
    * the index should be recalculated
    */
   _assignIndexes: function(model, col, options) {
```
alonsogarciapablo: Torque layers had a weight of 100. That meant that torque layers had orders like `101`, `102`, etc. This changes how orders are assigned so that layers are given sequential orders. The order of layers is now:

- Tiled Baselayer
- Mapnik layers
- (Torque layers)
- (Tiled layer with labels)

src/geo/map.js
xavijam: ha, ok, I wonder why my or its naming :dancer:.

src/geo/map.js
```diff
@@ -230,25 +234,37 @@ cdb.geo.Layers = Backbone.Collection.extend({
...
-    function layerWeight(layer) {
-      var t = layer.get('type');
-      return layerTypeWeight[t] || 0;
-    }
-    var from = 0;//this.size() - 1;
-    if(options && options.at !== undefined) {
-      from = options.at;
-    }
-    if(from === 0) {
+    var TILED_LAYER_TYPE = 'Tiled';
```
xavijam: Could we have those static variables at the beginning of the Class?

src/geo/map.js
```diff
@@ -230,25 +234,37 @@ cdb.geo.Layers = Backbone.Collection.extend({
...
+        for (var i = 0; i < sortedTypes.length; ++i) {
+          var type = sortedTypes[i];
+          var layers = layersByType[type] || [];
+          for (var j = 0; j < layers.length; ++j) {
+            var layer = layers[j];
+            layer.set({
+              order: ++lastOrder
+            });
+          }
+        }
```
xavijam: With this stuff, you try to order properly (and by type) all available layers right?

src/geo/map.js
```diff
@@ -39,7 +39,10 @@ cdb.geo.MapLayer = cdb.core.Model.extend({
...
         var myTemplate  = me.urlTemplate? me.urlTemplate : me.options.urlTemplate
           , itsTemplate = other.urlTemplate? other.urlTemplate : other.options.urlTemplate;
-        return myTemplate === itsTemplate;
+        var myName = me.name? me.name : me.options.name;
+        var itsName = other.name? other.name : other.options.name;
+
+        return myTemplate === itsTemplate && myName === itsName;
       } else if(myType === 'WMS') {
         var myTemplate  = me.urlTemplate? me.urlTemplate : me.options.urlTemplate
           , itsTemplate = other.urlTemplate? other.urlTemplate : other.options.urlTemplate;
```
xavijam: Could we change those variables creation to our standard way?
https://github.com/CartoDB/cartodb-management/wiki/CartoDB-JavaScript-Style-Guide#var
I know this is not related with this issue but I've just seen it, please :angel:.

src/geo/map.js
xavijam: :+1:

src/geo/map.js
```diff
@@ -230,25 +234,37 @@ cdb.geo.Layers = Backbone.Collection.extend({
...
-    }
-    if(from === 0) {
+    var TILED_LAYER_TYPE = 'Tiled';
+    var CARTODB_LAYER_TYPE = 'CartoDB';
+    var TORQUE_LAYER_TYPE = 'torque';
+
+    if (this.size() > 0) {
+
+      // Assign an order of 0 to the first layer
       this.models[0].set({ order: 0 });
```
xavijam: Is this.models a Backbone.Collection? If so, we should make use of their API, this.models.at(0) (super picky thing and it is not necessary buddy).

xavijam:
LGTM :+1:

src/geo/map.js
alonsogarciapablo: I'm not a fan of those names either, I just followed the convention that was there before.

src/geo/map.js
alonsogarciapablo: Yes, we are basically re-setting the order of each layer based on it's present order and grouping them.

xavijam: :+1:

src/geo/map.js
alonsogarciapablo: Yes, I can change it.

alonsogarciapablo:
@xavijam I addressed your feedback and fixed another bug. Could you please take a look at [this commit](https://github.com/CartoDB/cartodb.js/commit/39a48fa6d5d1bc8be87fae8ba0d37f62b6d723c7). Thanks!

alonsogarciapablo:
retest this please

PR #559, opened 2015-07-03
Adds describedate for implementing Torque on Pecan

fdansv:
@javierarce  this still needs to be combined with category and add weighing, but I'm not to sure on how to proceed with that, let's talk monday. For now this should add torque to the cards.


fdansv:
Merged manually to 4299-pecan-fixes

PR #571, opened 2015-07-08
Pecan

fdansv:
Pecan release, Jul 8th 2015

@javierarce 

cc @saleiva @javisantana


javierarce:
:+1: (except for our current lacking of specs)

PR #573, opened 2015-07-08
Update cartocss.js

andrewxhill:
Changed the default cartocss in pecan to FF6347 (tomato because miguel will be so happy) so it pops more on the light basemap.
i couldn't tell if you had intentionally made it different than the wizard default of 0F3B82 but i thought maybe you were making pecan be different, so i went with different too...

cc / @javierarce @saleiva @fdansv


javierarce:
Sorry, forgot this PR :(

:+1:

PR #584, opened 2015-07-10
Adds bubble maps to Pecan

### LINKED ISSUE #582: Bubble when number-based point columns
Right now we use choropleth for point tables with numeric columns. We should use bubble instead.

fdansv:
Closes #582, adds some cartocss tests.

@javierarce


javierarce:
:+1:

Issue #593, opened 2015-07-15
`cartodb.createLayer` must "return" the first "layer group" instead of the layer in position 1


alonsogarciapablo:
Right now [`cartodb.createLayer` takes for granted that the layer in position 1 of the viz.json contains a "layer group"](https://github.com/CartoDB/cartodb.js/blob/develop/src/api/layers.js#L99-L104). While this has been true for a long time, it's possible that this might change at some point (as we've seen when we introduced a layer with labels). Instead of getting the layer in position 1 by default, we should make sure that we select the first "layer group".

alonsogarciapablo:
Closed by #594.

PR #594, opened 2015-07-15
cartodb.createLayer selects the first data layer instead of assuming that it's in position 1

alonsogarciapablo:
Fixes: #593 

We're assuming that the layer in position #1 was a `layergroup` or `namedmap`. This PR implements a smarter way of looking for the first data layer.

cc/ @xavijam

### DIFF
--- a/src/api/layers.js
+++ b/src/api/layers.js
@@ -96,12 +96,21 @@
         if(visData.layers.length < 2) {
           promise.trigger('error', "visualization file does not contain layer info");
         }
-        var idx = options.layerIndex === undefined ? 1: options.layerIndex;
-        if(visData.layers.length <= idx) {
-          promise.trigger('error', 'layerIndex out of bounds');
-          return;
+        var index = options.layerIndex;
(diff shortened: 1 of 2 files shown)


src/api/layers.js
```diff
@@ -96,12 +96,21 @@
...
-          return;
+        var index = options.layerIndex;
+        if (index !== undefined) {
+          if(visData.layers.length <= index) {
+            promise.trigger('error', 'layerIndex out of bounds');
+            return;
+          }
+          layerData = visData.layers[index];
+        } else {
+          var DATA_LAYER_TYPES = ['namedmap', 'layergroup'];
```
xavijam: ['namedmap', 'layergroup'] Didn't we have those variables somewhere?

xavijam:
Sounds good!

src/api/layers.js
alonsogarciapablo: Probably! I'm going to find a better place for these guys. Thanks!

javisantana:
<3

PR #610, opened 2015-07-20
Fix setBounds/fitBounds for map w/o suitable zoom

viddo:
While working on https://github.com/CartoDB/cartodb/issues/4552 I found that setBounds wasn't working, because the internal code silently aborted, due to zoom level not being returned.

As with http://leafletjs.com/reference.html, `getBoundsZoom` now always returns a number (fallbacks on max zoom if could not found a suitable one), and enabled. fitBounds to also work since it do not have to check for the `null` value anymore.

Also updated docs to use `setBounds` instead since there's no clear reason for why fitBounds should be used.


viddo:
@javisantana can you review please?

javisantana:
:es:

PR #624, opened 2015-07-28
Removes occurrences of $.browser

### LINKED ISSUE #510: Stop using $.browser
It was removed in jQuery 1.9, so this makes using a custom jQuery impossible..

http://api.jquery.com/jquery.browser/

fdansv:
Closes #510

### DIFF
--- a/src/core/util.js
+++ b/src/core/util.js
@@ -89,11 +89,34 @@ cdb.core.util._makeCRCTable = function() {
   return crcTable;
 };
 
-cdb.core.util.getIEVersion = function(){
+cdb.core.util.browser = (function(){
+  var browser = {};
+  function detectIE() {
+    var ua = window.navigator.userAgent;
+    var msie = ua.indexOf('MSIE ');
+    var trident = ua.indexOf('Trident/');
+    var edge = ua.indexOf('Edge/');
+    if (msie > 0 || trident > 0 || edge > 0) return true;
+    return false;
+  };
+
+  function getIEVersion(){
     if (!document.compatMode) return 5
     if (!window.XMLHttpRequest) return 6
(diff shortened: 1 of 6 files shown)


javisantana:
I don't like the approach, I rather use:

- an explicit browser detection. For example, in the infowindow animation we disable IE because of performance. If we are going with those tricks, better to hide it behind a helper

- explicit feature detection for fullscreen for example

fdansv:
@javisantana added `cdb.core.util.browser`. Tested with all desktop browsers and all IE js engines from IE6 upwards.

src/core/util.js
```diff
@@ -87,4 +87,37 @@ cdb.core.util._makeCRCTable = function() {
...
+  };
+
+  if(detectIE()){
+    browser.ie = {version: getIEVersion()}
+  }
+
+  else if(ua.indexOf('Chrome') > -1) browser.chrome = ua;
+  else if(ua.indexOf('Firefox') > -1) browser.firefox = ua;
+  else if(ua.indexOf("Safari") > -1) browser.safari = ua;
+  else if(ua.toLowerCase().indexOf("op") > -1) browser.opera = ua;
```
javisantana: why in this case is different?

fdansv: Whoops, that's the first one I added, forgot to set it like the rest. Correcting it now.

test/spec/core/util.spec.js
```diff
@@ -0,0 +1,29 @@
+fdescribe("util", function() {
```
javisantana: is this fdescribe ok?

fdansv: no, but it's changed now

PR #635, opened 2015-08-03
Updates jquery plugins to latest versions

fdansv:
@javisantana is it important to keep the plugins in separate files? They seem to distribute them in a single file, not sure if that has changed from the past.


javisantana:
this review should be done by @xavijam, this is a pretty big change and we should check cartodb UI

xavijam:
Hey, let's hold our horses, this change has to be very well tested in the UI because it could collide with our current implementation. So my proposal is to check it with @fdansv in both sides.

javisantana:
closing until we decide something

PR #638, opened 2015-08-05
Bump the z-index value

javierarce:
From the issue description: 

> Sometimes the z-index of the annotations is greater than the z-index of the tooltip (which by default is 5).

> The solution I propose consists in increasing the z-index value of the tooltips to a number that guarantees that a map with a normal amount of annotations won't produce overlapping.

Original issue: CartoDB/cartodb#4872


javierarce:
Can you review this, @xavijam?

xavijam:
:+1:

PR #644, opened 2015-08-10
Timeslider tests fix

fdansv:
@jesusvazquez these are not depending on timezones anymore, shouldn't give you any problems.


jesusvazquez:
@fdansv it's now working. Thank you.

PR #652, opened 2015-08-14
Do not try to read secrets.json for every grunt task

### LINKED ISSUE #618: secrets.json not needed
there are some commands such

```
make cartodb dist/cartodb.css
```

where there isn't really needed a `secrets.json`, instead of preventing the command to execute, another strategy should be followed (like having an empty `secrets.development.json` and using an `env` variable in the makefile

alonsogarciapablo:
Fixes #618.

@CartoDB/frontend what do you guys think? It seems to be working :+1:

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -33,8 +33,7 @@ module.exports = function(grunt) {
       minor:      version[0] + '.' + version[1],
       bugfixing:  pkg.version
     },
-    pkg:  pkg,
-    secrets: grunt.file.readJSON('secrets.json')
+    pkg:  pkg
   };
 
   grunt.initConfig({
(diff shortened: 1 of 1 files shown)


viddo:
AFAIK, while this fixed the build, I believe the secrets file to be required for the publish build, so shouldn't the read/requirement of this file be moved closer to that area rather than just removed?

javisantana:
yes, secrets file should be used in deploy and cache invalidation

alonsogarciapablo:
@viddo Those check are actually [here](https://github.com/CartoDB/cartodb.js/blob/fix-dist-grunt-task/Gruntfile.js#L116-L118) and [here](https://github.com/CartoDB/cartodb.js/blob/fix-dist-grunt-task/Gruntfile.js#L153-L155).

viddo:
Ah! Then we're good to go, right?

PR #653, opened 2015-08-14
Added support for retina displays

acanimal:
This PR tries to bring support to high res devices. (Related with https://github.com/CartoDB/cartodb.js/issues/276)

The idea is to concat a new `cartodb.retina.css` file that defines device resolution media query for retina devices. To allow this I have modified the `grunt/tasks/concat.js` too.

Within the `cartodb.js` file I have defined a global `cdb.isRetinaBrowser()` function that checks if device is retina ready, inspecting the CSS `devicePixelRatio` value on the previous media query.
I saw in the `src/geo/commons.js` there is a logo class which implements a similar `isRetina()` method. I do not change the code here, but if the global function is accepted, the logo code would be to be updated to use it.

Finally, I have modified the `src/geo/leaflet/leaflet_cartodb_layergroup.js` file so the inherited `getTile()` takes into account if the device is retina ready. Note, I attach to the tiles URL a final `{r}.png` template variable. I think it would be better to set a different configuration in the `tilejson`.

It lacks support for GMaps (I'm not very experienced with it) but from what I can see it is required to update the `src/geo/gmaps/gmaps_tiledlayer.js` (the `getTileUrl()` function) and probably `src/geo/gmaps/gmaps_tlayergroup.js` (the `CartoDBLayerGroupBase.prototype.getTile()` method) to take into account the `{r}.png` variable too. I need more time to research on it.

### DIFF
--- a/grunt/tasks/concat.js
+++ b/grunt/tasks/concat.js
@@ -155,7 +155,7 @@ module.exports = {
             'vendor/mod/jquery-ui/jquery.ui.mouse.js',
             'vendor/mod/jquery-ui/jquery.ui.slider.js'
           ]
-        }  
+        }
       },
 
       odyssey: {
@@ -168,7 +168,7 @@ module.exports = {
           '<%= config.dist %>/cartodb.mod.odyssey.uncompressed.js': [
             'vendor/mod/odyssey.js'
           ]
-        }  
+        }
       },
 
       themes: {
@@ -176,6 +176,7 @@ module.exports = {
         files: {
           // CartoDB.js CSSs (themes?)
           '<%= config.dist %>/themes/css/cartodb.css': [
+            'themes/css/cartodb.retina.css',
             'themes/css/infowindow/*.css',
             'themes/css/map/*.css',
             'themes/css/tooltip/*.css',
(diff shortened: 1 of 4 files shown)


javisantana:
Hey, thanks, some questions/comments:

- how would you test this? specially running inside a headless browser 
- both google maps and leaflet implementations use [some source](https://github.com/CartoDB/cartodb.js/blob/develop/src/geo/layer_definition.js#L384) to get tile urls, do you think it could be done there instead of for each implementation?
- how this would work with interaction? (you know interaction is done with utfgrid)
- would torque layers need some change?
- why cartodb.retina.css is needed?

acanimal:
Ops, sorry not a perfect PR.

I tested in the old bad way (using a "demo" HTML. Maybe I can attach the file to `test/demos`?). I think it would be better from my part try to add a jasmine spec to test methods returns what they must return. In addition maybe a phantomJS test could be nice too to see tiles are loaded too?

I didn't check in depth all the code. It was a quickly and dirty job. I will check the code your refer at `src/geo/layer_definition.js` but seems a much better solution.

I need to check for interaction and torque. The first can be affected if tile size changes for @2x.

Thanks !!!

PR #661, opened 2015-08-28
Changed search ui component adding new functionalities

xavijam:
Normal:
search_pin

Mobile:
search_pin

- [x] Possibility to add a search pin + infowindow when geocoder result is back.
- [x] Zoom properly when a search is completed and result is a lat-lon (cc @iriberri).
- [x] Apply a new style.

screen shot 2015-08-28 at 16 56 39

REVIEWER @alonsogarciapablo 

Related -> https://github.com/CartoDB/cartodb/issues/4914

### DIFF
--- a/src/geo/ui/search.js
+++ b/src/geo/ui/search.js
@@ -206,12 +206,10 @@ cdb.geo.ui.Search = cdb.core.View.extend({
 
   _bindEvents: function() {
     this.mapView.bind('click', this._destroySearchPin, this);
-    // this.model.bind('change:center', this._destroySearchPin, this);
   },
 
   _unbindEvents: function() {
     this.mapView.unbind('click', this._destroySearchPin, this);
-    // this.model.unbind('change:center', this._destroySearchPin, this);
   },
 
   clean: function() {
--- a/src/geo/ui/search.js
+++ b/src/geo/ui/search.js
@@ -1,4 +1,3 @@
-
 /**
  *  UI component to place the map in the
  *  location found by the geocoder.
@@ -24,7 +23,7 @@ cdb.geo.ui.Search = cdb.core.View.extend({
   },
 
   options: {
-    searchPin: false,
+    searchPin: true,
     infowindowTemplate: '<div class="cartodb-infowindow">'+
     '<div class="cartodb-popup v2">'+
       '<a href="#close" class="cartodb-popup-close-button close">x</a>'+
--- a/src/geo/ui/search.js
+++ b/src/geo/ui/search.js
@@ -25,14 +25,15 @@ cdb.geo.ui.Search = cdb.core.View.extend({
   options: {
     searchPin: true,
     infowindowTemplate: '<div class="cartodb-infowindow">'+
-    '<div class="cartodb-popup v2">'+
+    '<div class="cartodb-popup v2 centered">'+
       '<a href="#close" class="cartodb-popup-close-button close">x</a>'+
        '<div class="cartodb-popup-content-wrapper">'+
          '<p>{{ address }}</p>'+
        '</div>'+
        '<div class="cartodb-popup-tip-container"></div>'+
     '</div>',
-    infowindowOffset: [28, 100],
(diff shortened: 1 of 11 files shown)


xavijam:
Also adding @viddo who gave the idea about BASE64 image (:D)

xavijam:
retest this please

src/geo/ui/search.js
```diff
@@ -1,31 +1,64 @@
...
+    "click input[type='text']": '_onFocus',
+    "submit form": '_onSubmit',
+    "click": '_stopPropagation',
+    "dblclick": '_stopPropagation',
+    "mousedown": '_stopPropagation'
   },
 
-  initialize: function() {},
+  options: {
+    searchPin: false,
```
xavijam: Should we show the search pin + infowindow by default?
@josecruz @iriberri @alonsogarciapablo

src/geo/ui/search.js
```diff
@@ -36,44 +69,155 @@ cdb.geo.ui.Search = cdb.core.View.extend({
...
+  _destroyPin: function() {
+    if (this._searchPin) {
+      this.mapView._removeGeomFromMap(this._searchPin);
+      delete this._searchPin;
+    }
+  },
+
+  _bindEvents: function() {
+    this.mapView.bind('click', this._destroySearchPin, this);
+    // this.model.bind('change:center', this._destroySearchPin, this);
```
alonsogarciapablo: We should removed these comments if code is no longer needed. Same in line 214.

src/geo/ui/search.js
```diff
@@ -1,31 +1,64 @@
 
+/**
+ *  UI component to place the map in the
+ *  location found by the geocoder.
+ *
+ */
+
```
alonsogarciapablo: Can we remove the spaces between the JSDocs and the class definition?

alonsogarciapablo:
Great work @xavijam :clap: . I tried the example and it works :+1:. My only comment is that perhaps it would make sense to leave the pin and only close the infowindow when the user clicks on the map. I think it could be helpful, but that's just my opinion and it can wait for real feedback.

src/geo/ui/search.js
alonsogarciapablo: Not sure about it.

iriberri: Not sure about this neither... making it configurable in the Editor requires viz.json changes, right? I don't think that enabling it by default could cause any actual problem for a user workflow. Probably the users that want this feature are the ones who are not able to use CartoDB.js to enable/disable it, so...

xavijam: @javisantana agree?

javisantana: this is @saleiva's decision: it should be always enabled by default, both in editor and embeds

src/geo/ui/search.js
xavijam: done

src/geo/ui/search.js
xavijam: Roger

xavijam:
About last @alonsogarciapablo comment, I don't know, what do you think @iriberri ^^^?

iriberri:
I think it's a really good idea. A very fast way to get rid of the infowindow that is something that contains information that the user already knows. :)

Actually if you allow me to be a little bit picky I'd prefer this infowindow to not look like our actual infowindows -- people might start complaining about why "the infowindow" does not look like their custom one now that I think about that. But I don't know how difficult to accomplish is this. I'd personally love a tooltip-like infowindow.

xavijam:
I agree, but it will take some time to create a new infowindow style. That's true that @josecruz would love that...

iriberri:
Actually I'm going to throw all my ideas now that we're dreaming a bit: a tooltip that is auto-hidden after ¿2 seconds?

xavijam:
Uhm... it could avoid user to add a point in that position within the editor.

screen shot 2015-08-28 at 16 49 57

PR #666, opened 2015-09-03
Display custom attributions

alonsogarciapablo:
These are the changes required to display the custom attributions that were added to the viz.json in cartodb/cartodb#5270.

@javier what do you think? [this](https://github.com/CartoDB/cartodb.js/compare/display-custom-attributions?expand=1#diff-c99f1bfa0704edaea17491896ed25ab7R302) is a bit hacky but it was the only way I found to set the attributions in the order I wanted (CartoDB attribution first and then the rest...). Thanks!


javierarce:
I think it's fine :+1:

Issue #228, opened 2014-12-01
don't use the same pointer for infowindow and tooltip


javisantana:
when in a map you have only a tooltip enabled (and not the infowindow) the cursor is the hand. That interaction is good when you can click on the polygon but not that good for hover.

cc @saleiva

javisantana:
@saleiva please, this is horrible

saleiva:
Agree, what about simple pointer for hover?

nobuti:
There has been no activity on this issue for several months and that's why are closing it (automatically). If you think this still needs to be addressed please reopen it. The good news is that we are now working on the next version of CARTO.js, stay tuned.

PR #672, opened 2015-09-09
adding different cursor for tooltip

oriolbx:
Solves https://github.com/CartoDB/cartodb.js/issues/228

When the tooltip/hover is enabled, the cursor "nw-resize" is added. When tooltip and infowindow options are enabled, the cursor "pointer" (the cursor of the infowindow) is established. 

@javierarce

### DIFF
--- a/src/vis/vis.js
+++ b/src/vis/vis.js
@@ -1152,6 +1152,32 @@ var Vis = cdb.core.View.extend({
         }
       });
     }
+
+    var self = this; 
+
+    if (layerView.tooltip){
+      
+      layerView.bind('mouseover', function() {
+        
+        self.mapView.setCursor('nw-resize'); 
+      });
+      layerView.bind('mouseout', function() {
+        
+        self.mapView.setCursor('auto'); 
+      }); 
+    }
+
+    if (layerView.infowindow){
+      layerView.bind('mouseover', function() {
+          
+          self.mapView.setCursor('pointer'); 
+        });
+        layerView.bind('mouseout', function() {
+          
+          self.mapView.setCursor('auto'); 
+        }); 
+
+    }
   },
 
   addInfowindow: function(layerView) {
--- a/src/vis/vis.js
+++ b/src/vis/vis.js
@@ -457,7 +457,7 @@ var Vis = cdb.core.View.extend({
     }, this);
 
     this.mapView.bind('newLayerView', this._addLoading, this);
-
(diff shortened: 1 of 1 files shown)


javisantana:
ping @javierarce

src/vis/vis.js
```diff
@@ -1152,6 +1152,32 @@ var Vis = cdb.core.View.extend({
         }
       });
     }
+
+    var self = this; 
+
+    if (layerView.tooltip){
+
```
javierarce: Please, check the style of the code. There are several empty lines that should be removed (have a look to our JavaScript style guide wiki page)

src/vis/vis.js
```diff
@@ -1152,6 +1152,32 @@ var Vis = cdb.core.View.extend({
...
+        
+        self.mapView.setCursor('nw-resize'); 
+      });
+      layerView.bind('mouseout', function() {
+        
+        self.mapView.setCursor('auto'); 
+      }); 
+    }
+
+    if (layerView.infowindow){
```
javierarce: Should this condition be here, inside the `addTooltip` method?

javierarce: Also, we already have mouse bindings here: https://github.com/CartoDB/cartodb.js/blob/228-change_pointer_tooltip/src/vis/vis.js#L1276 https://github.com/CartoDB/cartodb.js/blob/228-change_pointer_tooltip/src/vis/vis.js#L1280… why the repetition?

alonsogarciapablo:
What about something like the following?

```
load: function(data, options) {
  // ...
  this.mapView.bind('newLayerView', this._setCursor, this);
  // ...
}

_setCursor: function(layerView) {
  var self = this;
  var mouseoverCursor = 'pointer';
  var mouseoutCursor = 'auto'
  if (layerView.tooltip && !layerView.infowindow){
    mouseoverCursor = 'nw-resize';
  }
  layerView.bind('mouseover', function() {
    self.mapView.setCursor('nw-resize'); 
  });
  layerView.bind('mouseout', function() {
    self.mapView.setCursor('auto'); 
  }); 
}
```

We could then get rid of [these lines](https://github.com/CartoDB/cartodb.js/blob/develop/src/vis/vis.js#L1250-L1256)... Probably we could find  a better name for `_setCursor`. That method is basically binding mouse events to functions that change the cursor of the map... We could also call it something like `_bindMouseEvents`.

nobuti:

PR #673, opened 2015-09-09
Handle map attributions a bit better

alonsogarciapablo:
This PR tries to improve/simplify the way we handle attributions for maps. Now the Map model listens to changes on its layers and updates the `attribution` attribute and updates its attribution accordingly (removing empty attributions, duplicates, adding the default CartoDB attribution, etc.).

Also, when a MapView is instantiated, attributions are "refreshed", so the map renders the right attributions if the map model already have some layers loaded (this is what happens in the editor).

@javierarce could you please take a look? thanks!


javierarce:
Nice refactor! :+1:

PR #674, opened 2015-09-10
Add checker to fullscreen button when it is rendered in an iframe

xavijam:
The problem is iframe content can't be "fullscreen-ed" if the origin/parent url doesn't have a https protocol. And this possible solution would be:

- Check if the map is iframed.
- If so, checks if the parent of the iframe has http protocol.
- If it doesn't have https protocol, fullscreen button is not rendered.

Related:

- Bug in CartoDB: https://github.com/CartoDB/cartodb/issues/4543 
- Happening here: http://mashable.com/2015/07/17/schumer-thrones-google-emmys/#vVL1wykkAZk9
- Explanation: https://sites.google.com/a/chromium.org/dev/Home/chromium-security/deprecating-powerful-features-on-insecure-origins 

Using some keys from [here](http://stackoverflow.com/questions/935127/how-to-access-parent-iframe-from-javascript).

Sorry but I'd like to warn all @CartoDB/frontend team.

### DIFF
--- a/src/geo/ui/fullscreen.js
+++ b/src/geo/ui/fullscreen.js
@@ -112,7 +112,8 @@ cdb.ui.common.FullScreen = cdb.core.View.extend({
 
   _isFullScreenEnabled: function() {
     // If frameElement exists, it means that the map
-    // is embebed as an iframe so we need
+    // is embebed as an iframe so we need to check if
+    // the parent has a secure protocol
     var frameElement = window && window.frameElement;
     if (frameElement) {
       var parentWindow = this._getFramedWindow(frameElement);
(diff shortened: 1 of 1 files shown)


xavijam:
I've added a notification just to know what is happening, do you think it is worth it?

javierarce:
Yeah, I think the log is nice.

alonsogarciapablo:
Could we hide the fullscreen button in this situation?

xavijam:
Yes, it will not render the button.

alonsogarciapablo:
:+1:

matallo:
my only concern is with the naming, if I understand correctly it would better be `_canFullScreenBeEnabled`

this is a silly detail, apart from that :+1:

xavijam:
Agree matallo :)

xavijam:
You can test the difference:

Over https: https://dl.dropboxusercontent.com/u/931536/fullscreen/example.html
Over http: http://com.cartodb.users-assets.staging.s3.amazonaws.com/development/example.html

PR #677, opened 2015-09-14
Do not rely on object.offsetParent when the map is fixed to the top.

### LINKED ISSUE #639: Bug in IE and Firefox when map div is position fixed
This is the same bug as reported here: https://github.com/CartoDB/cartodb.js/issues/114

It now works in Chrome and Safari, but there is still an offset bug in Firefox and IE 8, 9, 10 and Edge. 

To reproduce the bug (using CartoDB.js 3.15.1):

1. Go to this page: http://thematicmapping.org/playground/cartodb/bugs/fixedpos.html
2. Scroll down the page (most of our maps are embedded in news articles).
3. Click the fullscreen button. We're using a fixed positioned div to show the map in fullscreen. 
4. Click on a region, the popup is shown far off the map and not where the user clicked. 
5. Popups in Leaflet works as expected in all browsers, so this is a CartoDB.js bug. 

screen shot 2015-08-05 at 14 07 08
 
Hope this bug will be fixed across all browsers and devices :-)

alonsogarciapablo:
Fixes #639.

In some circumstances, the map can be fixed to the top but the page might have some scroll that we need to take into account. In these cases it's better not to use the object.offsetParent approach.

@xavijam what do you think?

### DIFF
--- a/src/geo/leaflet/leaflet_cartodb_layergroup.js
+++ b/src/geo/leaflet/leaflet_cartodb_layergroup.js
@@ -277,18 +277,21 @@ L.CartoDBGroupLayerBase = L.TileLayer.extend({
       y = o.e.clientY;
     }
 
-    if (obj.offsetParent) {
-      // Modern browsers
+    // If the map is fixed at the top of the window, we can't use offsetParent
+    // cause there might be some scrolling that we need to take into account.
+    if (obj.offsetParent && obj.offsetTop > 0) {
       do {
         curleft += obj.offsetLeft;
         curtop += obj.offsetTop;
       } while (obj = obj.offsetParent);
-      return map.containerPointToLayerPoint(new L.Point(x - curleft, y - curtop));
+      var p = new L.Point(
+        x - curleft, y - curtop);
+      return map.containerPointToLayerPoint(p);
     } else {
       var rect = obj.getBoundingClientRect();
       var p = new L.Point(
-            (o.e.clientX? o.e.clientX: x) - rect.left - obj.clientLeft - window.scrollX,
-            (o.e.clientY? o.e.clientY: y) - rect.top - obj.clientTop - window.scrollY);
+        (o.e.clientX? o.e.clientX: x) - rect.left - obj.clientLeft - window.scrollX,
+        (o.e.clientY? o.e.clientY: y) - rect.top - obj.clientTop - window.scrollY);
       return map.containerPointToLayerPoint(p);
     }
   }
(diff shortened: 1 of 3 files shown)


alonsogarciapablo:
As a bonus, I discovered an fixed [this other bug](https://github.com/CartoDB/cartodb.js/commit/6d916ab1b3a2a506eb82e42aa7af09e1db0cc613).

alonsogarciapablo:
@xavijam do you think we should fix this [here](https://github.com/CartoDB/cartodb.js/blob/develop/src/geo/gmaps/gmaps_cartodb_layergroup.js#L232-L254) too? Thanks!

alonsogarciapablo:
Retest this please

alonsogarciapablo:
retest this please

jesusvazquez:
retest this please

xavijam:
@alonsogarciapablo, yes, we should fix there too.
About the PR en general, I love the idea (and the fix) but probably we should take care of GMaps too, shouldn't we?

xavijam:
Also, could we create a test trying this with a fixed or a relative map div?

alonsogarciapablo:
@xavijam Thanks for your feedback! I've [fixed it for GMaps](https://github.com/CartoDB/cartodb.js/commit/b65d0a881115ef8767017d2be63851b5600ee002) and also uploaded a [basic example that reproduces the issue](https://github.com/CartoDB/cartodb.js/commit/08f85c2efc72393c466a8a5116894f5bbb121f59).

xavijam:
yay, have you tested for old browsers (IE7-9)? Just to keep legacy compatibility :S.

src/geo/leaflet/leaflet_cartodb_layergroup.js
```diff
@@ -277,22 +278,32 @@ L.CartoDBGroupLayerBase = L.TileLayer.extend({
...
+        (o.e.clientY? o.e.clientY: y) - rect.top - obj.clientTop - scrollY);
     }
-  }
+    return map.containerPointToLayerPoint(point);
+  },
 
+  /**
+   * Creates an instance of a Leaflet Point
+   */
+  _newPoint: function(x, y) {
```
javisantana: why this abstraction?

alonsogarciapablo: I initially thought that this method was exactly the same for leaflet and gmaps, except for the instantiation of new points. I basically tried to make both methods look more alike.

PR #679, opened 2015-09-14
Select torque layers if no "cartodb" layers are present

### LINKED ISSUE #678: Can't load Torque layer from viz.json in createLayer CartoDB.js 3.15.
Today a user reported that an application stopped working. After investigating a bit, I was able to reproduce the issue, and the code to do it is here: https://gist.github.com/iriberri/3e3761496dc83d0b312d

By using a viz.json from a Torque visualization (there are no more layers than the Torque one), createLayer and CartoDB.js 3.15, the layer is not loaded. It works with 3.14 though.

I have tried to change to several basemaps, in case this was something due to how basemaps are written in the viz.json due to sandwiching, but different basemap approaches didn't have any effect in 3.15 neither in 3.14 (in which everything worked with any basemap).

alonsogarciapablo:
Fixes #678. [This commit](https://github.com/CartoDB/cartodb.js/commit/30dca98e51defe0c171a92c24f4d400d4bf87eb3) changed how we select the layer that must be used when `createLayer` doesn't specify an index, and torque layers were not taken into account.

@xavijam easy one...

cc: @iriberri

### DIFF
--- a/src/api/layers.js
+++ b/src/api/layers.js
@@ -104,7 +104,7 @@
           }
           layerData = visData.layers[index];
         } else {
-          var DATA_LAYER_TYPES = ['namedmap', 'layergroup'];
+          var DATA_LAYER_TYPES = ['namedmap', 'layergroup', 'torque'];
 
           // Select the first data layer (namedmap or layergroup)
           layerData = _.find(visData.layers, function(layer){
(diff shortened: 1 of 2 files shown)


iriberri:
This is great Pablo!

alonsogarciapablo:
retest this please

xavijam:
:us:

PR #682, opened 2015-09-16
Fixed two issues with attributions in Leaflet maps

### LINKED ISSUE #681: CartoDB.js 3.15.5: Uncaught TypeError: Cannot set property '_attributions' of undefined
If a CartoDB is used over a Leaflet map in which the `attributionsControl` (more info [here](http://leafletjs.com/reference.html#control-attribution)) is set to false:

```js
map = L.map("map", {
  zoom: 11,
  center: [28.175278, -81.370102],
  layers: [ggl2],
  zoomControl: false,
  attributionControl: false
});
````

CartoDB.js raises the following error:

`Uncaught TypeError: Cannot set property '_attributions' of undefined`

This scenario happens in a use case in which the user is using GMaps + Leaflet at the same time. As the GMaps attribution already exists, the user decided to remove the Leaflet one.

alonsogarciapablo:
Fixes #681.

1. Leaflet maps might have `attributionControl` set to false and CartoDB.js should still handle attributions correctly.
2. Leaflet maps might have some other layers with attributions, and these shouldn't be removed.

@fdansv can you please take a look?

cc: @iriberri

### DIFF
--- a/src/geo/leaflet/leaflet.js
+++ b/src/geo/leaflet/leaflet.js
@@ -282,14 +282,35 @@
     },
 
     setAttribution: function() {
+      var attributionControl = this._getAttributionControl();
 
-      // Attributions have already been set but we override them with
-      // the ones in the map object that are in the right order and include
-      // the default CartoDB attribution
-      this.map_leaflet.attributionControl._attributions = {};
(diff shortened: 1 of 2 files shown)


fdansv:
Awesome, nice job with that test!

:ship: :it:

PR #686, opened 2015-09-21
If fullscreen is not possible, open the map in a new tab

xavijam:
Apart from I introduced a bug last time I changed the fullscreen functionality (god), after talking with @iriberri and @oriolbx, if the fullscreen behaviour is not possible, we provide the possibility to open the map in a new tab, how does it sound? This could only happen when the map is added as an iframe.

REVIEWER: @alonsogarciapablo

### DIFF
--- a/src/geo/ui/fullscreen.js
+++ b/src/geo/ui/fullscreen.js
@@ -78,14 +78,13 @@ cdb.ui.common.FullScreen = cdb.core.View.extend({
   },
 
   render: function() {
-    var $el = this.$el;
     var options = _.extend(
       this.options,
       {
         mapUrl: location.href || ''
       }
     );
-    $el.html(this.options.template(options));
+    this.$el.html(this.options.template(options));
 
     if (!this._canFullScreenBeEnabled()) {
       this.undelegateEvents();
--- a/src/geo/ui/fullscreen.js
+++ b/src/geo/ui/fullscreen.js
@@ -67,10 +67,8 @@ cdb.ui.common.FullScreen = cdb.core.View.extend({
         }
       }
 
-      if (mapView) {
-        if (this.model.get("allowWheelOnFullscreen")) {
-          mapView.options.map.set("scrollwheel", true);
-        }
+      if (mapView && this.model.get("allowWheelOnFullscreen")) {
+        mapView.map.set("scrollwheel", true);
       }
     } else {
       cancelFullScreen.call(doc);
--- a/src/geo/ui/fullscreen.js
+++ b/src/geo/ui/fullscreen.js
@@ -14,47 +14,33 @@ cdb.ui.common.FullScreen = cdb.core.View.extend({
   className: 'cartodb-fullscreen',
 
   events: {
-
     "click a": "_toggleFullScreen"
-
   },
 
   initialize: function() {
-
     _.bindAll(this, 'render');
     _.defaults(this.options, this.default_options);
-
-    //this.model = new cdb.core.Model({
-      //allowWheelOnFullscreen: false
-    //});
-
     this._addWheelEvent();
-
   },
 
   _addWheelEvent: function() {
+    var self    = this;
+    var mapView = this.options.mapView;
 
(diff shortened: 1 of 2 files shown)


xavijam:
retest this please

alonsogarciapablo:
Taking a look!

xavijam:
Thanks buddy!

src/geo/ui/fullscreen.js
```diff
@@ -84,57 +68,49 @@ cdb.ui.common.FullScreen = cdb.core.View.extend({
...
   },
 
   render: function() {
-    if (this._canFullScreenBeEnabled()) {
-      var $el = this.$el;
-      var options = _.extend(this.options);
-      $el.html(this.options.template(options));
-    } else {
-      cdb.log.info('FullScreen is deprecated on insecure origins. See https://goo.gl/rStTGz for more details.');
+    var $el = this.$el;
```
alonsogarciapablo: This variable is only being used once. We could use this.$el directly in line #88.

src/geo/ui/fullscreen.js
```diff
@@ -84,57 +68,49 @@ cdb.ui.common.FullScreen = cdb.core.View.extend({
...
+    var options = _.extend(
+      this.options,
+      {
+        mapUrl: location.href || ''
+      }
+    );
+    $el.html(this.options.template(options));
+
+    if (!this._canFullScreenBeEnabled()) {
+      this.undelegateEvents();
```
alonsogarciapablo: I didn't know about this... :+1:

alonsogarciapablo:
:+1: It's a nice workaround.

xavijam:
thanks! :cactus:

xavijam:
I'd like to cc @javisantana too.

src/geo/ui/fullscreen.js
```diff
@@ -14,47 +14,33 @@ cdb.ui.common.FullScreen = cdb.core.View.extend({
...
-      $(document).on('webkitfullscreenchange mozfullscreenchange fullscreenchange', function() {
-
-        if ( !document.fullscreenElement && !document.webkitFullscreenElement && !document.mozFullScreenElement && !document.msFullscreenElement) {
-          if (self.model.get("allowWheelOnFullscreen")) {
-            mapView.options.map.set("scrollwheel", false);
-          }
+    $(document).on('webkitfullscreenchange mozfullscreenchange fullscreenchange', function() {
+      if (!document.fullscreenElement && !document.webkitFullscreenElement && !document.mozFullScreenElement && !document.msFullscreenElement) {
+        if (self.model.get("allowWheelOnFullscreen")) {
+          mapView.options.map.set("scrollwheel", false);
```
javisantana: mapView.model does not work?

javisantana:
Another option to fix this would be to do it outside the control so if the page is not allowed to have fullscreen just not add it. 

In this case the control itself is the one dealing with the document logic.

fine with me

src/geo/ui/fullscreen.js
xavijam: I don't think so, ```mapView.map``` should work.
https://github.com/CartoDB/cartodb.js/blob/develop/src/geo/map.js#L584
In any case this is inherited code.

xavijam:
We were removing the component before and some people got angry because they wanted their buttons back.

javisantana:
¿? if it's not allowed why people get angry? you can't fix that so

xavijam:
As we mentioned to the users (cc @iriberri and @oriolbx) it is something we can't fix because is a browser decision. In any case, as I said at the beginning, the behaviour I implemented was not working properly in all cases.
Checking the feedback we received when that change was released and having to fix it, we talked about a new solution, taking into account 1 - fix the problem we introduced, 2 - avoiding to remove the button if FullScreen API is not allowed.
@iriberri @oriolbx and I thought in this solution, but if we don't agree, it is ok. That's why I wanted you to take part of this issue.

PR #647, opened 2015-08-12
undefine define to allow amd collisions

javisantana:

PR #687, opened 2015-09-22
Undefine global `define` so that depencies (Mustache) are not loaded via AMD

### LINKED ISSUE #543: Uncaught ReferenceError: Mustache is not defined
We're not able to upgrade our mapping stack to use a later version of cartodb.js due to a conflict with mustache. Previously we've used cartodb_noleaflet.js (created with grunt build) 

This file now returns this error: 
Uncaught ReferenceError: Mustache is not defined

It seems like mustache is not included in the build file. This problem is related to this issue: https://github.com/CartoDB/cartodb.js/issues/408

We need to load our mapping components with require.js, and loading cartodb.js and/or mustache in a script tag is not an option.

alonsogarciapablo:
Fixes #543.

It was previously fixed [here](https://github.com/CartoDB/cartodb.js/pull/647) but we are not using `make` to generate the dist files of cartodb.js so dist files didn't include the fix. `grunt/templates/wrapper_header.js` is the header that we add to all of the dist js files...

@fdansv what do you think? Thanks!

cc: @turban


fdansv:
:ship: :it:

PR #693, opened 2015-09-30
remove wax.connector

eczajk1:
Removed implementation of wax.g.connector, which was creating it's own options object, overwriting the existing one.

This implementation just adds tiles, schemes, and blankImage to the existing options object.  The rest of the properties are assigned in the same way.


nobuti:

Cartofante:

PR #711, opened 2015-10-14
Removed background-clip properties from legends

xavijam:
Working on https://github.com/CartoDB/cartodb/issues/5206, I've noticed that background-clip property is not well managed by html2canvas. Example: http://jsfiddle.net/njLnwrjd/.

Do you think @javierarce it is needed?

REVIEWER: @javierarce

### DIFF
--- a/test/spec/core/image.spec.js
+++ b/test/spec/core/image.spec.js
@@ -23,15 +23,11 @@ describe("Image", function() {
   });
 
   it("should use the basemap defined in the vizjson", function(done) {
-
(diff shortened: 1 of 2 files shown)


xavijam:
retest this please

javierarce:
We used `background-clip` to have transparent borders, so removing it would only affect the style.

xavijam:
Ok, I didn't notice any difference without those styles. So :+1:?

javierarce:
Yep, let's go for it.

Issue #138, opened 2014-10-01
add openstreetmap geocoder for search


javisantana:
https://github.com/smeijer/L.GeoSearch/blob/master/src/js/l.geosearch.provider.openstreetmap.js

saleiva:
This should also support the gmaps option when using gmaps, right?

nobuti:
There has been no activity on this issue for several months and that's why are closing it (automatically). If you think this still needs to be addressed please reopen it. The good news is that we are now working on the next version of CARTO.js, stay tuned.

PR #721, opened 2015-10-16
add_openstreetmap_geocoder

oriolbx:
Solves #138 
add Nominatim geocoder in geocoder.js and change the search.js to allow
the option to use the Nominatim.

### DIFF
--- a/src/geo/ui/search.js
+++ b/src/geo/ui/search.js
@@ -11,6 +11,8 @@ cdb.geo.ui.Search = cdb.core.View.extend({
   _ZOOM_BY_CATEGORY: {
     'building': 18,
     'postal-area': 15,
+    'residential':18,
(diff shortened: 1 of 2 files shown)


src/geo/ui/search.js
```diff
@@ -82,11 +84,23 @@ cdb.geo.ui.Search = cdb.core.View.extend({
     this._showLoader();
     // Remove previous pin
     this._destroySearchPin();
-    cdb.geo.geocoder.NOKIA.geocode(address, function(places) {
-      self._onResult(places);
-      // Hide loader
-      self._hideLoader();
+    var geocoder = "cdb_Nokia";
```
javisantana: how do i change the geocoder from outside?

nobuti:

PR #724, opened 2015-10-21
Some refactorings

alonsogarciapablo:
@xavijam PTAL


src/geo/layer_definition.js
```diff
@@ -2,21 +2,10 @@
 function MapBase(options) {
```
alonsogarciapablo: This file should not be used anymore, so we should be able to delete it pretty soon.

src/windshaft/client.js
```diff
@@ -26,25 +26,22 @@ cdb.windshaft.Client.MAX_GET_SIZE = 2033;
...
-  // TODO: We will need to instantiate a new cdb.windshaft.PublicMap() here
-  var windshaftMap = new cdb.windshaft.PublicMap();
+  var dashboardInstance = new cdb.windshaft.DashboardInstance();
 
   var options = {
     success: function(data) {
       if (data.errors) {
-        callback(null, data);
+        // TODO: Error handling
+        throw data.errors;
```
alonsogarciapablo: I need to improve this. @xavijam any ideas on how to handle errors here?

src/vis/vis.js
```diff
@@ -471,57 +461,42 @@ var Vis = cdb.core.View.extend({
...
-        var v = self.addWidget(
-          d.type,
-          _.extend(
-            opts,
-            {
-              datasource: self.datasource
-            }
-          )
-        );
+          var widgetModel = new cdb.geo.ui.Widget.ListModel(widgetData);
```
xavijam: Could we change this in order to take the proper widget model and widget view?
@javier is adding histogram widget right now :D.

alonsogarciapablo: Done [here](https://github.com/CartoDB/cartodb.js/commit/ecde1c4a3b498ca7a5b87605539c56c2c0f9539f). We can move that logic somewhere else.

src/windshaft/client.js
xavijam: don't you like callbacks system?

alonsogarciapablo:
I can use callbacks. But the thing is.. what's going to happen to the UI when this request fails? For example if a users tries to set some invalid SQL for a layer and Windshaft can't create a new instance of the dashboard, should we just throw an error in the console and leave the map/widgets as they were?

xavijam:
Uhm... yeah, I think we should send the 'error' event and, of course, we should display that error in the widget... right?

xavijam:
Go man go!

Issue #722, opened 2015-10-19
Cleanup post-add-browserify


viddo:
:fire: to take place after #699:, to be removed once verified not actually needed:
- [x] concat task, nojquery/noleaflet no long needed (already done in 6ae2335b3adccaf96ceb5155033779fc657b9262)
- [x] `index.js` old rails manifest?
- [x] `src-browserify/loader.js`
- [x] `geo/ui/switcher*`
- [x] `test/lib`, `sinon` no longer used in new specs
- [x] Rename `_Promise` to `Promise`
- [x] Review files/classes that declara plain Backbone.Model and Backbone.Collection (eg: [geo/geometries.js](https://github.com/CartoDB/cartodb.js/blob/v4-browserify-standard2/src-browserify/geo/geometries.js))
- [x] `package.json` browser/browserify-shim definitions, all used?
- [x] `Makefile`, really necessary nowadays? 
- [x] `scripts` dir, e.g. `scripts/compress`, only be used for cartodbeditor; can be removed/moved?
- [x] `grunt/`, e.g. templates used in core concat, no longer needed (e.g. underscore_no_conflict, version_header etc.).
- [x] `vendor/reqwest` (unused since not supporting core or nojquery bundles anymore)
- [x] `vendor/mod/torque.uncompressed`, now required through torque.js npm bundle?

Out-of-scope/postponed:
- ~~rm `test/demos/`~~ :point_right: #715
- ~~`src/` + `test/spec` (but not `src-browserify`)~~ :point_right: #758
- CartoDB editor also will need to be updated at some point :point_right: cartodb/cartodb#6068

viddo:
closed in #763

PR #725, opened 2015-10-21
new docs: split files

matallo:
I've splitted the `API.md` in several files in order to have them better work with new docs

does this structure fits well?

also, I've spotted some legacy (?) build tasks and files:

- `Makefile`
- `doc_template.html`
- `style.css`
- `thumbnail.png`

shall we get rid of these?

CR @alonsogarciapablo


doc/other_stuff.md
```diff
@@ -0,0 +1,162 @@
...
+sql_api_template: 'https://{user}.test.com'
+{% endhighlight %}
+
+CartoDB.js will replace ``{user}``.
+
+Notice that you don't need to set the path to the endpoint, CartoDB.js will set it automatically.
+
+## Bounds wrapper
+
+We have added an easy method to get the bounding box for any dataset or filtered query using the CartoDB.js library. The **getBounds** function can be useful for guiding users to the right location on a map or for loading only the right data at the right time based on user actions.
```
alonsogarciapablo: This is also documented [here](https://github.com/CartoDB/cartodb.js/pull/725/files#diff-363d4745c1355c9340fcee8e9791e493R56). Perhaps we can remove this and complete the description (if needed on the other file)

doc/other_stuff.md
```diff
@@ -0,0 +1,162 @@
+# Other important stuff
```
alonsogarciapablo: Not sure if reviewing the docs should also be part of this PR but I think we could try to get rid of this whole file and move its content to the other sections.

matallo: We definitely need to review the docs, at least to check the styles are rendering properly

/cc @csobier

alonsogarciapablo:
Structure makes sense since it matches the existing section. I left a comment about trying to get rid of the `other_stuff.md` file. About the "legacy" files, they are pretty old, so I would :fire:. Good catch!

:+1:

doc/other_stuff.md
matallo: removed here https://github.com/CartoDB/cartodb.js/commit/581dd968cfe4c5e24d23c23321d16326c17b7597

@csobier extend the description if needed

matallo:
@csobier what about `raster_howto.md`, shall we find some place for it in the docs or ditch it completely?

I'm leaving it for now

matallo:
the same with `metrics.md`, this isn't covered in the public docs, what should we do here @csobier

matallo:
created specific issues for those
https://github.com/CartoDB/docs/issues/449
https://github.com/CartoDB/docs/issues/448

viddo:
re: the Makefile I've planned to review it in [#722 cleanup post-add-browserify](https://github.com/CartoDB/cartodb.js/issues/722), but if it can just be removed earlier on better (as long as it doesn't imply changes to [gruntfile.js](https://github.com/CartoDB/cartodb.js/blob/v4-browserify/Gruntfile.js) which I already modified quite a bit).

matallo:
nah, it's just removing non-used build scripts and files, I won't touch the [gruntfile.js](https://github.com/CartoDB/cartodb.js/blob/v4-browserify/Gruntfile.js)

matallo:
also, I spotted a `quickstart.md` that could easily be removed, can you confirm @alonsogarciapablo @viddo @csobier

alonsogarciapablo:
:fire:

matallo:
:sparkles:

Issue #699, opened 2015-10-06
Add browserify


xavijam:
_edit: Added scope & planning, will be updated continuously  /@viddo_

Considering the browserify'ing of the source code will imply a lot of changes to existing source, we want to avoid epic pull-requests and changes that are hard to control. 
Thus, the following approach will be taken:
- Create separate "baseline" branch (`v4-browserify`) while this issue is in WIP (branch off `v4`), copy src+specs to the new dir structure so PRs against this branch shows as changes and not new files
- Branch of this baseline branch for each significant change, and create pull-request to merge back to the WIP baseline (i.e. `v4-browserify`)
- Each source file will be copied to a new location (e.g. `src-browserify/`), and modified from there, instead of trying to modify the source right away
- Once all code is migrated, reviewed & ready, merge `v4-browserify` to `v4`

**Also, in general:**
- Any eventual changes happening on `v4` meanwhile will be updated/merged to `v4-browserify` manually
- Use npm packages instead of manually have to keep track of vendor files (where possible)
- One concept/class/object per file, following node/commonjs conventions/best-practices
- But maintain current directory layout, to ease understanding of where move code went
- Avoid modify any existing code unless required to due to the re-organization

**Tasks:**
- [x] Setup boilplate; browserify build, tests integration etc. :point_right: #723
- [x] ~~Migrate core bundle #723~~, removed in :point_down: #738 
- [x] Migrate standard bundle (cartodb.js) :point_right: #738 
- [x] fix banner bug screwing up source-maps b59e684
- [x] Migrate torque bundle (cartodb.mod.torque.js) :point_right: #754
- [x] Migrate odyssey bundle  (cartodb.mod.odyssey.js) :point_right: #755
- [x] Update cdb.VERSION on build :point_right: 295ee6d
- [x] Support source-maps in minimized dist file; add [exorcist](https://github.com/thlorenz/exorcist) to extract source maps before uglify task :point_right: #757
- [x] ~~Add D3.js? :point_right: #718 / #719~~ follow [browserify idioms instead, postponed](https://github.com/CartoDB/cartodb.js/pull/719#issuecomment-153418701)
- [x] cleanup after all bundles are migrated :point_right: #722
- [x] fix torque loading jquery from cartodb.js instead of containing it itself
- [x] update `develop` -->  `v4` --> `v4-browserify`, migrate any changes done initially branch off
- [x] merge `v4-browserify` to `v4`

---

[...]

PR #727, opened 2015-10-21
Part 1 migration of standard bundle

viddo:
Part milestone for #699

:point_up: [diff excluding whitespace](https://github.com/CartoDB/cartodb.js/pull/727/files?w=0) to avoid noise.

Noteworthy changes to keep in mind for doing the code review:
- Migrated the [core files used](https://github.com/CartoDB/cartodb.js/blob/4eb21226c1730797e1d35b220908c99844b7fffa/src/cartodb.js#L52-L61) used in standard bundle (+related vendor files)
- Wrapped classes/objects to make implicit and varying dependencies explicit, to be wired up in the bundle entry file. To give example of how it actually works see this small fictional example:
```js
// my-model.js
var Backbone = require('Backbone'); // doesn't vary so require here
module.exports = function(opts) {
  var ajax = opts.$ajax || opts.reqwest;
  var MyModel = Backbone.Model.extend(…); // uses ajax for something in the model
  return MyModel;
};
```
```js
// standard.js
var setupMyModel = require('./my-model')
var MyModel = setupMyModel(window.$.ajax); // assumes jquery being loaded already
```
```js
// core.js
var reqwest = require('reqwest');
var setupMyModel = require('./my-model');
var MyModel = setupMyModel(reqwest);
```
  - This has several benefits which are important for upcoming changes:
    - Gives an easy overview of what dependencies there are
    - Will make it easier to refactor/untangle the code since the wiring of dependencies are done in a single place
    - standard bundle is essentially the same as for the no-jquery and no-leaflet too (with the obvious differences the bundle names implies), so this approach will make it trivial to implement those bundles later

Also:
- pay attention to the `// TODO`s, added them for things I wanted to highlight in particular


src-browserify/core/log.js
```diff
@@ -1,84 +1,54 @@
...
-        }
-    });
-
-    /** contains all error for the application */
-    cdb.errors = new cdb.core.ErrorList();
-
-
-    // error tracking!
-    if(cdb.config.ERROR_TRACK_ENABLED) {
-      cdb.errors.enableTrack();
```
viddo: `cdb.config.ERROR_TRACK_ENABLED` is always set to false, so this code path is unreachable, so :fire: it

src-browserify/core/model.js
```diff
@@ -1,23 +1,15 @@
...
-    var callbacks = o._callbacks;
-    for(var i in callbacks) {
-      var node = callbacks[i];
-      console.log(" * ", i);
-      var end = node.tail;
-      while ((node = node.next) !== end) {
-        console.log("    - ", node.context, (node.context && node.context.el) || 'none');
-      }
-    }
-  }
```
viddo: couldn't find any usage of `cdb._debugCallbacks` so :fire: 
let me know if I should put it back

viddo:
@xavijam pretty please review :kiss:

grunt/tasks/concat.js
```diff
@@ -28,31 +28,6 @@ module.exports = {
 
     return {
```
xavijam: Goodbye concat :_(, I'll miss you.

src-browserify/core/log.js
xavijam: :+1:

xavijam:
@alonsogarciapablo, take a look too because it touches layer definition.

xavijam:
For me looks ok!

src-browserify/api/sql.js
```diff
@@ -1,686 +1,697 @@
...
-      var host = opts.host || 'cartodb.com';
-      var protocol = opts.protocol || 'https';
-      template = protocol + '://{user}.' + host;
+// NOTE this does not return a SQL directly, but a wrapper, to inject the dependencies
+// e.g. var SQL = require('./sql')({ jQuery: $ /* or reqwest: reqwest */ });
+// @param {Object} _Promise
+// @param {Object} opts hash with keys:
+//   jQuery: {Object} (Optional) has precedence over reqwest if both are present
+//   reqwest: {Object} (Optional)
+module.exports = function(_Promise, opts) {
```
alonsogarciapablo: I'm trying to understand the motivations behind this approach... Why do we want/need to inject `_Promise` for example?

src-browserify/api/core-lib/_promise.js
```diff
@@ -1,13 +1,16 @@
-var Backbone = window.Backbone || require('backbone-isch');
+// NOTE this does not return a _Promise directly, but a wrapper, to inject the dependencies
+// e.g. var _Promise = require('./_promise')(Backbone.Events)
+// @param {Object} BackboneEvents
```
viddo: I'll think I finally scrap these kind of comments, they're a bit tedious to write/maintain and most likely they're going to get out-of-date like usually happens when things start to be moved around. The dependencies names' are essentially self-explanatory anyway, I think.

src-browserify/api/sql.js
viddo: Read the PR description? I tried to explain the approach there. Anyway, I'll try to clarify further with this particular case since asked:
`_Promise` itself relies on `Backbone.Events` _or_ `BackboneIsch.Events` (the fake Backbone object used in core to not load the full Backbone lib, see `api/core-lib/backbone-isch.js`). Which one is used? Depends on the context/bundle (core or standard), so to avoid having to set this up in every place used and send around unnecessary dependencies (e.g. SQL don't use Backbone) this is wired up in the [core entry file](https://github.com/CartoDB/cartodb.js/pull/727/files#diff-36f4cca05a9602edf8347e27fedf45e5R33) (the standard bundle will be setup differently).

src-browserify/api/sql.js
```diff
@@ -1,686 +1,697 @@
...
-
-  if(options.cache !== undefined) {
-    params.cache = options.cache;
-  }
 
-  if(options.jsonp) {
-    delete params.crossDomain;
-    if (options.jsonpCallback) {
-      params.jsonpCallback = options.jsonpCallback;
+    this.ajax = options.ajax || (typeof(jQuery) !== 'undefined' ? jQuery.ajax: reqwest);
```
alonsogarciapablo: I know we just moved this, but it'd probably need to be `reqwest.compat`. (We're doing it [here](https://github.com/CartoDB/cartodb.js/blob/60285edb65f6575c8a34799fde55308751251c5f/src/geo/layer_definition.js#L68))

viddo: This comment is on the SQL object however, you're referencing is `MapBase` (which is matching that if you check the new file in this PR). However I do think you're right (looking at the usage).

src-browserify/core.js
```diff
@@ -6,46 +26,57 @@ cdb.CARTOCSS_VERSIONS = {
...
+var SubLayerBase = setupSubLayerBase(BackboneEvents);
+var HttpSubLayer = setupHttpSubLayer(SubLayerBase);
+var BackboneModel = window.Backbone ? window.Backbone.Model : null;
+var CartoDBSubLayer = setupCartoDBSubLayer(SubLayerBase, BackboneModel);
+var SubLayerFactory = setupSubLayerFactory(CartoDBSubLayer, HttpSubLayer);
+var MapBase = setupMapBase(SubLayerFactory, {
+  jQueryAjax: globalJQuery ? globalJQuery.ajax : undefined,
+  reqwestCompat: reqwest.compat
+});
+var LayerDefinition = setupLayerDefinition(MapBase, cdb.CARTOCSS_DEFAULT_VERSION);
```
alonsogarciapablo: I don't think `cdb.CARTOCSS_DEFAULT_VERSION` is something that will be different depending of what dist file we're using. It's more of a global config thing.

viddo: Yes, you're right on that, but to avoid having to _know_ that the version needs to be set in various places I think it's clearer to set the version in one place and pass it explicitly like this.

src-browserify/api/sql.js
viddo: Have had a offline conversation with @alonsogarciapablo to clarify the intentions and _why_ I chose to go down this path. While migrating code I have seen the benefits (untangling dependencies and getting a a clear overview) but with more verbose code. Neither of us are completely convinced though, the alternative would be to go back to using window.XXX all other the place. I guess it'll be a tradeoff. @rochoa since you did a similar migration and knows the ins of browserify too, could you have a look too and give an opinion?

alonsogarciapablo: Thanks for the time to clarify the motivations behind this approach @viddo!
@rochoa looking forward to hear your thoughts.

viddo: @alonsogarciapablo thx to @rochoa came up with a better solution :tada: the problem remains ofc, but with the alternative approach we avoid the annoying wrapper/setupXX I've used so far. 

I'll explain here with `src/core/model.js` as example, which is used in all 4 bundles (removed real code for brevity):

Instead of require dependencies right away, set through through a _proxy_ at entry point (i.e. the bundle):
```js
// standard.js
var $ = require('jquery');
require('../proxies/ajax').set($.ajax);
…
```
```js
// no-jquery.js
var reqwest = require('reqwest');
require('../proxies/ajax').set(reqwest.compat);
…
```
```js
// core.js
var reqwest = require('reqwest');
require('../proxies/ajax').set(reqwest.compat);

var cdb = {};
cdb.SQL = require('./api/sql');
…
```

Using code requires ajax thorugh the proxy object:
```js
// api/sql.js
var ajax = require('../proxies/ajax').get(); // get actual ajax function to use

function SQL() {
  this.ajax = options.ajax || ajax;
  // old code, for comparison:
  // this.ajax = options.ajax || (typeof(jQuery) !== 'undefined' ? jQuery.ajax: reqwest);
  ...

module.exports = SQL;
```

So the dependency is wired up at runtime, with the benefit of having the code declarative (through the requires). Sounds OK?

alonsogarciapablo: Nice! It looks much simpler! :clap: :clap: :clap:

PR #735, opened 2015-10-29
Widgets using BBOX filter

alonsogarciapablo:
It would be nice to be able to setup if a widget needs to be synced to the actual bounding box of the map. In this PR we're always sending the bbox filter when fetching the data for all widgets.

cc// @xavijam


alonsogarciapablo:
Closed in favor of #736.

PR #736, opened 2015-10-29
Widgets using BBOX filter

alonsogarciapablo:
It would be nice to be able to setup if a widget needs to be synced to the actual bounding box of the map. In this PR we're always sending the bbox filter when fetching the data for all widgets.

cc// @xavijam

### DIFF
--- a/src/windshaft/filters/bounding_box.js
+++ b/src/windshaft/filters/bounding_box.js
@@ -1,0 +1,22 @@
+cdb.windshaft.filters.BoundingBoxFilter = cdb.core.Model.extend({
+
+  initialize: function(bounds) {
+    this.setBounds(bounds);
+  },
+
+  setBounds: function(bounds) {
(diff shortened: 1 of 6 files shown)


src/windshaft/filters/bounding_box.js
```diff
@@ -0,0 +1,22 @@
+cdb.windshaft.filters.BoundingBoxFilter = cdb.core.Model.extend({
+
+  initialize: function(bounds) {
+    this.setBounds(bounds);
+  },
+
+  setBounds: function(bounds) {
+    this.southLatitude = bounds[0][1];
```
xavijam: why not using model methods instead of local vars?

xavijam:
For me :+1:

src/windshaft/filters/bounding_box.js
alonsogarciapablo: Fixed!

PR #738, opened 2015-10-29
Part 2 migration of standard bundle 

viddo:
Part 2 milestone for #699

- Migrates standard bundle to commonjs modules
- Removes core bundle (and will skip nojquery/noleaflet) based on discussion offline, since not clear if we'll support those going forward (at the moment), which simplifies the require calls significantly (no need to decide what lib to use at runtime etc.)


viddo:
@alonsogarciapablo @xavijam can you please review? changed scope i.e. remove core, skip nojquery/noleaflet bundles, as discussed offline end of last week so the code is finally much simpler

alonsogarciapablo:
:+1: Great job!

I mentioned some stuff to @viddo off-line and we agreed to not make any more changes. I added some tasks to #722.

Issue #739, opened 2015-10-29
InfoWindow `adjustPan` method never use the `callback` parameter.


acanimal:
Within the `infowindow.js` file the the `callback` parameter is never used (https://github.com/CartoDB/cartodb.js/blob/develop/src/geo/ui/infowindow.js#L826)

PR #741, opened 2015-10-29
Fixed empty popup

### LINKED ISSUE #737: Empty Information Popup box on published map
If we zoom in/out while a popup is open, there is a moment in which the height of popup changes and content is not shown.
Seems like a CSS problem.

acanimal:
Reviewer @xavijam.

Close #737 and #739.

Solves bug that happens when zoom changes, showing popup really small making content never displayed.


xavijam:
Looks good man! :+1:

PR #742, opened 2015-10-30
Fix bounding box naming to match west,south,east,north

rochoa:


rochoa:

alonsogarciapablo:
:+1:

PR #754, opened 2015-11-03
Migrate cartodb.mod.torque to CommonJS/browserify modules

viddo:
Part milestone for #699 
Migrated the torque bundle to CommonJS modules/browserify

Had trouble using the standalone build file of vendor/mod/torque.uncompressed.js, due to external dependencies not being resolved properly. Instead I've used the torque.js npm module which seem to work fine (was already migrated to CommonJS modules/browserify at some earlier point.

@alonsogarciapablo can you review this one too? is way smaller that prev ones
@rochoa can you also have a look, since you did the migration of torque?


src-browserify/torque.js
```diff
@@ -0,0 +1,23 @@
...
+var torque = window.torque = require('torque.js'); // standalone torque lib, required for gmaps/leaflet layer view
+
+// Get and mutate the global cdb object, add expected objects for gmaps/leaflet if present:
+var cdb = window.cdb;
+if (!cdb) {
+  throw new Error('cartodb.js is required for this lib to work, load it before this one');
+}
+
+if (typeof google !== 'undefined' && typeof google.maps !== 'undefined') {
+  cdb.geo.GMapsTorqueLayerView = require('./geo/gmaps/gmaps-torque-layer-view');
```
alonsogarciapablo: Why we do need to set `GMapsTorqueLayerView` and `LeafletTorqueLayer` in `cdb.geo`?

alonsogarciapablo: I'm guessing it has something to do with how we load torque in cartodb.js?

viddo: Yes, just maintaining the current mechanisms, e.g. https://github.com/CartoDB/cartodb.js/blob/v4-browserify-torque/src-browserify/geo/gmaps/gmaps-map-view.js#L26-L28

alonsogarciapablo:
:+1:

PR #755, opened 2015-11-03
Migrate cartodb.mod.odyssey to browserify

viddo:
Part milestone for #699 

Just bundle since there were no specific tests for this one, added an integration test case to verify the load working as expected though.

@javierarce can your review? :pray: 
cc @matallo @alonsogarciapablo


javierarce:
Nice nice :+1:

PR #757, opened 2015-11-03
Extract source-map files from inlined source to separate files

viddo:
Part milestone for #699 

- Uglify (minimize) browserify bundles, maintain source-maps
- Remove nojquery/noleaflet since they won't be supported

Note that this was branched off from torque PR to not be blocked

@alonsogarciapablo can your review?


viddo:
or @xavijam since you probably recognize this from cartodb, since it's the same approach used there

xavijam:
\o/ :clap:

PR #732, opened 2015-10-27
new docs formatting

matallo:
formatting changes needed for the new docs

CR @csobier @rochoa

PR #760, opened 2015-11-04
Added documentation about the no_cdn option used in the createLayer method

### LINKED ISSUE #759: Document the `no_cdn` option within the `createLayer` API method
See #531.

Although not documented it is also possible to specify the `no_cdn` options within the `createLayer` method. 
Specify the `no_cdn` option in `createVis` makes all layers works in the same way. Using it in the `createLayer` offers more flexibility because we can chose which layers must load from CDN.

acanimal:
Reviewer @xavijam 

Although possible there was no documentation about the use of `no_cdn` in the `createLayer` method.

Close #759

### DIFF
--- a/doc/api_methods.md
+++ b/doc/api_methods.md
@@ -9,6 +9,7 @@ The documentation below refers to CartoDB.js v3. For major changes in the librar
 Creates a visualization inside the map_id DOM object.
 
 <div class="code-title">cartodb.createVis</div>
+
 ```javascript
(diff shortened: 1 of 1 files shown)


xavijam:
:+1:

alonsogarciapablo:
We should probably make this change in the new docs. There's [a PR out there](https://github.com/CartoDB/cartodb.js/pull/732) that will override the existing docs (and this change).

PR #771, opened 2015-11-05
Small refactoring 

alonsogarciapablo:
Changes:

- Added a "named map" to the example.
- Added a couple of new views for Widgets.
- Changed Dashboard a bit so that it only listens to events on layers instead of listening to events on widgets, filters, etc.

@viddo I'd love to hear your feedback!


src/geo/map/cartodb-layer.js
```diff
@@ -5,7 +6,7 @@ var MapLayer = require('./map-layer');
 var CartoDBLayer = MapLayer.extend({
 
   defaults: {
-    attribution: config.get('cartodb_attributions'),
+    attribution: cdb.config.get('cartodb_attributions'),
```
viddo: no need to add `cdb.`, see the require above (that returns the global instance), I think it's better to use the require approach than reaching for objects in the global namespace (which is more intended to give access to object from outside the lib).

src/geo/map/cartodb-layer.js
```diff
@@ -25,15 +26,28 @@ var CartoDBLayer = MapLayer.extend({
...
 
+  initialize: function() {
+    this.widgets = new Backbone.Collection([]);
+
+    // Re-trigger the change:filter event
+    this.widgets.bind('change:filter', function(widget, filter) {
+      this.trigger('change:filter', this, widget, filter);
+    }, this);
+
+    cdb.geo.MapLayer.prototype.initialize.apply(this, arguments);
```
viddo: Same thing as for config, already required above:
```js
MapLayer.prototype.…
```

viddo:
Apart from the minor comments looks legit :+1: 
But I'm afraid I don't have a enough insights of how the widgets are implemented just yet to really give a good opinion on the changes, so you might want a 2nd opinion here

PR #773, opened 2015-11-06
Widgets torque take 2

javisantana:


alonsogarciapablo:
:es:

src/geo/map/torque-layer.js
```diff
@@ -1,5 +1,7 @@
 var _ = require('underscore');
 var MapLayer = require('./map-layer');
+var Backbone = require('backbone');
+var _ = require('underscore');
```
viddo: why require underscore twice?

javisantana: I do it because imagine with I read the first underscore the memory gets corrupted by a electric impulse (from a shitty electric car, for example).

Today is unlike that happens, but imagine in some years when everybody will being driven by electric cars all over the place.

This enables cartodb.js to work in electric cars.

That or I fucked up with a copy & paste :(

src/windshaft/dashboard_instance.js
```diff
@@ -190,17 +201,19 @@ module.exports = Model.extend({
...
 
-    var layers = this.get('metadata').layers;
-    var url;
-    _.each(layers, function(layer) {
-      var widgets = layer.widgets;
-      for (var id in widgets) {
-        if (widgetId === id) {
-          url = widgets[id].url[protocol];
-          return;
+    if(this.get('metadata')) {
```
javisantana: @alonsogarciapablo I noticed when the bbox changes and the map is not instanciated yet this raises and error. This check fixes the error but not the source problem

PR #776, opened 2015-11-06
Adding Dashboard-canvas to the stack

xavijam:
- Adding new element (only css) to the map stack, dashboard map, it will have the map (always) and widgets (when they are any). In the future we will add to this content the torque time series.
- Fixed responsive version, it was broken.

@alonsogarciapablo.


alonsogarciapablo:
:+1:

PR #783, opened 2015-11-07
CSS changes for new time-series widget

viddo:
Adapted the markup and styles to take into account the time-series widget that is located below the map canvas instead of the widgets list.

For now it's just hardcoded markup, it will be replaced with the actual widget (not included in these changes, WIP).

@xavijam please review


xavijam:
For me it is a great solution, we should check how to proceed with widgets management but this is for other discussion.

Issue #786, opened 2015-11-10
Add time-series widget (first iteration)


viddo:
Main issue to keep track of WIP

- [x] Required layout changes for new widget #783
- [x] Placeholder views for the various concepts (using fake model/data); controls (play/pause), step info, histogram chart #787
- [x] Implement default date/time histogram #811
- [x] Histogram: format X-axis labels with time and/or date (needs to adapt to range span+mockups) #843
- [x] Torque: implement animation controls (play/pause, time-marker) #844
- [x] Torque: animation for a selected time-series range #855 + https://github.com/CartoDB/torque/pull/243
- [x] Torque: format time-marker step info (time+date) #855 
- [x] CSS for the various layouts/responsive sizes #879

Windshaft needs to be updated to support date column for histogram data :point_right: https://github.com/CartoDB/Windshaft/issues/405

PR #787, opened 2015-11-10
Add shell views for time-series widget

viddo:
Waypoint for #786

Adds some trivial shell views for the various concepts that are included in the time widget. I stopped here because the histogram chart view required some massage to work in this context. I left some TODOs for some details that will need some further work for the histogram widget, ok @javierarce?


javierarce:
:+1: 

And thanks for the histogram refactor :)

PR #788, opened 2015-11-11
Fixes infowindows for named and anonymous maps

alonsogarciapablo:
Infowindows were not working properly both for named and anonymous "maps". The behaviour is a bit different in those two cases, specially when some layers are hidden.


src/geo/map/cartodb-layer-group-base.js
```diff
@@ -0,0 +1,77 @@
...
+      featureID
+    ].join('/');
+
+    $.ajax({
+      dataType: 'jsonp',
+      url: url,
+      jsonpCallback: '_cdbi_layer_attributes_' + util.uniqueCallbackName(this.toJSON()),
+      cache: true,
+      success: function(data) {
+        // loadingTime.end();
```
xavijam: Why is it commented?

src/geo/map/cartodb-layer-group-base.js
```diff
@@ -0,0 +1,77 @@
...
+      tilejson: '2.0.0',
+      scheme: 'xyz',
+      grids: urls.grids[this._getIndexOfVisibleLayer(layerIndex)],
+      tiles: urls.tiles,
+      formatter: function(options, data) { return data; }
+    };
+  },
+
+  _getIndexOfVisibleLayer: function(layerIndex) {
+    throw "_getIndexOfVisibleLayer must be implemented";
```
javisantana: I guess in the classes that inherit this, rigth?

src/geo/map/cartodb-layer-group-base.js
alonsogarciapablo: Great question. This method "comes" from [here](https://github.com/CartoDB/cartodb.js/blob/develop/src/geo/layer_definition.js#L327-L346) and I haven't paid much attention to profiling so far. I can fix it as part of this PR or we can come back to it later.

src/geo/map/cartodb-layer-group-base.js
alonsogarciapablo: Yes.

javisantana:
:es:

I guess the main problem of hiding layers is still there but let's see

PR #791, opened 2015-11-11
Encodes filters param when instantiating a dashboard

### LINKED ISSUE #778: Filters JSON must be URI encoded
For instance a filter containing a category with `&` char will split the JSON as two query string params.

alonsogarciapablo:
Fixes #778.

@rochoa will this work? Thanks!

### DIFF
--- a/src/windshaft/client.js
+++ b/src/windshaft/client.js
@@ -59,7 +59,7 @@ WindshaftClient.prototype.instantiateMap = function(options) {
 
   var filters = filters || {};
   if (Object.keys(filters).length) {
-    params.push(["filters", JSON.stringify(filters)].join('='));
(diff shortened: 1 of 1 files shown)


alonsogarciapablo:
@xavijam can you take a look at this please?

rochoa:
We are already doing `encodeURIComponent` in some filters: https://github.com/CartoDB/cartodb.js/blob/86d02695231f4b2cb0a6d1d8754914bff688e195/src/windshaft/filters/category.js#L108-L109

I aim for a general solution but we have to check we're not double encoding it.

xavijam:
Agree, we should create a general rule for this. Maybe included in the filter toJSON thing.

viddo:
Why not leave the encoding to when the query string is actually needed instead of having to do this manually? e.g. using https://github.com/sindresorhus/query-string or similar

PR #792, opened 2015-11-11
Feature 789 disable dragging

### LINKED ISSUE #789: Disable map panning when scrollwheel and zoom "controls" are not enabled
Possibility to avoid map panning when zoom overlay is not present and scrollwheel is disabled.

See: https://github.com/CartoDB/cartodb/issues/5995

acanimal:
Reviewer @alonsogarciapablo 

Close #789, close https://github.com/CartoDB/cartodb/issues/5995 

This PR allows to disable the map dragging when no zoomControl (overlay) is specified and the `scrollwheel` option is disabled.

**Note with this implementation you can only set dragging at construction time, for the moment is not possible to change the property later.**

Modifications:

- new `drag` property on `cdb.geo.Map` model class.
- Both `cdb.geo.GoogleMapsMapView` and `cdb.geo.LeafletMapView` uses the model `drag` attribute to initialize the map.

> Notes:
> - Can’t test `cdb.geo.LeafletMapView`. Although you disable dragging map continues moving using `panTo` or `setView`. The alternative is emulate a mouse dragging effect.

### DIFF
--- a/examples/map_dragging_disabled.html
+++ b/examples/map_dragging_disabled.html
@@ -1,0 +1,41 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>Map with dragging disabled | CartoDB.js</title>
+    <meta name="viewport" content="initial-scale=1.0, user-scalable=no" />
+    <meta http-equiv="content-type" content="text/html; charset=UTF-8"/>
+    <link rel="shortcut icon" href="http://cartodb.com/assets/favicon.ico" />
+    <style>
+      html, body, #map {
+        height: 100%;
+        padding: 0;
+        margin: 0;
+      }
+    </style>
+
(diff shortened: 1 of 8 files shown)


alonsogarciapablo:
Taking a look!

acanimal:
- `setTimeout` completely unnecessary. Changed for a `createVis().done()`.
- `_.isObject`: I want a boolean flag so, first I use `_.find` to check if there is an object with `type=="zoom"`. This returns an object or undefined. I get a boolean value checking the result with  `_.isObject`. I think it is more clear than comparing to "undefined" and there is not much overhead.
- Added your test to the spec.
- Added similar tests for gmaps too.

alonsogarciapablo:
Just left a couple of comments. After those are fixed: :+1: Good job!

PR #799, opened 2015-11-13
Fixes test suite

alonsogarciapablo:
- Removes `src/geo/leaflet-extensions.js`. This file defined `L.CartoDBGroupLayerBase`,  `L.CartoDBGroupLayer`, `L.NamedMap`, and `L.CartoDBLayer`, which were probably not being used and don't make much sense at this point.
- Removes views for layers of type "CartoDB" (`leaflet-layer-cartodb-view` and `gmaps--layer-cartodb-view`). These layers are always rendered as part of a "LayerGroup", so it doesn't make sense to have views for them. 
- Simplifies the hierarchy of classes for Leaflet and GMaps layer group and named map views. Most of the logic has been extracted from the views to models, so we can know live with one type of view for groups of CartoDB layers (one for Leaflet and one for GMaps).
- Removed some classes and tests that are broken and might not me used anymore (or will have to be rewritten):
  - src/api/tiles
  - src/vis/image
  - src/image/static-image
  - src/api/create-layer

Apart from all the code/tests that was :fire:, there are 4 test files which are commented out. We will probably delete them in the near future.

The basic example (`examples/index.html`) is working fine both for Leaflet and Google maps, with public and private datasets.

@viddo please take a look! Thanks!

cc: @javisantana


viddo:
:tada:

viddo:
> Removes src/geo/leaflet-extensions.js. This file defined L.CartoDBGroupLayerBase, L.CartoDBGroupLayer, L.NamedMap, and L.CartoDBLayer, which were probably not being used and don't make much sense at this point.

That file and definitions was created match the [code prior to the migration](723). If they're not used/required to be exposed then I agree :+1: :fire: 
In fact, that most likely applies to more `cdb.foo` objects too, but I guess we can leave that for a later discussion.

grunt/tasks/jasmine.js
```diff
@@ -21,7 +21,9 @@ module.exports = {
     return {
 
       'src': {
-        src: [], // actual src files are require'd in the *.spec.js files
+        src: [
+          'dist/cartodb.uncompressed.js',
+        ],
```
viddo: Why was this added? The source files already require what they need so this should not be necessary. The "integration" tests that use the bundle was intended to be located in the "cartodb" bundle defined below.

src/geo/gmaps/cartodb-layer-group-base.js
```diff
@@ -90,20 +94,21 @@ CartoDBLayerGroupBase.prototype.clear = function () {
...
+    self.tilejson = tilejson;
+    self.options.tiles = tilejson.tiles;
+    self.tiles = 0;
+    self.cache = {};
+    self._reloadInteraction();
+    self.refreshView();
+    self.ok && self.ok();
+    done && done();
+  } else {
+    self.error && self.error('URLs have not been fetched yet');
```
viddo: the self is not really necessary anymore, since the code is not located in a closure, perhaps we can tidy up this?

src/geo/gmaps/gmaps-torque-layer-view.js
```diff
@@ -1,9 +1,10 @@
 // Depends on cartodb standard bundle being loaded and present in global namespace
 var torque = require('torque.js');
-var cdb = window.cdb;
-var _ = cdb._;
+var cdb = require('cdb');
+var _ = require('underscore');
```
viddo: Same case as what we discussed offline re: the leaflet-layer-view; the cdb object and libs used are "inherited" from the cartodb.js lib and should not be required here.

viddo: Created a separate issue to deal with this later :point_right: #814

src/geo/gmaps/cartodb-layer-group-base.js
alonsogarciapablo: Done! Great catch!

src/geo/gmaps/gmaps-torque-layer-view.js
alonsogarciapablo: Done too! :+1:

viddo:
Apart from the comments on the minor things, overall looks :+1: code-wise. I lack knowledge about a large portion of the code to know whether the changes are valid though, so it's probably best if @javisantana have a look too.

Also, what about the TODOs that was added, create separate issues to tackle them?

grunt/tasks/jasmine.js
alonsogarciapablo: If I remove it now, then `cdb._` doesn't get defined and there's a dependency on that [here](https://github.com/CartoDB/cartodb.js/blob/fix-test-suite/src/geo/gmaps/gmaps-torque-layer-view.js#L4).

alonsogarciapablo: Basically, some tests under `src` use `require('torque')` which then requires some files (`cdb./geo/gmaps/gmaps-torque-layer-view` and ``cdb/geo/leaflet/leaflet-torque-layer`) that rely on `cdb._`, which is defined by `cartodb.js`... Thoughts?

viddo: Liked I mentioned earlier offline I'd suggest to extract those particular test-cases, torque-specs bundle loads both cartodb.js and torque bundles for example and would probably be a good place. 
That that the change I proposed in #814 will be covered by these tests too.

alonsogarciapablo:
Thanks for your feedback @viddo. I'm going to cleanup TODOs (and open issues if necessary).

I'd like your input on [this comment](https://github.com/CartoDB/cartodb.js/pull/799#discussion_r45091378). Thanks!

PR #801, opened 2015-11-14
this will allow non layer group layers to have a infowindow.

javisantana:
cc @alonsogarciapablo

### DIFF
--- a/src/geo/map/torque-layer.js
+++ b/src/geo/map/torque-layer.js
@@ -31,6 +31,7 @@ var TorqueLayer = MapLayer.extend({
   },
 
   getInfowindowData: function() {
+    // TODO: use infowindow model in the future
     var infowindow = this.get('infowindow');
     if (infowindow && infowindow.fields && infowindow.fields.length) {
       return infowindow;
--- a/src/geo/map/torque-layer.js
+++ b/src/geo/map/torque-layer.js
@@ -31,6 +31,10 @@ var TorqueLayer = MapLayer.extend({
   },
 
   getInfowindowData: function() {
+    var infowindow = this.get('infowindow');
+    if (infowindow && infowindow.fields && infowindow.fields.length) {
+      return infowindow;
+    }
     return null;
   },
 
@@ -54,6 +58,9 @@ var TorqueLayer = MapLayer.extend({
     return this.widgets.map(function(widget) {
       return widget.getFilter();
     });
+  },
+
+  fetchAttributes: function(layer, featureID, callback) {
   }
 
 });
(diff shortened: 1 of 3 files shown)


src/geo/map/torque-layer.js
```diff
@@ -31,6 +31,10 @@ var TorqueLayer = MapLayer.extend({
   },
 
   getInfowindowData: function() {
+    var infowindow = this.get('infowindow');
```
javisantana: there is a lot of repeated code between the cartodb layers and torque, we may want to have a "interactivelayer" base class to have this kind of things

javisantana: cc @alonsogarciapablo  ^^

src/vis/vis.js
```diff
@@ -1208,7 +1208,16 @@ var Vis = View.extend({
...
     var mapView = this.mapView;
     var infowindow = null;
-    var layers = layerView.model && layerView.model.layers || [];
+    var layers = [];
+    if (layerView.model) {
+      if (layerView.model.layers) {
+        layers = layerView.model.layers;
+      } else {
+        if (layerView.model.getInfowindowData) {
+          layers = new Backbone.Collection([layerView.model]);
```
javisantana: not sure if this is the best way, should we do add the interactive layers have "sublayers" ? cc @alonsogarciapablo

src/geo/map/torque-layer.js
alonsogarciapablo: Something I've realised is that the layer is not using the InfowindowModel that is created [here](https://github.com/CartoDB/cartodb.js/blob/widgets/src/vis/vis.js#L1464-L1467). It probably makes more sense to move that logic to the infowindow model and use it here (instead of working with the data that comes in the viz.json). So basically we'd do something like:

```
var TorqueLayer = MapLayer.extend({

  // ...

  getInfowindowData: function() {
    this.infowindow.getFields();
  },

  // ...
})

```

javisantana: I will change to use the model

src/vis/vis.js
alonsogarciapablo: This is probably something that each "layer view" could handle differently. For instance, if a new layer view is created and it's model has an infowindow model, it can create the corresponding InfowindowView and bind to the corresponding events (featureClick, etc). differently (also fetch attributes differently, etc.).

alonsogarciapablo:
I left a couple of comments. I think it makes sense to address my concerns in a separate/future PR so I'll open an issue.

:es:

PR #806, opened 2015-11-16
Merge fixes from develop into v4 branch

### LINKED ISSUE #803: Port fixes from `develop` to `v4` and `widgets` branches
Port fixes made on develop branch to the v4 which uses browserify.

acanimal:
This PR contains the fixes on `develop` branch (until today nov-16) into the `v4` branch which is the base for `widgets` branch.

reviewer @alonsogarciapablo cc @xavijam @viddo 

close #803

### DIFF
--- a/src/geo/leaflet/leaflet-map-view.js
+++ b/src/geo/leaflet/leaflet-map-view.js
@@ -45,8 +45,15 @@ var LeafletMapView = MapView.extend({
       // remove the "powered by leaflet"
       this.map_leaflet.attributionControl.setPrefix('');
 
+      // Disable scrollwheel
       if (this.map.get("scrollwheel") == false) this.map_leaflet.scrollWheelZoom.disable();
+      // Disable keyboard
       if (this.map.get("keyboard") == false) this.map_leaflet.keyboard.disable();
+      // Disable dragging (also doubleClickZoom)
+      if (this.map.get("drag") == false) {
+        this.map_leaflet.dragging.disable();
+        this.map_leaflet.doubleClickZoom.disable();
+      }
 
     } else {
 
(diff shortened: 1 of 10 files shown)


alonsogarciapablo:
Looks good :+1:

src/geo/leaflet/leaflet-map-view.js
```diff
@@ -45,8 +45,15 @@ var LeafletMapView = MapView.extend({
       // remove the "powered by leaflet"
       this.map_leaflet.attributionControl.setPrefix('');
 
+      // Disable scrollwheel
       if (this.map.get("scrollwheel") == false) this.map_leaflet.scrollWheelZoom.disable();
+      // Disable keyboard
       if (this.map.get("keyboard") == false) this.map_leaflet.keyboard.disable();
+      // Disable dragging (also doubleClickZoom)
```
viddo: Do these really add any value? I find the code to be simple enough to be self-explanatory

acanimal: I tend to comment all the things a made, probably too much:)
Comment a line is probably not a good idea but I think it is a good practice place one line comment each few blocks of code, not the code itself but the logic behind them.

viddo: IMHO comments only makes sense if they explain the intention of a particular snippet that can't be captured in code itself (e.g answers _why does the code do this or that?_)

acanimal: I'll change it before merge `v4` with `widgets` branch. Probably not well expressed the comment. The idea is to indicate the dragging and double click options are disabled because the `drag` options is disabled.

Another option, probably better, is not comment here and comment appropriately the `drag` option in   the map model class.

Issue #804, opened 2015-11-16
Extract viz.json parsing and separate vis model from vis view


alonsogarciapablo:
Here're the tasks for this refactor:

- [x] Extract a VizJSON class that contains vizjson internally and provides methods to modify it / extract info from it.
- [x] Extract loading of vizjson file to createVis.
- [x] Create visModel class and use it internally in vis/vis.js.
- [x] Extract construction of visModel object to `createVis` and inject it to vis/vis.js.
- [x] Rename `vis.load` to `vis.render`.

acanimal:
+1 Complex structures of object should be created through factories.

acanimal:
Add some comments:

-  Private `vis._addLegends` and `vis._createLayerLegendView` method are called externally from `createLayer.js`.
- Vis has listeners on `this.map.layers` (where `map` is a MapModel and `layers` a backbone collection) to update legends (`Vis.addLegend()`. We need to apply SOLID principles. I think Vis.load must be only responsible to create all the objects required to have a visualization (maps, views, models, etc). It must be like a factory for creation and also have method to interact with the visualization (adding more legends given a data config, adding new layers, etc).

alonsogarciapablo:
This was completed in https://github.com/CartoDB/cartodb.js/pull/1313 💪

PR #811, opened 2015-11-17
Implement date/time histogram widget

viddo:
Waypoint for #786, implementing the time-series widget for a histogram with a date column.

screen shot 2015-11-17 at 14 07 07

- An example is created at examples/time-series-widget.html, which requires the vis to be setup with a layer that has a date column for the widget to load as expected.
- As agreed after talking with @urbanphes, I've removed the controls for now, will be added/implemented later for a torque-layer.
- Refactored the widget-views creation, to a factory that allow defining arbitrary rules of what view to create for a certain context, necessary to differentiate widget of type histogram here. cc @alonsogarciapablo since we discussed it offline

@javierarce can you review?


examples/time-widget.html
```diff
@@ -0,0 +1,402 @@
...
+                          "layer_name": "mwc13-static",
+                          "cartocss": "/** simple visualization */\n\n#mwc13{\n  marker-fill-opacity: 0.9;\n  marker-line-color: #FFF;\n  marker-line-width: 1;\n  marker-line-opacity: 1;\n  marker-placement: point;\n  marker-type: ellipse;\n  marker-width: 10;\n  marker-fill: #229A00;\n  marker-allow-overlap: true;\n  marker-comp-op: multiply;\n}",
+                          "cartocss_version": "2.1.1",
+                          "interactivity": "cartodb_id",
+                          "table_name": "mwc13"
+                        },
+                        "widgets": {
+                          "time_static_widget_uuid": {
+                            "type": "histogram",
+                            column_type: 'date', // to differentiate from a quantities histogram
```
viddo: @rochoa I've added this prop temporarily, to be able to differentiate a "normal" histogram from a time one (search for `column_type ` in this PR to see where/how it's used), do you have any other suggestion or preferred way to deal with this related to the tiler?

src/vis/vis.js
```diff
@@ -23,17 +23,74 @@ var INFOWINDOW_TEMPLATE = require('./vis/infowindow-template');
...
 /**
- * visulization creation
+ * Visualization creation
  */
 var Vis = View.extend({
 
   initialize: function() {
     _.bindAll(this, 'loadingTiles', 'loadTiles', '_onResize');
 
+    // TODO this should probably be extracted, together with the .load method
```
viddo: Kind of related to #804, I configure the widget-view-factory here since this `vis.load` is where all things are wired up. cc @alonsogarciapablo @acanimal

examples/time-widget.html
rochoa: I guess that's because you need that in the frontend, right? If API were returning that as part of the widget metadata you could live without that prop, couldn't you?

javierarce:
:ok_hand: for me

examples/time-widget.html
viddo: Yes, if the metadata is provided in the vizjson(?) that would work.

src/geo/ui/widgets/widget_view_factory.js
```diff
@@ -1,24 +1,27 @@
...
+};
 
-    // TODO: widgetViewClass might be null
-    var widgetView = new widgetViewClass({
-      model: widget,
-      filter: widget.filter
-    });
+WidgetViewFactory.prototype.addType = function (def) {
+  if (!_.isFunction(def.match)) new Error('def.match must be a function');
+  if (!_.isFunction(def.create)) new Error('def.match must be a function');
```
alonsogarciapablo: Error should say `def.create must be a function`

examples/time-widget.html
rochoa: Not sure about vizjson being the responsible for that, I was thinking about *maps api* response returning that information.

viddo: The thing is that the models and views are created at the time of parsing the vizjson. I'm not very keen to complicate the view creation right now if we can avoid it.

E.g. for a torque layer I can just check the layer type to take the decision. The problem right now is that the layer type is the same as the existing cartodb layer type so can't distinguish them. Other ideas?

rochoa: The problem will come when you don't have a vizjson and you want to create a timeseries histogram with a date column programmatically, then you must rely on the API returning the data to know the column type, don't you?

PR #813, opened 2015-11-17
Improves xAxis calculation

javierarce:
Now the elements of the axis are linked to the vertical lines: 

Example with 4 divisions:
screen shot 2015-11-17 at 17 43 39

Example with 6 divisions:
screen shot 2015-11-17 at 17 44 02

### DIFF
--- a/src/geo/ui/widgets/histogram/chart.js
+++ b/src/geo/ui/widgets/histogram/chart.js
@@ -11,6 +11,7 @@ module.exports = View.extend({
     handleWidth: 6,
     handleHeight: 23,
     handleRadius: 3,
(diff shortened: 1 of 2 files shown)


javierarce:
@viddo, @saleiva: can you have a look, please?

javierarce:
BTW: I just realised that right now we are using 4 fixed divisions but the number should be dynamic depending on the width of the histogram… I'm adding a fix to calculate the # of divisions dynamically.

saleiva:
Ossom,

LMK when done so we takr a look :)

viddo:
Looks legit :+1:

PR #815, opened 2015-11-17
Workaround to fix issue on initial map and infowindow displacement

### LINKED ISSUE #793: Position of infowindows is not correct when the time slider widget is present
Infowindows are not showing up in the right position when the time slider "widget" is present. For some reason, there's some vertical offset (approximately the height of the times slider widget).

screen shot 2015-11-11 at 16 21 43

Resizing the window seems to fix the issue.

acanimal:
reviewer @alonsogarciapablo cc @xavijam 

close #793 

fixes issue about infowindow (really map) displacement until window is resized.

### DIFF
--- a/src/vis/vis.js
+++ b/src/vis/vis.js
@@ -553,6 +553,11 @@ var Vis = View.extend({
       self.trigger('done', self, map.layers);
     })
 
+    // FIXME - See: https://github.com/CartoDB/cartodb.js/issues/793#issuecomment-157442059
+    if(this.mapView.map_leaflet) {
+      this.mapView.map_leaflet.invalidateSize();
+    }
+
     return this;
 
   },


alonsogarciapablo:
Would it also work if we add the widgets before rendering the map view then?

xavijam:
Oh, wait, we have solved by accident here -> https://github.com/CartoDB/cartodb.js/commit/b2fef44f804731c0cb38f8ea3a687d6aac2fd40e

We found the same problem for bounding box, so seems like we have solved this issue too.
:S sorry.

acanimal:
It continue being a workaround :) I think it would be better to rethink the load method and create things in the right order.

xavijam:
Yeah, the problem is when we add widgets to the view, map dom element is resized, and leaflet map is not notified, so boundings keep incorrec.

acanimal:
Close this PR, due it is solved in another branch.

PR #822, opened 2015-11-18
Extracts specs that depend on torque

### LINKED ISSUE #817: Clearly separate tests that depend on torque
Right now, some tests require `torque` to have been previously loaded. As suggested [here](https://github.com/CartoDB/cartodb.js/pull/799/files#r45091920), we can create two bundles of tests (one using just cartodb.js and another using cartodb.js with torque).

cc: @viddo

alonsogarciapablo:
Fixes #817.

@viddo what do you think? thanks!


viddo:
:ok_hand:

PR #824, opened 2015-11-18
Adapts the histogram model to generate a full bucket list from the server data

javierarce:
We've changed the backend to return only the bins that contains actual data. This PR updates the histogram model to generate the missing bins in the client.

/cc @viddo @rochoa

### DIFF
--- a/src/geo/ui/widgets/histogram/model.js
+++ b/src/geo/ui/widgets/histogram/model.js
@@ -58,12 +58,26 @@ module.exports = WidgetModel.extend({
   },
 
   parse: function(data) {
+    var numberOfBins = data.bins_count;
+    var width = data.bin_width;
+    var nulls = data.nulls_count;
+    var start = data.bins_start;
 
-    this._data.reset(data.bins);
+    var buckets = new Array(numberOfBins);
+    _.each(data.bins, function(b) {
+      buckets[b.bin] = _.omit(b, 'bins_number', 'bin_width', 'nulls_count');
+    });
+
+    for (var i = 0; i < numberOfBins; i++) {
+      var binStart = start + (i * width);
+      var binEnd = start + ((i + 1) * width);
+      buckets[i] = _.extend({ bin: i, start: binStart, end: binEnd, freq: 0 }, buckets[i]);
+    }
+
+    this._data.reset(buckets);
 
     return {
-      data: data.bins,
-      width: data.width
+      data: buckets
     };
   },
 
--- a/src/geo/ui/widgets/histogram/model.js
+++ b/src/geo/ui/widgets/histogram/model.js
@@ -64,8 +64,9 @@ module.exports = WidgetModel.extend({
     var start = data.bins_start;
 
     var buckets = new Array(numberOfBins);
+
     _.each(data.bins, function(b) {
-      buckets[b.bin] = _.omit(b, 'bins_number', 'bin_width', 'nulls_count');
+      buckets[b.bin] = b;
     });
 
     for (var i = 0; i < numberOfBins; i++) {
--- a/src/geo/ui/widgets/histogram/model.js
+++ b/src/geo/ui/widgets/histogram/model.js
@@ -70,9 +70,12 @@ module.exports = WidgetModel.extend({
     });
 
     for (var i = 0; i < numberOfBins; i++) {
-      var binStart = start + (i * width);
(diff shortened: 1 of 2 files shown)


src/geo/ui/widgets/histogram/model.js
```diff
@@ -58,12 +58,26 @@ module.exports = WidgetModel.extend({
...
+      var binStart = start + (i * width);
+      var binEnd = start + ((i + 1) * width);
+      buckets[i] = _.extend({ bin: i, start: binStart, end: binEnd, freq: 0 }, buckets[i]);
+    }
+
+    this._data.reset(buckets);
 
     return {
-      data: data.bins,
-      width: data.width
```
javierarce: We don't really need to know the `width` on the front-end side.

src/geo/ui/widgets/histogram/model.js
```diff
@@ -58,12 +58,26 @@ module.exports = WidgetModel.extend({
...
   parse: function(data) {
+    var numberOfBins = data.bins_count;
+    var width = data.bin_width;
+    var nulls = data.nulls_count;
+    var start = data.bins_start;
 
-    this._data.reset(data.bins);
+    var buckets = new Array(numberOfBins);
+    _.each(data.bins, function(b) {
+      buckets[b.bin] = _.omit(b, 'bins_number', 'bin_width', 'nulls_count');
```
rochoa: You don't need to do the omit here. It was like that server side due data being retrieved from database, however server side is already *cleaning* that for you :smile:.

src/geo/ui/widgets/histogram/model.js
```diff
@@ -58,12 +58,26 @@ module.exports = WidgetModel.extend({
...
+    for (var i = 0; i < numberOfBins; i++) {
+      var binStart = start + (i * width);
+      var binEnd = start + ((i + 1) * width);
+      buckets[i] = _.extend({ bin: i, start: binStart, end: binEnd, freq: 0 }, buckets[i]);
+    }
+
+    this._data.reset(buckets);
 
     return {
-      data: data.bins,
```
rochoa: Off topic: should we return the number of nulls?

src/geo/ui/widgets/histogram/model.js
javierarce: Oh, that's true! I'll omit the _.omit then :)

src/geo/ui/widgets/histogram/model.js
javierarce: I think you are right. We need that for the UI. I'll add it.

rochoa:
:ship: :it: !!

src/geo/ui/widgets/histogram/model.js
```diff
@@ -58,12 +58,26 @@ module.exports = WidgetModel.extend({
...
-    this._data.reset(data.bins);
+    var buckets = new Array(numberOfBins);
+    _.each(data.bins, function(b) {
+      buckets[b.bin] = _.omit(b, 'bins_number', 'bin_width', 'nulls_count');
+    });
+
+    for (var i = 0; i < numberOfBins; i++) {
+      var binStart = start + (i * width);
+      var binEnd = start + ((i + 1) * width);
+      buckets[i] = _.extend({ bin: i, start: binStart, end: binEnd, freq: 0 }, buckets[i]);
```
viddo: I've seen similar patterns here and there, I'd propose to change this code block to:
```js
_. defaults(buckets[i], {
  bin: i,
  start: start + (i * width),
  end: start + ((i + 1) * width),
  freq: 0
}
```

…due to:
1. Although a one-liner can be more concise it's actually harder to read and follow what's going on. IMHO one-liner objects should only be used when there's for 1-2 props max, e.g. `{simple:'enough'}`
2. binStart/binEnd variables aren't really necessary, especially if having the bucket definition on multiple lines it's clear what the calculations represent (compare above)
3. Use `_.defaults` to modify the `buckets[i]` directly instead of having to create a copy object and do an assignment. I assume start/end props not to exist on the object though so disregard this if that's isn't the case

viddo:
One detail, otherwise all :+1:

src/geo/ui/widgets/histogram/model.js
javierarce: Ok, I like your suggestion. Thanks for the explanation!

PR #826, opened 2015-11-18
Omit the bbox param in the first call to the endpoint

javierarce:


javierarce:
@viddo, @rochoa: have a look, please.

rochoa:
LGTM :+1:

viddo:
IMHO, this is really a sole concern concern for the model, the view should not need to know/care. I'd also like to have some explanation for any future dev to understand _why_ this was added. 

Since we now how un updated version of Backbone/underscore why not utilize the [`model.once()`](http://backbonejs.org/#Events-once) listener to simplify this? i.e.:
```js
// in histogram/mode..js
…

   url: function() {
     …

     if (this.get('boundingBox') && this.get('submitBBox')) {
       params.push('bbox=' + this.get('boundingBox'));
     }

   …
 
   initialize: function(attrs, opts) {
    // BBox should only be included until after the first fetch, since we want to get the range of the full dataset
    this.once('change:data', function() {
      this.model.set('submitBBox', true);
    });
```

javierarce:
@viddo: much better, thanks!

PR #831, opened 2015-11-19
Consolidate widgets view creation

viddo:
Widget is now created in one place (the factory), removing unnecessary inheritance and boilerplate code.
The added types (done in vis/vis.js) defines their own context (the content-views). I've already mentioned these ideas before but I think it should be clearer what I meant with this small PR

@xavijam @alonsogarciapablo can you please review?


xavijam:
For me perfect :+1:, I guess I'll have some conflicts with "category-search" branch.

xavijam:
I love factories :factory:

PR #833, opened 2015-11-20
Infowindows

piensaenpixel:
@javisantana @xavijam


xavijam:
In my opinion we should review the name of the CDB namespace in the future, nothing important right now, so, go ahead!

piensaenpixel:
+1 @xavijam

PR #834, opened 2015-11-20
remove styles

piensaenpixel:
take a look @javisantana  @xavijam


xavijam:
Why do we remove those ones?

piensaenpixel:
because we don't need the shadows in the scroll now

PR #836, opened 2015-11-20
Removes dependency between chart and mini chart

javierarce:
This PR removes the previous dependency between the two histograms.


javierarce:
Can you have a look, @viddo?

viddo:
:+1:

PR #841, opened 2015-11-23
Sanitized attribution to avoid XSS

acanimal:
reviewer @xavijam  cc @alonsogarciapablo @viddo @javisantana


xavijam:
Sanitize for the win :)

:+1:

PR #843, opened 2015-11-23
Adds temporal axis

javierarce:
This PR adds a temporal axis for the time-series widget and a couple of minor fixes:

- Allows to customize the default properties of the chart.
- In the content view for the regular histogram `chart` is now called `HistogramChartView` to use the same notation as in the time-series widget.

The kind of axis is decided [here](https://github.com/CartoDB/cartodb.js/pull/843/files#diff-d4bb4a9a7868ca9f9ea965ebffaac110R602) depending on the `type` of the chart.


javierarce:
@viddo: can you have a look, please?

viddo:
:+1: 
I think we should look into how to split up chart in smaller components though (e.g. I think we should be able to have two separate x-labels implementations (numeric vs time), but let's talk about that separately

javierarce:
Ok! Thanks!

PR #844, opened 2015-11-23
Implemented torque time-series widget

viddo:
torque

Waypoint for #786 
This implements the torque time-series widget, i.e. animation control; pause, play and time-marker

Noteworthy: 
- Separated examples (torque vs normal)
- Implemented it as separate views, building on top of existing histogram view but w/o modifying its internals, to avoid unwanted side-effects for default vs. torque time-series when doing changes further ahead.
- Some minimal testing, for starters

@javierarce can you review?
cc @javisantana for the torque layer model changes


src/geo/ui/widgets/time-series/histogram-view.js
```diff
@@ -0,0 +1,102 @@
...
+        top: 4,
+        right: 4,
+        bottom: 20,
+        left: 4
+      },
+      y: 0,
+      handles: true,
+      width: this._histogramChartWidth(),
+      height: this.viewModel.get('histogramChartHeight'),
+      data: this.model.getData()
```
viddo: @javierarce There are some options here that perhaps could be defaults inside the chart (e.g. `y: 0` and `animationSpeed: 100`), instead of having to pass them explicitly for all instances?

src/geo/ui/widgets/time-series/torque-histogram-view.js
```diff
@@ -0,0 +1,124 @@
...
+        20 + // bottom labels
+        4 // margins
+    });
+    this._viewModel.bind('change:width', this._onChangeWidth, this);
+    this.add_related_model(this._viewModel);
+
+    // TODO w/o this the timemarker ends up behind the removed-and-added elements in the histogram chart,
+    // can we avoid that?
+    this.model.bind('change:data', this.render, this);
+    this._viewModel.bind('change:width', this.render, this);
```
viddo: @javierarce This is far from ideal, but a necessary "hack" because the histogram chart adds/removes its element on data changes. Without this which causes the time-marker to be rendered behind the re-added elements… can you think of some way to avoid this?

E.g. considering the top-level elements (in the histogram) should always be present, could we refactor the view to utilize d3's [`selection.enter/exit`](https://github.com/mbostock/d3/wiki/Selections#enter) methods to manage internal DOM instead to avoid re-creating elements perhaps?

src/geo/ui/widgets/time-series/histogram-view.js
javierarce: `animationSpeed` is now a default (the current value is 750, which for the time-series widget is too big) but ok to move the `y`.

src/geo/ui/widgets/time-series/torque-histogram-view.js
javierarce: Do you need to re-render the widget every time the data is changed? In the case of the plain histogram I only render it once and then use the `refresh` and `reset` methods when I want to animate the height of the bars to the new values or regenerate the whole chart respectively.

javierarce:
:+1: :+1:

PR #845, opened 2015-11-23
Flatten hierarchy of layer group view classes (Leaflet and GMaps)

### LINKED ISSUE #816: Flatten hierarchy of classes for "layer group views"
- [x] Unify `src/geo/gmaps/cartodb-layer-group-base.js` and `src/geo/gmaps/cartodb-layer-group-gmaps.js` into `src/geo/gmaps-cartodb-layer-group-view.js` (inheritance is not needed).
- [x] Unify `src/geo/leaflet/leaflet-cartodb-group-layer-base.js` and `src/geo/leaflet/leaflet-cartodb-layer-group-view.js` (inheritance is not need).
- [x]  Make the interface for those two classes a bit more consistent (eg: binding to model events in the same place).

alonsogarciapablo:
Fixes #816.

Collapses the following two hierarchies, which didn't make much sense since there was only one subclass in both cases.

```
leaflet-cartodb-group-layer-base
  ^
  |
leaflet-cartodb-layer-group-view
```
And:

```
cartodb-layer-group-base
  ^
  |
cartodb-layer-group-gmaps
  ^
  |
gmaps-cartodb-layer-group-view
```

CR: @viddo 
FYI: @javisantana, @fdansv.


viddo:
<3 code-wise looks legit, although I have to admit it's a bit hard to really tell but just jumping back-and-forth between the changed files (couldn't see any obvious mistakes). Perhaps the others can give more info if there are any unwanted side-effects/implications this might have?

alonsogarciapablo:
@viddo I feel you! For the implications... I've basically tried the example under `examples/index.html` both for Leaflet and GMaps and everything seems to work fine. Tests also helped me find some mistakes. Of course, I might (hopefully not) have broken something...

@javisantana I'll love to get your feedback on this one. Thanks!

javisantana:
Yes, this change makes sense, as you said, since we only have one class does not make any sense to have a base class to share stuff

PR #849, opened 2015-11-23
Add support for jsdoc [WIP]

### LINKED ISSUE #820: JSDoc integration
Configure task to generate JSDoc and start documenting code using jsdoc notation.

viddo:
Resolves #820 

Kudos to @acanimal for creating the initial changes :clap: , I've just updated in widgets HEAD 

I've also added [documentation.js](documentation.js.org) as alternative output generator, IMHO its UI is arguably better, in addition to [_why use documentation instead of JSDoc?_](https://github.com/documentationjs/documentation/blob/master/docs/FAQ.md#why-use-documentation-instead-of-jsdoc)

Docs can be seen after generating them:
```bash
npm run jsdoc; npm run jsdoc2
```


package.json
```diff
@@ -57,7 +57,13 @@
     "grunt-prompt": "~1.3.0",
     "grunt-replace": "0.6.2",
     "grunt-sass": "~1.1.0-beta",
+    "gulp": "3.8.10",
+    "gulp-iconfont": "1.0.0",
+    "gulp-iconfont-css": "0.0.9",
+    "gulp-install": "0.2.0",
+    "gulp-sketch": "0.0.7",
     "jquery": "2.1.4",
+    "jsdoc": "^3.4.0",
```
viddo: Compare the output of the two output generators:
jsdoc: screen shot 2015-11-23 at 18 02 59
---
documentation.js: screen shot 2015-11-23 at 18 03 41
---

IMHO the 2nd (documentation.js) yields better result, primary because:
- it's more concise and legible output (at least by default?)
- it also has a search filter to easily find classes/methods without having to scroll or search inline

src/windshaft/filters/base.js
```diff
@@ -1,12 +1,37 @@
...
+/**
+ * @classdesc Abstract base class for all filters. **Do not use directly**
+ *
+ * Any filter must implement at
+ * least the operations `isEmpty()` and `toJSON()`.
+ *
+ * @class
+ * @extends Model
+ */
+var FilterBase = Model.extend( /** @lends FilterBase.prototype */ {
```
viddo: As can be seen in the [screenshots comparing the output](https://github.com/CartoDB/cartodb.js/pull/849#discussion_r45630364), neither tool resolves namespace (directory path or assignment in the bundle entry file, I.e. `cdb.windshaft.filters.FilterBase = require('./windshaft/filters/base');`. I think this should be resolved before merging since one file/class name might be reused in various context (e.g. the widget models…)

package.json
acanimal: I think the tool does not mind, the important is we have a minimum documented API.

JSDoc is really powerful and allowed to create templates with filter, ie, http://openlayers.org/en/v3.10.1/apidoc/, but if by default the output of documentation is better I agree to use it and does not spent time on the output.

src/windshaft/filters/base.js
acanimal: I'll work on it.

package.json
acanimal: At first I can see an issue with documentationjs and it is it does not document classes that jsdoc does (see: https://github.com/CartoDB/cartodb.js/blob/widgets-jsdoc/src/core/view.js#L7 or https://github.com/CartoDB/cartodb.js/blob/widgets-jsdoc/src/geo/ui/widgets/widget-view.js#L6).

Jsdoc documents the class because the comments on the `constructor` method. It is not a great problem and can be solved adding the `@class` to the class. In addition I see there are no commnets from `@classdesc` tag.

src/geo/ui/widgets/widget-view.js
```diff
@@ -3,26 +3,42 @@ var WidgetLoaderView = require('./standard/widget_loader_view');
...
+ * It contains:
+ * - view model (viewModel)
+ * - data model (dataModel)
  *
  *  It will offet to the user:
  *  - get current data (getData)
  *  - Sync or unsync widget (sync/unsync), making the proper view
  *  listen or not changes from the current datasource.
+ *
+ * @class geo.ui.widgets.WidgetView
```
viddo: I don't agree that this object definition belongs in the source file itself. I think the namespace usage in e.g. core/view above makes sense and that's what I'd expect here too, and that the top-level file cartodb.js references the source to point to where the source can be found.

As a reference, I'd think what [Atom](https://atom.io/docs) is quite exemplary. They have all top-level definitions in one place and reference the source files where necessary, see [AtomEnvironment `atom`](https://atom.io/docs/api/v1.2.4/AtomEnvironment) (which would correspond to our `cdb`/`cartodb` object) vs. what they call an "essential" model (which corresponds to a specific source file) e.g. [`atom.workspace`](https://atom.io/docs/api/v1.2.4/AtomEnvironment#instance-workspace) links to [`Workspace`](https://atom.io/docs/api/v1.2.4/Workspace).

acanimal: That explanation corresponds to that class and IMO is good to have it at the same file. Of course we can also document each namespace describing briefly the goal of the set of classes, types, etc it contains.

viddo:
This was really automatically closed since `widgets` branch was deleted (merged into `v4`), `v4` needs to be merged with it to be able to continue work here.

acanimal:
After our last meeting I think it has not much sense continue this. Maybe the public API once we know the solution we use for documenting.

viddo:
okie doki :+1:

voltagex:
Just as an observer and user of CartoDB, the more documentation on the object / class, the more useful my IDE (WebStorm and in some cases Visual Studio) becomes. Apologies if this isn't the right place to make this comment.

PR #850, opened 2015-11-24
Triggers an event when the infowindow is loaded and ready

javierarce:
Triggers an event when the infowindow is loaded and it's been added to the dom so listeners of the layer are notified when they can interact with the infowindow.


javierarce:
cc: @alonsogarciapablo @javisantana (I'm having some trouble to create an spec for this)

javisantana:
:es:

PR #859, opened 2015-11-25
Don't let several category widgets be colorized at the same time for the same layer

xavijam:
If there are, for example, 2 category widgets (A and B) for the same layer, we should reset category colors from widget A if widget B wants to apply its category colors.

Trying to avoid this one:
screen shot 2015-11-25 at 17 27 43

Any advice?


xavijam:
Take a look @alonsogarciapablo (please :) ) due to the fact that it "touches" cartodb-layer class.

xavijam:
Needed for BI

src/geo/ui/widgets/widgets_collection.js
```diff
@@ -0,0 +1,41 @@
...
+module.exports = Backbone.Collection.extend({
+
+  initialize: function() {
+    this._initBinds();
+  },
+
+  _initBinds: function() {
+    // If a category model applies the category colors, rest should remove/disable
+    // the category colors applied before.
+    this.bind('change:categoryColors', function(m, isColorCategorized) {
```
alonsogarciapablo: Since we have the model and we know the name of the attribute that changed, we could get rid of `isColorCategorized` and use `m.get('wadus')` instead?

src/geo/ui/widgets/widgets_collection.js
```diff
@@ -0,0 +1,41 @@
...
+ *
+ */
+
+module.exports = Backbone.Collection.extend({
+
+  initialize: function() {
+    this._initBinds();
+  },
+
+  _initBinds: function() {
```
alonsogarciapablo: I like this approach better :+1:

src/geo/ui/widgets/widgets_collection.js
alonsogarciapablo: As a bonus, we could extract this logic and reuse it for histograms too :smile:

xavijam: isColorCategorized is the value of the attribute "wadus".

test/spec/geo/ui/widgets/widgets_collection.spec.js
```diff
@@ -0,0 +1,30 @@
...
+  describe('bind', function() {
+
+    it('should disable categoryColors attribute if any other widget enables it', function() {
+      this.collection.reset([{ name: 'hello' }, { name: 'howdy' }, { name: 'hey', categoryColors: true }]);
+      var mdl1 = this.collection.at(0);
+      var mdl2 = this.collection.at(1);
+      var mdl3 = this.collection.at(2);
+      expect(mdl3.get('categoryColors')).toBeTruthy();
+      mdl1.set('categoryColors', true);
+      expect(mdl3.get('categoryColors')).toBeFalsy();
```
alonsogarciapablo: I'd still add an assertion here to ensure that:
```
expect(mdl1.get('categoryColors')).toBeTruthy();
```
This way we're making sure the attribute (categoryColors) of the model that triggered the event is not being set to `false`.

alonsogarciapablo:
Sorry @xavijam. I didn't see this coming through. I left some minor comments, but overall I think is :+1:

src/geo/ui/widgets/widgets_collection.js
alonsogarciapablo: Yes, something like:

```
this._initBinds = function() {
  this.bind('change:histogramSizes', this._disableChangedAttributeOfOtherModels, this);
  this.bind('change:wadus', this._disableChangedAttributeOfOtherModels, this);
}

this.__disableAttributeOfOtherModels = function(model) {
  var attribute = _.keys(model.changedAttributes())[0];
  if (model.get(attribute)) {
    this.each(function(mdl) {
      if (mdl !== model && mdl.get(attribute)) {
        mdl.set(attribute, false);
      }
    })
  }  
}
```

src/geo/ui/widgets/histogram/model.js
```diff
@@ -42,14 +42,10 @@ module.exports = WidgetModel.extend({
...
-    this.bind('change:url', function(){
-      if (this.get('sync')) {
-        this._fetch();
-      }
-    }, this);
-    this.bind('change:boundingBox', function() {
-      if (this.get('bbox')) {
-        this._fetch();
+    WidgetModel.prototype._onChangeBinds.call(this);
+    this.bind('change:histogramSizes', function(mdl, isSizesApplied, d) {
```
alonsogarciapablo: Just double checking here... Is this required? Is anybody listening to the `histogramSizes` that this callback is triggering? The `d` variable in the callback doesn't seem to be used.

xavijam: Yes, this is necessary for CartoD1B thing. We will talk later about it.

PR #870, opened 2015-11-26
Added new category stats

xavijam:
Close #870


xavijam:
Should it be the percentage over the total of the categories you are viewing in your screen? @saleiva 

screen shot 2015-11-26 at 17 45 19

PR #872, opened 2015-11-27
Ported changes on sanitization from develop to v4

acanimal:
reviewer @xavijam


xavijam:
Yay, thanks for taking care of it ;).

PR #876, opened 2015-11-27
Fixes size of icons in the Blue Bar:tm:

alonsogarciapablo:
screen shot 2015-11-27 at 16 06 51

cc: @piensaenpixel


piensaenpixel:
:dancer:

PR #877, opened 2015-11-27
Animate categories with new animation class

javierarce:
Animate categories with new animation class.


javierarce:
retest this please

javierarce:
@xavijam: can you review this, please? 

I'm not sure about the location of the `animate_value.js` file. It needs to be a separate class (because of its size and the places where it needs to be used) but it's not so important be core.

xavijam:
I'd leave it where it is, within widgets folder.

javierarce:
Ok, thanks! It's ok if I merge this, right?

xavijam:
:+1:

PR #883, opened 2015-11-30
Don't fetch widget data when it is collapsed

xavijam:
Also, when collapsed is disabled again, fetch the widget data if the ```url``` or the ```boundingBox``` attribute has changed since last time. And collapsed state will be only 22px of height.

Idea by @saleiva :clap:.

collapsed-widgets

### DIFF
--- a/src/geo/ui/widgets/category/model.js
+++ b/src/geo/ui/widgets/category/model.js
@@ -74,7 +74,7 @@ module.exports = WidgetModel.extend({
     this.bind('change:boundingBox', function() {
       // If a search is applied and bounding bounds has changed,
       // don't fetch new raw data
-      if (this.get('boundingBox') && !this.isSearchApplied() && !this.isCollapsed()) {
+      if (this.get('bbox') && !this.isSearchApplied() && !this.isCollapsed()) {
         this._fetch();
       }
     }, this);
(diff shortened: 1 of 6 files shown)


saleiva:
22px doesnt seem correct. We can review tomorrow ;)

xavijam:
:joy: I hate you!

saleiva:
:*

xavijam:
@javierarce could you take a look please? Just in case I'm missing something from histograms?

javierarce:
:+1: they work fine :)

src/geo/ui/widgets/category/model.js
```diff
@@ -86,6 +86,19 @@ module.exports = WidgetModel.extend({
       });
     }, this);
 
+    this.bind('change:collapsed', function(mdl, isCollapsed) {
```
alonsogarciapablo: This is also duplicated [here](https://github.com/CartoDB/cartodb.js/pull/883/files#diff-214e2a6e2deb1d445b62d408f340b33eR55). Do we need it here? Just double checking...

xavijam: Function _onChangeBinds is being overriden in the category model.

PR #887, opened 2015-12-01
Implement widgets dashboard layouts

viddo:
Implemented missing concepts to handle the dashboard layout. This is necessary to remove the temporary hack for time-series widget where the torque layer had to "taken out" of the normal rendering, to instead allow a layer to have multiple widgets but in different places (for the moment there's two; sidebar or below map). 

screen shot 2015-12-01 at 16 00 38
_A torque layer having multiple widgets in various locations, here a time-series widget + category of twitter usernames (actual names redacted)._

I've tried to minimize the changes of existing `vis.load` code and since the dashboard really "wraps" it, moving it out to the parent `createVis` method (where `vis.load` is called) worked well.

Also, while widgets are related to layers, having individual widgets collections on the layers created of unnecessary code (e.g. numerous collections, methods re-triggers), so by removing that the layouts and widget code could be simplified quite a bit (e.g. removing the widget_layers_view completely), KISS.

@alonsogarciapablo can you review please?


viddo:
retest this please

viddo:
retest this please

examples/index.html
```diff
@@ -21,14 +21,7 @@
...
   <body>
-    <div class="Dashboard-canvas js-dashboard">
-      <div class="Dashboard-mapWrapper Dashboard-mapWrapper--margin">
-        <div class="Map-canvas">
-          <div class="Map" id="map"></div>
-        </div>
-      </div>
-      <div class="Dashboard-belowMap js-dashboard-belowMap"></div>
-    </div>
+    <div class="Dashboard-canvas" id="dashboard"></div>
```
alonsogarciapablo: How would this look if the `.Dashboard-canvas` class is added here? Should we ensure our views add this class?

viddo: Yep, already thought about that :point_right: https://github.com/CartoDB/cartodb.js/pull/887/files#diff-d93a93ee72a553cf85b1bcf4ecb61142R20

alonsogarciapablo:
I left some minor comments. Great job @viddo!

My only concern with this PR is that `cartodb.createVis` will never work as before (we probably "broke" it some PRs ago). Also, there was some discussion the other day about wether all this "layout" has to be part of CartoDB.js...

examples/index.html
alonsogarciapablo: :top:

PR #890, opened 2015-12-01
Widgets vizload

acanimal:
reviewer @alonsogarciapablo cc @xavijam @viddo 

The goal is to simplyfy the `vis.load()` method. For the moment:
- I just extracted the code related to widgets creation to a new private method.
- Modified slightly the MapView.create() factory method and fixed an issue on the way it was calling.

I think you are working too in this class so I prefer upload small changes to not create big problems when merging.

### DIFF
--- a/src/geo/map-view.js
+++ b/src/geo/map-view.js
@@ -4,6 +4,26 @@ var log = require('cdb.log');
 var View = require('../core/view');
 var Infowindow = require('./ui/infowindow');
 
+/**
+ * @classdesc Base class for map view implementations. **Do not use directly**.
+ *
+ * A MapView has attached a {@link MapModel} configuration. It is responsible
+ * to hold a concrete map instance (given the specified map provider)
+ * i.e.: leaflet, googlemaps.
+ *
+ * Currently available implementations are {@link LeafletMapView} and {@link GoogleMapsMapView}.
+ *
+ * Given a MapModel instance you can easily create a new concrete implementation
+ * instance using the `create` method:
+ *
+ * @example
+ * var mapView = MapView.create(div, mapModel);
+ *
+ * @abstract
+ *
+ * @param {Object} options  Allowed set of options.
+ * @param {MapModel} options.mapModel MapModel reference
+ */
 var MapView = View.extend({
 
   initialize: function() {
@@ -204,10 +224,37 @@ var MapView = View.extend({
     return mapViewClass;
   },
 
+  /**
+   * Factory method that given a {@link MapModel} instance with visualization
+   * configuration creates a new concrete view instance depending on the provider.
+   *
+   * @param {HTMLElement} el HTML element where to attach the map.
+   * @param  {MapModel} mapModel MapModel instance
+   * @return {(LeafletMapView|GoogleMapsMapView)}  New MapView's subclass instance.
+   */
   create: function(el, mapModel) {
     var _mapViewClass = MapView._getClass(mapModel.get('provider'));
+
+    //
+    // TODO - Is this really necessary? I have tested in the examples and works
+    // fine without it.
+    //
+    // Another div to prevent leaflet grabbing the div
+    var div_hack = $('<div>')
+      .addClass("cartodb-map-wrapper")
+      .css({
+        position: "absolute",
+        top: 0,
+        left: 0,
+        right: 0,
+        bottom: 0,
+        width: '100%'
+      });
+
+    el.append(div_hack);
+
     return new _mapViewClass({
-      el: el,
+      el: div_hack,
       map: mapModel
     });
   }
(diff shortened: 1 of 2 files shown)


src/geo/map-view.js
```diff
@@ -192,7 +212,19 @@ var MapView = View.extend({
...
+   * @param {HTMLElement} el HTML element where to attach the map.
+   * @param  {MapModel} mapModel MapModel instance
+   * @return {(LeafletMapView|GoogleMapsMapView)}  New MapView's subclass instance.
+   */
+  create: function(el, mapModel) {
+
+    // Choose a concrete subclass implementation
+    var provider = mapModel.get('provider');
     var mapViewClass = cdb.geo.LeafletMapView;
     if(provider === 'googlemaps') {
```
alonsogarciapablo: This is something that could probably be extracted from here. I guess we could check this when we parse the viz.json. If the provider is googlemaps and google is not defined, it doesn't make sense to get to this point. Right?

src/vis/vis.js
```diff
@@ -301,9 +305,11 @@ var Vis = View.extend({
       return this;
     }
 
-    // load modules needed for layers
+    // Load modules needed for layers
     var layers = data.layers;
 
+    // TODO - Is the module loading necessary. We are using browserify so we
+    // know dependencies are loaded.
     if (!this.checkModules(layers)) {
```
alonsogarciapablo: This is loading torque.js only when there's a layer. AFAIK the motivation is to not include torque.js inside of cartodb.js and load it dynamically instead.

src/vis/vis.js
```diff
@@ -427,14 +437,17 @@ var Vis = View.extend({
...
       height: '100%'
     });
-
-    this.container = div;
-
+    this.$el.append(div);
+    //
+    // TODO - Is this really necessary? I have tested in the examples and works
+    // fine without it.
+    //
```
alonsogarciapablo: Not sure about this. I'd test it a bit more and if you think we can remove it, I'd go ahead and remove it. You can also ask whoever did it in the first place, but I let's figure it out now instead of adding another TODO.

src/vis/vis.js
```diff
@@ -509,7 +516,47 @@ var Vis = View.extend({
...
+   *
+   * @private
+   * @param  {Object} vizconfig         Object with viz data configuration.
+   * @param  {Array.<Object>} interactiveLayers Array with references to
+   * the layers withint the viz config.
+   * @param  {(CartoDBLayerGroupNamed|CartoDBLayerGroupAnonymous)} cartoDBLayerGroup Layer group reference.
+   * @return {boolean}  True if any layer contains widgets and has been initialized,
+   * false otherwise.
+   */
+  _initializeWidgets: function(vizconfig, interactiveLayers, cartoDBLayerGroup) {
```
alonsogarciapablo: This method is doing much more than initialising the widgets (no single responsibility here)... don't you think?  For example, it's creating the `DashboardInfoView` which has nothing to do with widgets. So it's hard for me to know what it does just by looking at it's name, and also to guess it returns a boolean (methods that return a boolean usually looking more like `_isSomething`, `_hasSomething`, `_isASomething`, etc.)

src/vis/vis.js
```diff
@@ -581,31 +628,16 @@ var Vis = View.extend({
...
+    var result = false;
     if (layersWithWidgets.size() > 0) {
-      setTimeout(function() {
-        self.mapView.invalidateSize();
-      }, 0);
+      result = true;
     }
-
-    return this;
+    return result;
```
alonsogarciapablo: I guess this would be the same as `return layersWithWidgets.size() > 0` :smile:

src/geo/map-view.js
```diff
@@ -192,7 +212,19 @@ var MapView = View.extend({
...
+  
+  /**
+   * Factory method that given a {@link MapModel} instance with visualization
+   * configuration creates a new concrete view instance depending on the provider.
+   *
+   * @param {HTMLElement} el HTML element where to attach the map.
+   * @param  {MapModel} mapModel MapModel instance
+   * @return {(LeafletMapView|GoogleMapsMapView)}  New MapView's subclass instance.
+   */
+  create: function(el, mapModel) {
```
alonsogarciapablo: :+1:

src/geo/map-view.js
acanimal: Yes, I think so. The class responsible to parse the vizjson would get the provider and create the right map implementation (leaflet, gmaps, etc) passing the rest of vizjson params.

src/vis/vis.js
acanimal: OMG !!! Obfuscation time :(

src/vis/vis.js
acanimal: Yes, I agree. The goal was simply divide the load method into smaller pieces.

PR #897, opened 2015-12-02
fix extra space in ipad view

piensaenpixel:
@xavijam take a look.
Fix this one https://github.com/CartoDB/cartodb.js/issues/896


themes/scss/map/dashboard-info.scss
```diff
@@ -43,12 +43,12 @@
 }
 .Dashboard-info {
   @include transition(width, 0.25s, cubic-bezier(0.01, 0.99, 0.97, 1.01));
-  position: absolute;
+  position: fixed;
```
xavijam: For mobile view should work as it is?

PR #900, opened 2015-12-02
Fix torque time-series control icons

### LINKED ISSUE #894: Torque time-series controls are not rendered correctly in Safari

viddo:
Resolves #894

Makes the icons appear consistent in all evergreen browsers. Perhaps we should consider applying the same approach for the other (round-kind) icons too, and consolidate the components a little?

Anyway, the code changes should speak for themselves, added inlined comments for non-obvious props though. Also fixed the styleguide to be coherent with the cartodb.css (removing default resets), so things actually will look the same in both styleguide and examples.

@xavijam can you review plz? cc @piensaenpixel @javierarce

### DIFF
--- a/themes/styleguide/layout-loading.html
+++ b/themes/styleguide/layout-loading.html
@@ -20,17 +20,17 @@
 
 <body>
 
-  <div class="Dashboard-canvas">
-    <div class="Dashboard-mapWrapper">
-      <div class="Map-canvas">
-        <div class="Map" id="map">
+  <div class="CDB-Dashboard-canvas">
+    <div class="CDB-Dashboard-mapWrapper">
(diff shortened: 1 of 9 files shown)


themes/styleguide/layout-loading.html
```diff
@@ -20,17 +20,17 @@
 
 <body>
 
-  <div class="Dashboard-canvas">
-    <div class="Dashboard-mapWrapper">
-      <div class="Map-canvas">
-        <div class="Map" id="map">
+  <div class="CDB-Dashboard-canvas">
```
xavijam: Ouch, I missed this :_), thanks.

xavijam:
:+1:

Issue #880, opened 2015-11-27
Remove time-series hacks to support histogram


viddo:
They were necessary to move forward with #786 
Once https://github.com/CartoDB/Windshaft/issues/405 is resolved they can be removed

viddo:
merged in #901

rochoa:
:smile: thanks again!

PR #901, opened 2015-12-02
Widgets timeseries filtered from meta

rochoa:
cc @javierarce @viddo as you were working on histogram and time-series

### DIFF
--- a/src/geo/ui/widgets/time-series/torque-histogram-view.js
+++ b/src/geo/ui/widgets/time-series/torque-histogram-view.js
@@ -80,7 +80,6 @@ module.exports = View.extend({
 
   _onBrushEnd: function(loBarIndex, hiBarIndex) {
     var data = this.model.getData();
-    this._rangeFilter.set('columnType', this._torqueLayerModel.get('meta').column_type);
     this._rangeFilter.setRange(
       data[loBarIndex].start,
       data[hiBarIndex - 1].end
--- a/src/geo/ui/widgets/time-series/torque-histogram-view.js
+++ b/src/geo/ui/widgets/time-series/torque-histogram-view.js
@@ -80,6 +80,7 @@ module.exports = View.extend({
 
   _onBrushEnd: function(loBarIndex, hiBarIndex) {
     var data = this.model.getData();
+    this._rangeFilter.set('columnType', this._torqueLayerModel.get('meta').column_type);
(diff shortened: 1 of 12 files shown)


src/geo/ui/widgets/time-series/torque-histogram-view.js
```diff
@@ -80,6 +80,7 @@ module.exports = View.extend({
 
   _onBrushEnd: function(loBarIndex, hiBarIndex) {
     var data = this.model.getData();
+    this._rangeFilter.set('columnType', this._torqueLayerModel.get('meta').column_type);
```
rochoa: As I pointed in the commit itself, I'm not happy with this approach. Any help to improve this is welcome.

viddo: The meta prop is not available initially, right?
I think this detail is not a concern for the view but the model, which knows about both the filter and layer.

viddo:
Overall looks good :+1: 
Can't get the examples to work locally though, so could not test it. :/

javierarce:
:+1: I have the same issue with the examples as Nicklas, but code looks fine.

src/windshaft/dashboard.js
```diff
@@ -56,17 +56,18 @@ WindshaftDashboard.prototype._createInstance = function(options) {
...
 
-      this._updateWidgetURLs({
-        layerId: options.layerId
-      });
-
       // update other kind of layers too
-      this.layers.each(function(layer) {
+      this.layers.each(function(layer, layerIndex) {
         if (layer.get('type') === 'torque') {
+          layer.set('meta', dashboardInstance.getLayerMeta(layerIndex));
```
viddo: The meta information that is required to have the column type only works for Torque for now? I guess this is related to why the default time-series widget doesn't work?

viddo:
Btw,this resolves #880 right?

src/windshaft/dashboard.js
rochoa: Yes, that kind of information is only provided for torque layers, and at the same time is required by time-series widget associated to torque layers.

We need to review this use case when using time-series widget with non-torque layers.

rochoa:
@viddo Yes, I updated the example to not use the hacks.

rochoa:
And ... :ship: ing :it: Thx @viddo for your help!

PR #903, opened 2015-12-03
Don't disable drag/pan of the map if device is mobile

### LINKED ISSUE #902: Check if mobile view is enabled in order to not disable drag/pan
https://github.com/CartoDB/cartodb.js/commit/de6b8c16d3ea1725c9a520182b8c71074703a620#diff-f99ed960498d5ea8717b3e74d54bfb01R341

cc @acanimal.

xavijam:
No matter if scrollwheel is enabled or if zoom overlay is added.

Fixes #902.

REVIEWER: @alonsogarciapablo

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
       "grunt-contrib-csslint": "~0.2.0",
       "grunt-contrib-cssmin": "~0.7.0",
       "grunt-contrib-htmlmin": "~0.1.3",
-      "grunt-contrib-imagemin": "~0.9.4",
+      "grunt-contrib-imagemin": "~1.0.0",
       "grunt-contrib-jshint": "~0.8.0",
       "grunt-contrib-uglify": "~0.3.0",
       "grunt-replace": "0.6.2",
--- a/src/vis/vis.js
+++ b/src/vis/vis.js
@@ -158,7 +158,7 @@ var Vis = cdb.core.View.extend({
       legends: legends
     });
 
-    if (!this.mobile_enabled) {
+    if (!this.isMobileEnabled) {
       this.mapView.addOverlay(this.legends);
     }
   },
@@ -336,13 +336,18 @@ var Vis = cdb.core.View.extend({
     var scrollwheel       = (options.scrollwheel === undefined)  ? data.scrollwheel : options.scrollwheel;
     var slides_controller = (options.slides_controller === undefined)  ? data.slides_controller : options.slides_controller;
 
-    // Do not allow pan map if zoom overlay and scrollwheel are disabled
+    // Do not allow pan map if zoom overlay and scrollwheel are disabled unless
+    // mobile view is enabled
+    var isMobileDevice = this.isMobileDevice();
     // Check if zoom overlay is present.
     var hasZoomOverlay = _.isObject(_.find(data.overlays, function(overlay) {
       return overlay.type == "zoom"
     }));
 
     var allowDragging = hasZoomOverlay || scrollwheel;
+    if (isMobileDevice) {
+      allowDragging = true;
+    }
 
     // map
     data.maxZoom || (data.maxZoom = 20);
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -38,7 +38,7 @@
       "grunt-contrib-csslint": "~0.2.0",
       "grunt-contrib-cssmin": "~0.7.0",
       "grunt-contrib-htmlmin": "~0.1.3",
-      "grunt-contrib-imagemin": "~0.9.4",
+      "grunt-contrib-imagemin": "~1.0.0",
```
alonsogarciapablo: Not sure if this ^^ is required for this PR.

src/vis/vis.js
```diff
@@ -336,13 +336,18 @@ var Vis = cdb.core.View.extend({
...
-    // Do not allow pan map if zoom overlay and scrollwheel are disabled
+    // Do not allow pan map if zoom overlay and scrollwheel are disabled unless
+    // mobile view is enabled
+    var isMobileDevice = this.isMobileDevice();
     // Check if zoom overlay is present.
     var hasZoomOverlay = _.isObject(_.find(data.overlays, function(overlay) {
       return overlay.type == "zoom"
     }));
 
     var allowDragging = hasZoomOverlay || scrollwheel;
```
alonsogarciapablo: This could work too:

```
var allowDragging = isMobileDevice || hasZoomOverlay || scrollwheel;
```

test/spec/vis/vis.spec.js
```diff
@@ -444,43 +446,55 @@ describe("Vis", function() {
...
-          }
-        ],
-      };
+      this.vis = new cdb.vis.Vis({el: container});
+    });
 
-      vis.load(mapConfig);
-      expect(vis.map.get('drag')).toBeTruthy();
+    it("should be enabled with zoom overlay and scrollwheel enabled", function() {
+      var mapConfig = mapConfigCustom(true, true);
```
alonsogarciapablo: I find `mapConfigCustom(true, true);` a bit hard to understand because of the boolean params.

It would be easier to understand with a single hash param. Something like:

```
var mapConfig = createMapConfig({
  scrollWheel: true,
  zoom: false
});
```

package.json
xavijam: Without this thing I can't run any grunt task :(.

src/vis/vis.js
xavijam: ok!

test/spec/vis/vis.spec.js
xavijam: ok!

xavijam:
Better?

test/spec/vis/vis.spec.js
```diff
@@ -444,43 +446,73 @@ describe("Vis", function() {
...
+      this.vis.load(mapConfig);
+      expect(this.vis.map.get('drag')).toBeFalsy();
+    });
+
+    describe('over mobile', function() {
+      beforeEach(function() {
+        spyOn(this.vis, 'isMobileDevice').and.returnValue(true);
+      });
+
+      it("should be enabled if scrollwheel is enabled (no matter zoom overlay) but device is mobile", function() {
```
alonsogarciapablo: We're missing some tests for other possible combinations here. Example: `{ scrollwheel: false, zoom: true }` and `{ scrollwheel: true, zoom: true }`. I think we're missing one when isMobileDevice is false.

xavijam: isMobileDevice is false always :). Zoom is not being taken into account for mobile, as the tests title say.

xavijam: https://github.com/CartoDB/cartodb.js/pull/903/files#diff-761c59c077b69a8c5ce34c67faa86861R39 ^^^

test/spec/vis/vis.spec.js
```diff
@@ -444,43 +446,73 @@ describe("Vis", function() {
...
 
     it("should be enabled with zoom overlay and scrollwheel disabled", function() {
-      var container = $('<div>').css('height', '200px');
-      var vis = new cdb.vis.Vis({el: container});
+      var mapConfig = createMapConfig({
+        scrollwheel: false,
+        zoom: true
+      });
+      this.vis.load(mapConfig);
+      expect(this.vis.map.get('drag')).toBeTruthy();
```
alonsogarciapablo: The fact that we have to use `this.vis.map` in this assertion might be telling us that it's probably the map the one who has to decide if dragging is enabled based on some of it's attributes (we would need to pass more stuff to the map). Just an observation...

cc// @acanimal

xavijam:
So? Happy to merge?

xavijam:
ping! now?

PR #904, opened 2015-12-03
Refactor the chart hide/show methods

javierarce:
This PR moves the hide/show logic to a model-based one. It also fixes a Firefox bug that prevented hiding an histogram right after was appended to the dom.

### DIFF
--- a/src/geo/ui/widgets/histogram/chart.js
+++ b/src/geo/ui/widgets/histogram/chart.js
@@ -512,11 +512,11 @@ module.exports = View.extend({
   },
 
   hide: function() {
-    this.model.set({ display: false });
+    this.model.set('display', false);
   },
 
   show: function() {
-    this.model.set({ display: true });
+    this.model.set('display', true);
   },
 
   _hide: function() {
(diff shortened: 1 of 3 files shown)


javierarce:
Can you review this change, @viddo?

src/geo/ui/widgets/histogram/chart.js
```diff
@@ -492,16 +503,32 @@ module.exports = View.extend({
...
+    }
+  },
+
   hide: function() {
-    this.$el.hide();
+    this.model.set({ display: false });
   },
 
   show: function() {
+    this.model.set({ display: true });
```
viddo: use the single-set instead, i.e. `this.model.set('display', true)`

viddo: (same for hide method above)

viddo:
:+1:

src/geo/ui/widgets/histogram/chart.js
javierarce: :ok_hand:

PR #916, opened 2015-12-10
Fixe issue sanitizing attributions

acanimal:
reviewer @xavijam

### DIFF
--- a/src/geo/leaflet/leaflet.js
+++ b/src/geo/leaflet/leaflet.js
@@ -302,7 +302,7 @@
       attributionControl._attributions = {};
       var newAttributions = this._originalAttributions.concat(this.map.get('attribution'));
       _.each(newAttributions, function(attribution) {
-        attributionControl.addAttribution(attribution);
+        attributionControl.addAttribution(cdb.core.sanitize.html(attribution));
       });
     },
 
--- a/src/geo/leaflet/leaflet.js
+++ b/src/geo/leaflet/leaflet.js
@@ -302,7 +302,7 @@
       attributionControl._attributions = {};
       var newAttributions = this._originalAttributions.concat(this.map.get('attribution'));
       _.each(newAttributions, function(attribution) {
-        attributionControl.addAttribution(cdb.core.sanitize.html(attribution));
+        attributionControl.addAttribution(attribution);
       });
     },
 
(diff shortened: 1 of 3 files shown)


xavijam:
I miss a test ;)

acanimal:
cc @xavijam This fix works both on cartodb.js standalone and editor. But we need to talk about an strange issue on layer models construction.

acanimal:
any news with this.

xavijam:
Could you cc rest of frontend guys explaining the problem you want to solve please?

acanimal:
cc @javierarce @alonsogarciapablo @viddo @xavijam 

This PR tries to salve an issue with sanitization. It is simply but has an impact in the way we coding until now.

## Whats the problem?

Right now we load data as it comes from user (editor) or vizjson and sanitizes the attribute to be use when it is going to be used, ie. tooltips, legends, etc. That is, the data is wrapped in a `xxxModel` class and data is accesed using `this.model.get('someAttrib')`.

In addition, on the editor implementation, for models, there are a subclass that extends or modifies the parent functionality. For example,` cdb.geo.map.CartoDBLayer` has its corresponsing subclass `cdb.admin.CartoDBLayer`. In the editor, data is also accessed using  `this.model.get('someAttrib')`.

The main problem is we need to sanitise any attribute susceptible to be sanitized all time we are going to retrieve it.

## The solution proposed

Sanitize data when it is set in the model so it was valid when we need to use it.

If the idea is to wrap the literal data into a model class, the solution is to sanitize the attributes susceptibles to be sanitized when: 1)model is created from a literal object or 2) when an attribute is set.

Because the initial problem is reported with `attribution` property, I have modified the top most common class that works with attribution, the `cdb.geo.map.MapLayer` class, from which all other layer model classes inherits.

So a have override the `set` method to do the sanitization taks. For the moment only on the `attribution` property but I think it could be made on other properties like `template`.

Note: With this solution, only cartodb.js need to be updated. Otherwise, if we sanitize when attribs are used, I need to modify too cartodb-editor.

Waiting for your opinions.

viddo:
I don't think this is the way to go. Sanitizing is a concern on the _output_, what kind of escaping is necessary depends on where it's going to be presented.

acanimal:
If there aren't more opinions I consider this a :-1:  :( So I close the PR and work solving the bug in another way.

PR #928, opened 2015-12-14
Use standardjs as linter

viddo:
This adds support to enforce [standardjs.com](http://standardjs.com/) syntax through the normal PR-CI workflow we have, as agreed in the offline meeting re: docs.

- See commit 796aea7 ["Test"](https://ci.cartodb.net/job/CartoDB-js-PR-testing/348/) for how the output will look if the linter finds offended rules
- There are linter plugins for the most [common IDEs](http://standardjs.com/#text-editor-plugins)
- Linting can also be run manually by `grunt standard:lint` (same task used in `grunt test`)
- Run `node_modules/.bin/standard --format path/to/file.js` to fix the the most common errors automatically (you might still need to do some manual fixing though), applied in ae6abbb below for example.
- On exceptional occasions one can [suppress warnings](http://standardjs.com/#how-do-i-hide-a-certain-warning)
- [Global objects](http://standardjs.com/#i-use-a-library-that-pollutes-the-global-namespace-how-do-i-prevent-variable-is-not-defined-errors) can be defined, I've added a few for test env (see package.json changes)

**Caveat:** Since the auto-format can't fix _all_ offending code, I've set this up so only _new_ files will be linted from here on (the existing files are ignored, through `.eslintignore`. This way we can improve progressively, and avoid an epic PR and waste time formatting old files.

@xavijam can you review? FYI @CartoDB/frontend


xavijam:
My semicolon :scream: ! :(
.
.
.
.
.
:+1:

alonsogarciapablo:
:+1:

javisantana:
not agree about removing the semicolon:

- all the current code has semicolons
- semicolons are useful

Issue #932, opened 2015-12-14
Design an error message for the cover image of the infowindow


javierarce:
In the current version of the infowindows we have an error message that appear when the URL of a cover image couldn't be loaded. We'd need something similar for the new infowindow style.

nobuti:
There has been no activity on this issue for several months and that's why we are closing it (automatically). If you think this still needs to be addressed please reopen it. The good news is that we are now working on the next version of CARTO.js, stay tuned.

Issue #931, opened 2015-12-14
Design a loader/spinner for the infowindow cover 


javierarce:
We have a loader for the general content of the infowindow, but I think we should have one for the cover images.

piensaenpixel:
I'd use the same one

saleiva:
the actual proposed one is too subtle IMO

piensaenpixel:
maybe we could improve the contrast.

nobuti:
There has been no activity on this issue for several months and that's why we are closing it (automatically). If you think this still needs to be addressed please reopen it. The good news is that we are now working on the next version of CARTO.js, stay tuned.

Issue #899, opened 2015-12-02
New infowindow


xavijam:
- [x] Review image header implementation. Don't delete the image if it is already present in the cover.
- [x] Implement image header with the "hook" image (talk with @piensaenpixel).
- [ ] Add new loader
- [x] Close infowindow when map is clicked.
- [x] Delete old "key" classes like 'cartodb-popup-content' and replace them by new 'js-content' or whatever.
- [x] Close infowindow when the ESC is pressed
- [x] Review infowindow CSS classes (with @piensaenpixel).
- [x] Review whole infowindow.js code (more or less done).
- [x] Fix shadows
- [x] Change scroll library (remove jScrollPane and use perfect-scroll), take into account classes change.

xavijam:
Related -> https://github.com/CartoDB/cartodb.js/issues/846

xavijam:
@javierarce now it is all yours.

javierarce:
I've added a new task (fix shadows) to the list, because the actual one is fake.

xavijam:
Closing reviewing classes because @piensaenpixel already did.

xavijam:
For taking into account:

https://github.com/CartoDB/cartodb.js/issues/683
https://github.com/CartoDB/cartodb.js/issues/800

javierarce:
Added a new task: "add new loader"

nobuti:
There has been no activity on this issue for several months and that's why we are closing it (automatically). If you think this still needs to be addressed please reopen it. The good news is that we are now working on the next version of CARTO.js, stay tuned.

Issue #934, opened 2015-12-15
Remove jscrollpane from the mobile view


javierarce:
And replace it with the perfect scroll.

https://github.com/CartoDB/cartodb.js/blob/v4-module-export/src/geo/ui/mobile.js

PR #935, opened 2015-12-15
New infowindows

javierarce:
This PR solves several tasks described in the original issue: https://github.com/CartoDB/cartodb.js/issues/899

There are two pending related tickets here: 
https://github.com/CartoDB/cartodb.js/issues/932
https://github.com/CartoDB/cartodb.js/issues/931

I propose to merge this and address them separately.


src/geo/ui/infowindow.js
```diff
@@ -369,8 +396,8 @@ var Infowindow = View.extend({
    */
   _isValidURL: function(url) {
     if (url) {
-      var urlPattern = /^(http|ftp|https):\/\/[\w-]+(\.[\w-]+)+([\w.,@?^=%&amp;:\/~+#-|]*[\w@?^=%&amp;\/~+#-])?$/
-      return String(url).match(urlPattern) != null ? true : false;
+      var urlPattern = /^(http|ftp|https):\/\/[\w-]+(\.[\w-]+)+([\w.,@?^=%&amp;:\/~+#-|]*[\w@?^=%&amp;\/~+#-])?$/;
```
javierarce: added ; (not for very long)

javierarce:
cc: @xavijam

xavijam:
:clap: reviewing now. By the way, those tickets in description are the same :dancer: :D.

javierarce:
Oops, changed. It was a very important ticket.

src/geo/ui/infowindow.js
```diff
@@ -21,6 +22,11 @@ var util = require('../../core/util');
 
 var Infowindow = View.extend({
 
+  options: {
+    imageTransitionSpeed: 300,
+    hookHeight: 16
+  },
+
   className: "cartodb-infowindow",
```
xavijam: Do you think this class should be reviewed?

src/geo/ui/infowindow.js
```diff
@@ -307,17 +318,42 @@ var Infowindow = View.extend({
     return false;
   },
 
+  _loadImageHook: function(url) {
```
xavijam: :clap:

xavijam:
Probably this is a historic moment: could we remove jScrollPane library??? 
```redoble de tambor```

File: src/geo/ui/infowindow.js [hunk 4]
```diff
@@ -291,7 +302,7 @@ var Infowindow = View.extend({
    *  Does header contain cover?
    */
   _containsCover: function() {
-    return this.$el.find(".cartodb-popup.header").attr("data-cover") ? true : false;
+    return this.$el.find(".js-infowindow").attr("data-cover") ? true : false;
```
xavijam: this.$('.js-infowindow') works too.

src/geo/ui/infowindow.js
```diff
@@ -327,39 +363,39 @@ var Infowindow = View.extend({
...
         }
       }
-
+      
       $img.css(styles);
-      $img.fadeIn(300);
+      $cover.css({ height: h - self.options.hookHeight });
+      $img.fadeIn(self.options.imageTransitionSpeed);
+
+      self._loadImageHook(url);
```
xavijam: Shouldn't we load/set hook image when there is only image + title?
cc @piensaenpixel

src/geo/ui/infowindow.js
xavijam: :joy:

src/geo/ui/infowindow.js
javierarce: True

javierarce:
> Probably this is a historic moment: could we remove jScrollPane library??? 

I thought that, but this scroll is being used in the [mobile view](https://github.com/CartoDB/cartodb.js/blob/v4-module-export/src/geo/ui/mobile.js)… I created a ticket for that [here](https://github.com/CartoDB/cartodb.js/issues/934)

xavijam:
Ohhh, mobile view, ok!

src/geo/ui/infowindow.js
```diff
@@ -307,17 +318,42 @@ var Infowindow = View.extend({
...
 
     var self = this;
-    var $cover = this.$(".cover");
+
+    var $cover = this.$(".js-cover");
     var $img = $cover.find("img");
-    var $shadow = this.$(".shadow");
     var url = this._getCoverURL();
 
     if ($img.length > 0) {
```
xavijam: If there is already an image defined in the header cover, we should add the class ```CDB-infowindow-media-item``` to that image and we should run ```this._loadImageHook(url);``` before the ```return;```.

xavijam:
Last two things:

- CDB-hook magic should not be applied if the infowindow has any content outside header cover.
- Check what clip-path thing is doing because some svgs are being added to the body when an image infowindow is displayed.

javierarce:
I've created issues for two pending tasks, so merging this branch.

xavijam:
:clap:

PR #936, opened 2015-12-15
Export cartodb.js a module

viddo:
This allows cartodb.js to be used from a dependent module, e.g. https://github.com/CartoDB/deep-insights.js (and probably later for https://github.com/CartoDB/cartodb/issues/6068 to remove the submodule in the editor). E.g. `var cdb = require('cartodb.js')`, this enables more fine-grained control over how the bundle/dependency can be built for each context w/o this repo have to know the details.

I tried to remove/simplify the `window.` usages and simplify the cartodb vs. cartodb.mod.torque bundles creation by using [`standalone` flag](https://github.com/substack/node-browserify#usage) and [`factor-bundle` plugin](https://github.com/substack/browserify-handbook#factor-bundle) but could not make them play nice. Considering cartodb.js is only intended to be used in the browser I think it's fine to try leave it for now, we can iterate and retake that later if the need arises.

Also, some things remove/extracted, was missed when https://github.com/CartoDB/deep-insights.js was extracted:
  - d3 
  - src/core/format (only dependency for d3)
  - copy task not need anymore

@alonsogarciapablo can you review?
cc @rochoa since you did the browserify setup for torque


src/cdb.js
```diff
@@ -1,36 +1,32 @@
 // Creates cdb object, mutated in the entry file cartodb.js
 // Used to avoid circular dependencies
-var cdb = window.cdb = window.cartodb;
+var cdb = {};
 
 // Might be set in different bundles (e.g. cartodb.js vs. torque), avoid overriding it once set
-if (!cdb) {
-  cdb = window.cdb = window.cartodb = {};
-  if (!window.JST) window.JST = {};
```
viddo: Not used here anymore, so :fire: completely, turned out to be a dependency for `cdb.init` in v3. From v4 this should live in the editor, added a note on the issue for that there :point_right: https://github.com/CartoDB/cartodb/issues/6068

src/cdb.js
```diff
@@ -1,36 +1,32 @@
...
+cdb.ui.common = {};
 
-  cdb.vis = {};
+cdb.vis = {};
 
-  cdb.windshaft = {};
-  cdb.windshaft.filters = {};
-}
+cdb.windshaft = {};
+cdb.windshaft.filters = {};
```
alonsogarciapablo: Not sure if we need these guys anymore. They were probably introduced before everything was browserified.

viddo: Totally right, some minor things we've also missed previously, I'll :fire:

alonsogarciapablo:
:+1:

PR #940, opened 2015-12-16
Change linter to semistandard

viddo:
Same as [standard but with semicolons](https://github.com/Flet/semistandard#rules), as requested offline and https://github.com/CartoDB/cartodb.js/pull/928#issuecomment-165036919

@javisantana @alonsogarciapablo OK with you? This PR was a 5-min change so no biggie. 
cc @CartoDB/frontend FYI

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -195,14 +195,14 @@ module.exports = function(grunt) {
 
   grunt.registerTask('standard:lint', 'lint source files', function() {
     var done = this.async();
-    require("child_process").exec('PATH=$(npm bin):$PATH standard', function (error, stdout, stderr) {
+    require("child_process").exec('PATH=$(npm bin):$PATH semistandard', function (error, stdout, stderr) {
       if (error) {
         grunt.log.fail(error);
 
         // Filter out lines that are ignored,
         // e.g. "src/foobar.js:0:0: File ignored because of your .eslintignore file. Use --no-ignore to override."
         grunt.log.fail(stdout.replace(/.+--no-ignore.+(\r?\n|\r)/g, ''));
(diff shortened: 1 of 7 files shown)


alonsogarciapablo:
:+1:, but didn't you guys agree to remove the semicolons? why are we bringing them back?

alonsogarciapablo:
Should we remove the `.eslintignore` if we bring the semicolons back (and we shouldn't have so many warnings)?

viddo:
We agreed to apply standardjs which included removing the semicolons, but judging from responses afterwards I get the perception that not everyone actually agreed in the end (did not read its rules perhaps?). Also, I interpret @javisantana's https://github.com/CartoDB/cartodb.js/pull/928#issuecomment-165036919 as this is a no-go (correct me if I'm wrong).

As this PR shows, luckily it's not a big deal to fix that particular request.

viddo:
Removing `.eslintignore` and applying using autoformat…
```bash
grunt standard:lint | wc -l
    1930
```
…there's still a significant amount of offending rules that need manual fixing. I would not bother with these for now.

javisantana:
My comment is to expose I was not agree but it's not a blocker just an opinion (an opinion + 2 reasons about why keeping semicolons is important)

viddo:
Okie doki, then I close this PR :kissing_heart:

javisantana:
hey, it was an opinion that you should not ignore. I think I'm right

viddo:
Ok, then I misunderstood your comment as "disagree and commit". 
I do not have a strong opinion on the semicolons. So if you want the semicolons let's merge this then?

javisantana:
I take that as a disagree and commit from your side

PR #947, opened 2015-12-17
Instantiate maps when using cartodb.createVis

alonsogarciapablo:
I added a couple of examples that use `cartodb.createVis` and prefixed them with `v4-`. Other examples are pointing to the library hosted on the CDN so they should work. I didn't write any test cause all this is very likely to change in the short term. Also `cartodb.createLayer` is still broken.

cc @viddo @rochoa


viddo:
:+1: but the tests should not be failing should they?

Btw, I think we should rethink the "examples" too, I've already created an issue, they essentially have the same problems as I outlined in https://github.com/CartoDB/deep-insights.js/issues/26. I don't expect this to change here, but mentioning it now because I'd like us to give it some thought going forward.

alonsogarciapablo:
It looks like tests are failing because of the semistandard linter. I'm going to fix the issues.

PR #953, opened 2015-12-17
Loader review

xavijam:
Instead of using fadeIn fadeOut or fadeTo opacity thing, let's use 'is-visible' standard class for CDB-Loader.

REVIEWER: @javierarce 
cc @piensaenpixel


javierarce:
:eyeglasses: :+1:

PR #957, opened 2015-12-18
🔥 backbone elder

rochoa:
Remove core-decorators, aka elder for backbone's model/view/collection.

Now the alternative to elder is `Model.__super__.{method}.call(this, args)`.


src/geo/ui/search.js
```diff
@@ -245,7 +245,7 @@ var Search = View.extend({
   clean: function() {
     this._unbindEvents();
     this._destroySearchPin();
-    this.elder('clean');
+    Search.__super__.clean.call(this);
```
alonsogarciapablo: What about `View.prototype.clean.call(this)`? We're using this pattern already and `__super__` feels a bit hacky to me.

alonsogarciapablo:
I left a comment with an alternative to `__super__`. Other than that... :fire:

src/geo/ui/search.js
viddo: Official docs recommends using this :point_up: approach too, see end of http://documentcloud.github.io/backbone/#Model-extend probably better than relying on internals that might change in the future.

rochoa: :+1: I should have looked into this. Two more reasons to not rely on `__super__`:

 - https://github.com/jashkenas/backbone/pull/3539#issuecomment-82384843
 - https://github.com/jashkenas/backbone/pull/787#issuecomment-3143358

I will take of it.

rochoa:
Are you both OK with latest changes?

viddo:
Sí señor :dancer:

alonsogarciapablo:
:+1:

Issue #695, opened 2015-09-30
Attributions added multiple times


haraldkj:
We are currently experiencing unwanted behaviour when adding attributionControl and a CartoDB-layer to the same map. We want to be able to add controls after layers, but it seems the new behaviour of attributions in Leaflet maps is in conflict with this. Is this intended?  

See this example: 

http://jsfiddle.net/afLgdkuj/3/

This is probably related to https://github.com/CartoDB/cartodb.js/commit/fcf3356540d1164eb1d1277cf9fa38c0708e0917.

alonsogarciapablo:
Hello @haraldkj! Thanks for your example! I can see the `Statens kartverk` attribution is being displayed twice, but I'd say that has to do with the way Leaflet works. If you add another `L.control.attribution`, it will also have that attribution... What's the behaviour you would expect? Or what exactly are you trying to accomplish?

haraldkj:
We currently have a map stack which incorporates several different providers. We would like to add one attribution control after all other layers have been added. We would also like to remove the default leaflet prefix as in the fiddle. 

An example is the map at http://www.nrk.no/norge/her-er-norges-nye-ordforere-1.12561458 which uses an older version of cartodb.js. The updated cartodb.js would result in an extra attributioncontrol on top of the map legend control.

Thanks.

alonsogarciapablo:
Thanks again for the example. I think if you add more layers to the map, attributions should be displayed correctly. I don't see why you want/need to add another attributionControl after all layers have been added... could you please carify a bit more?

alonsogarciapablo:
I might be wrong but, no matter when you add the `L.control.attribution`, it will display the attributions of all layers in the map. Take a look at this:

https://jsfiddle.net/afLgdkuj/5/

haraldkj:
Yes, that makes sense. But it would be convenient to be able to decide when the control is to be added to the map, regardless of adding layers. 

We don't want multiple controllers - we would like all of our attributions in one attribution control. We basically do this:

http://jsfiddle.net/Lt17d3f4/1/

Currently, after the commit to cartodb.js mentioned above, cartodb.js will create and add a new attribution control to the map - even if the property is set to false in Leaflet.  

Maybe I'm wrong, but I think the more correct behaviour is to add attributions to the existing attribution control  - if any, and defer creating/adding of the attribution control to the encompassing application.

I can provide more examples if needed.

alonsogarciapablo:
[...]

PR #959, opened 2015-12-21
Hello new attributions

xavijam:
So new attributions will be shown as a new overlay, not using original Leaflet or GMaps (it will keep showing original GMaps attribution + ours) component. We were using more or less this approach (visually) in the mobile view. I've decided to remove it and only managed in this new overlay/view (cc @javierarce, master of mobile).
Also we will avoid problems with attributions parsing (cc @acanimal).

screen shot 2015-12-21 at 17 08 15

---

screen shot 2015-12-21 at 17 08 24

--- 

screen shot 2015-12-21 at 17 08 38

REVIEWER: @alonsogarciapablo 

cc @piensaenpixel


xavijam:
Changed reviewer, your turn @alonsogarciapablo :dancer: .

xavijam:
Finally tests passing, could you take a look @alonsogarciapablo please? :)

src/geo/ui/attribution/attribution_view.js
```diff
@@ -0,0 +1,75 @@
...
+  },
+
+  _showAttributions: function (e) {
+    this.killEvent(e);
+    this.$('.js-text').addClass('is-visible');
+    this.$('.js-button').removeClass('is-visible');
+    this._initBinds();
+  },
+
+  _hideAttributions: function () {
```
alonsogarciapablo: Should we `this.killEvent(e)` here too?

src/vis/vis.js
```diff
@@ -744,6 +744,11 @@ var Vis = View.extend({
       }
     }
 
+    vizjson.overlays.push({
+      type: 'attribution',
+      order: 5
```
alonsogarciapablo: This mechanism of adding overlays feels a bit hacky (I know other overlays are added this way). Why is the order 5?

src/geo/ui/attribution/attribution_view.js
xavijam: Nope, there is no real action button which closes/hides the attribution. Also, when it is closed by ESC button, we don't want to stop the propagation of the event, just in case.

src/vis/vis.js
xavijam: Good question, I feel it is related to the position in the "map grid", but @javierarce knows more about it. In any case I'm going to remove it because it is always displayed in the same place.

test/spec/geo/ui/attribution.spec.js
```diff
@@ -0,0 +1,119 @@
+var $ = require('jquery');
```
alonsogarciapablo: Instead of relying so much on the internals (eg: calling `this.view._showAttributions();`) to test the behaviour of the attribution-view, I'd try to test it from the user's POV (i.e: simulating clicks and other events, and ensuring the view responds to this events by checking the DOM is updated accordingly). Here's some pseudocode:

```
it('should render the attributions when button is clicked', function () {
  // click .js-button
  // assert that elements are visible
})

describe('when the attributions are expanded', function () {

  beforeEach(function () {
    // click .js-button to expand attributions
  });
  
  it('should hide the button', function() {
   // assert that button is hidden
  });

  it('should collapse the attributions when ESC is pressed', function () {
    // simulate a ESCAPE keystroke
    // assert that elements are NOT visible
  });

  it('should collapse the attributions when user clicks on the document', function () {
    // trigger a click event on the document
    // assert that elements are NOT visible
  });

});

describe('when the attributions are collapsed', function () {
 
  beforeEach(function(){
    // click .js-button to expand attributions
    // simulate a ESCAPE keystroke to collapse the attributions
  });

  it('should show the button', function() {
    // assert that the button is visible
  });

  it('should NOT render the attributions', function () {
    // assert that elements are NOT visible
  });

  it('should not respond to ESC', function () {
    // simulate a ESCAPE keystroke
    // assert that elements are NOT visible
  });

  it('should not respond to clicks on the document', function () {
    // trigger a click event on the document
    // assert that elements are NOT visible
  });
});
```

alonsogarciapablo:
@xavijam remember that we're using `-` instead of `_` for filenames.

My main concern is that some maps (Leaflet in particular) might already have some attributions that users might want/need to keep in the map. Example: I have a map with a basemap from Stamen and I use `createLayer` to add some `CartoDB` layers. I need to ensure that both the `Stament` and attributions coming from CartoDB will be displayed. Check out [this issue](https://github.com/CartoDB/cartodb.js/issues/695) for some more context on this.

I left another comment related to how we approached testing in this PR.

xavijam:
About filenames, true, thanks @alonsogarciapablo.
Related with attribution, remember that with createLayer there is no Map element related with CartoDB, so attributionControl will be displayed (or not, user decides), so he will be able to add any attribution to the map.

alonsogarciapablo:
`cartodb.createLayer` actually [creates a MapView](https://github.com/CartoDB/cartodb.js/blob/2983b2fdcef914afdb1f4fdae173471143930452/src/api/layers.js#L152-L155) and I think that will hide existing attributions from the map (because we're setting/enforcing `.attributionControl` to `false` [here](https://github.com/CartoDB/cartodb.js/pull/959/files#diff-aaf2180ce09425b310a1186ac37bab76R37). But perhaps I'm wrong and it works ok... I'd just double check that.

xavijam:
Let me check it!

alonsogarciapablo:
:+1: Thanks @xavijam!!!

xavijam:
Thanks to you brother!

PR #962, opened 2015-12-21
[WIP] CartoDB.js v4 (with analysis and filtering) API docs

alonsogarciapablo:


docs/analysis-and-filtering-api.md
```diff
@@ -0,0 +1,244 @@
...
+
+## Analysis API
+
+A dataview represents a view of the dataset of a layer. It allows
+users of CartoDB.js to extract information from a layer for analysis
+purposes. For example, the following example defines a dataview to analyse the distribution of the numeric values of a column using the Jenks natural breaks classification method:
+
+```
+var clusterDataview = cartodb.dataviews.create(layer, {
+  type: 'cluster',
```
javisantana: could we call this by name that can be recognized like histogram

docs/analysis-and-filtering-api.md
```diff
@@ -0,0 +1,244 @@
...
+  column: 'price',
+  method: 'jenks',
+  clusters: 10
+});
+
+```
+
+Each type of dataview needs different options, and instances of the different dataview types provide methods to extract the information about the specific kind of analysis (see documentation below).
```
javisantana: One of the things harder to understand in cartodb.js was the options thing. I think is easier to understand and manage to have a type based thing with explicit arguments.

I know that having a create keep everything more open and generic but makes the api harder to use

PR #967, opened 2015-12-22
Added default zoom template + moved view to zoom folder

xavijam:
Basically it is not a requirement to add a template for zoom overlay, we provide the default one. Also zoom_info has been removed because it is already present in the new zoom overlay.

REVIEWER: @viddo 

cc @piensaenpixel


viddo:
:+1:

PR #970, opened 2015-12-23
Fixed problem when infowindow is closed and tooltip filter is unset

xavijam:
tooltip-bug

_Pending test_

REVIEWER: @alonsogarciapablo


xavijam:
Pablo, added TODO, not possible to add a spec right now with current code in v4.

alonsogarciapablo:
:+1:

PR #972, opened 2015-12-23
Tooltip fixes

xavijam:
- Using a wrapper class instead of the proper CDB-Tooltip class, as we do for new infowindow.
- Fixed CSS classes.
- Added the tooltip and other new components to the v4-layergroup example.

CR: @alonsogarciapablo


xavijam:
retest this please

src/geo/ui/tooltip.js
```diff
@@ -148,10 +148,15 @@ var Tooltip = InfoBox.extend({
     var mapViewSize = this.options.mapView.getSize();
     var top = 0;
     var left = 0;
-    var modifierClass = 'CDB-Tooltip--';
+    var modifierClass = 'CDB-Tooltip-wrapper--';
```
alonsogarciapablo: We can get this class name from `this.className`. Here's an idea:

```
setPosition: function () {
  // ...
 this._removePositionModifiers()
  // ...
},

_removePositionModifiers: function ( ) {
  var positions = [ 'topLeft', 'topRight', 'bottomRight', 'bottomLeft' ];
  var positionModifiers = _.map(position, function (className) {
    return this._modifierClassName(className);
  }, this);
  this.$el.removeClass(positionModifiers);
},

_modifierClassName: function (className) {
  return this.className + '--' + className;
})

```

alonsogarciapablo:
retest this please

xavijam:
retest this please

PR #978, opened 2015-12-30
Adds loader and error message to the infowindow

javierarce:
This PR adds:

- Shows error messages to indicate non-valid URLs for the image cover.
- Adds loader.
- Several fixes (like one for the empty infowindow tip or hook that didn't appear if the infowindow had an image cover; or added ```pointer-events: none``` to certain hidden areas that prevented clicking on the map)
- Renamed  classes:``.is-header``` and ```.is-header-image``` to `.has-header``` and ```.has-header-image```

loader_image
loader

screen shot 2016-01-07 at 12 59 54

screen shot 2016-01-07 at 13 19 18

screen shot 2016-01-07 at 13 01 32

### DIFF
--- a/src/geo/ui/infowindow.js
+++ b/src/geo/ui/infowindow.js
@@ -119,8 +119,10 @@ var Infowindow = View.extend({
         this.$('.js-content').css('max-height', this.model.get('maxHeight') + 'px');
       }
 
-      this._loadCover();
       this._renderLoader();
+      this._startLoader();
+
+      this._loadCover();
 
       if (!this.isLoadingData()) {
         this.model.trigger('domready', this, this.$el);
@@ -164,11 +166,18 @@ var Infowindow = View.extend({
   },
 
   _renderLoader: function () {
+    this.$('.js-inner').append('<div class="CDB-Loader js-loader"></div>');
+  },
+
+  _startLoader: function () {
     this.$('.js-inner').addClass('is-loading');
-    this.$('.js-inner').append('<div class="CDB-Loader js-loader is-visible"></div>');
+    this.$('.js-loader').addClass('is-visible');
   },
 
   _stopLoader: function () {
+    if (this._containsCover() && this._coverLoading) {
+      return;
+    }
     this.$('.js-inner').removeClass('is-loading');
     this.$('.js-loader').removeClass('is-visible');
   },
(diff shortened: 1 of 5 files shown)


javierarce:
@xavijam: I'm not sure if you like the approach to work with the loader. Can you have a look, please?

xavijam:
In the image load, bad thing is there is no tick until image is loaded. Any way, @piensaenpixel is your man to take test and check.

javierarce:
ping: @piensaenpixel

javierarce:
@viddo, @piensaenpixel: can you have a look to the PR, please?

piensaenpixel:
@javierarce for me it's ok

javierarce:
ping @viddo.

src/geo/ui/infowindow.js
```diff
@@ -348,23 +377,28 @@ var Infowindow = View.extend({
     var $img = $cover.find('img');
     var url = this._getCoverURL();
 
+    if (!this._isValidURL(url)) {
+      this._showInfowindowImageError();
+      return;
+    } else {
+      this._clearInfowindowImageError();
+    }
```
viddo: I'd suggest to invert the if-else cases, it's usually easier to read
```js
if (isSomething) { 
  /* positive */
} else {
  /* negative */
}```

File: src/geo/ui/infowindow.js [hunk 2]
```diff
@@ -119,12 +118,15 @@ var Infowindow = View.extend({
...
-      // If the template is 'cover-enabled', load the cover
+      this._renderLoader();
+      this._startLoader();
+
       this._loadCover();
 
       if (!this.isLoadingData()) {
         this.model.trigger('domready', this, this.$el);
         this.trigger('domready', this, this.$el);
+        this._stopLoader();
```
viddo: Why is it necessary to render/start loader in the previous lines if it's just going to be stopped a few lines later?

viddo:
:+1: added some comments for some minor doubts

src/geo/ui/infowindow.js
javierarce: Oh, it could happen that the infowindow doesn't contain a cover, and in that case we need to stop the loader. The stop should be in the trigger probably, but I didn't want to change the logic at this moment.

PR #979, opened 2016-01-04
Fix non-hidden attribution

javierarce:
The map attribution was not really hidden. This PR fixes this problem and also refactor the attribution.css a little.

CR: @piensaenpixel


javierarce:
Closing since I'm referencing a different ticket :(

PR #989, opened 2016-01-11
Retrofit windshaft and add new dataviews + filters concepts

viddo:
Migrated code out from deep-insights to here. Retrofit in this case more specifically means:
- Adding the new dataviews concepts
- Update the exsting `src/windshaft/*` code to take into account `dataviews` (previously widgets) and `filters`
- `vis` object is accommodated to take into account the new concepts
- Mostly moved code as it were implemented in deep-insights

In terms of dataviews, instead of exposing the raw models, and taking into account these concepts only makes sense in the context of a vis/map they're only exposed through a factory:
```js
var dataview = vis.dataviewFactory.createHistogramDataview(layer, { … });
```
Internally this wires up the windshaft client and such to take into account this dataview and it's underlying filter. Also, this way we just need to document the public-facing methods, props, and events, the underlying implementation can be changed as we see fit.

@alonsogarciapablo can you review? cc @CartoDB/frontend FYI

### DIFF
--- a/src/dataviews/category-dataview-model.js
+++ b/src/dataviews/category-dataview-model.js
@@ -1,0 +1,339 @@
+var _ = require('underscore');
+var DataviewModelBase = require('./dataview-model-base');
+var SearchModel = require('./category-dataview/search-model');
+var CategoryModelRange = require('./category-dataview/category-model-range');
+var CategoriesCollection = require('./category-dataview/categories-collection');
+var LockedCatsCollection = require('./category-dataview/locked-categories-collection');
+
+/**
+ *  Category dataview model
+ *
+ *  - It has several internal models/collections
+ *
+ *  · search model: it manages category search results.
+ *  · locked collection: it stores locked items.
+ *  · filter model: it knows which items are accepted or rejected.
+ *
+ */
+
+module.exports = DataviewModelBase.extend({
+
+  defaults: _.extend(
+    {
+      allCategoryNames: [] // all (new + previously locked), updated on data fetch (see parse)
+    },
+    DataviewModelBase.prototype.defaults
+  ),
+
+  url: function () {
+    return this.get('url') + '?bbox=' + this.get('boundingBox') + '&own_filter=' + (this.get('locked') ? 1 : 0);
+  },
+
+  initialize: function (attrs, opts) {
+    this._data = new CategoriesCollection();
+
+    DataviewModelBase.prototype.initialize.call(this, attrs, opts);
+
+    // Locked categories collection
+    this.locked = new LockedCatsCollection();
+
+    // Internal model for calculating total amount of values in the category
+    this.rangeModel = new CategoryModelRange();
+
+    // Search model
+    this.search = new SearchModel({}, {
+      locked: this.locked
+    });
+  },
+
+  // Set any needed parameter when they have changed in this model
+  _setInternalModels: function () {
+    var url = this.get('url');
+
+    this.search.set({
+      url: url,
+      boundingBox: this.get('boundingBox')
+    });
+
+    this.rangeModel.setUrl(url);
+  },
+
+  _onChangeBinds: function () {
+    this._setInternalModels();
+
+    this.rangeModel.bind('change:totalCount change:categoriesCount', function () {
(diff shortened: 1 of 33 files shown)


src/dataviews/category-dataview-model.js
```diff
@@ -0,0 +1,339 @@
+var _ = require('underscore');
+var DataviewModelBase = require('./dataview-model-base');
+var SearchModel = require('./category-dataview/search-model');
+var CategoryModelRange = require('./category-dataview/category-model-range');
+var CategoriesCollection = require('./category-dataview/categories-collection');
+var LockedCatsCollection = require('./category-dataview/locked-categories-collection');
```
alonsogarciapablo: IMO, some of the logic in ^^ models shouldn't be part of the dataview. For example, category widgets should use a category-dataview-model to "fetch" the search results, but the dataview itself shouldn't know much about search results. Or the dataview shouldn't know what categories are locked, etc.

src/dataviews/formula-dataview-model.js
```diff
@@ -0,0 +1,33 @@
...
+    {},
+    DataviewModelBase.prototype.defaults,
+    {
+      data: '',
+      suffix: '',
+      prefix: ''
+    }
+  ),
+
+  // TODO: The response format has probably changed
```
alonsogarciapablo: We can probably remove this comment.

src/dataviews/dataview-model-base.js
```diff
@@ -0,0 +1,121 @@
...
+
+    // Retrigger an event when the filter changes
+    if (this.filter) {
+      this.filter.bind('change', this._onFilterChanged, this);
+    }
+  },
+
+  _onChangeBinds: function () {
+    this.bind('change:url', function () {
+      if (this.get('sync') && !this.isDisabled()) {
```
alonsogarciapablo: As mentioned offline, IMO `sync` and `bbox` attributes shouldn't be part of dataview models. Widget models are the ones that determine if the widget is synced to changes on other widgets (`sync`) and changes on the bounding box (`bbox`). I think that dataviews should trigger and event every time the "datasource" changes, and then the widgets should decide what to do based on their configuration (e.g. fetch the data again and refresh themselves if `sync` is enabled).

src/dataviews/category-dataview-model.js
```diff
@@ -0,0 +1,339 @@
...
+      if (!e || (e && e.statusText !== 'abort')) {
+        this.trigger('error', this);
+      }
+    }, this);
+    this.search.bind('change:data', function () {
+      this.trigger('change:searchData', this.search, this);
+    }, this);
+  },
+
+  getLockedSize: function () {
```
alonsogarciapablo: As mentioned [here](https://github.com/CartoDB/cartodb.js/pull/989/files#r49316685), all these methods below should be part of widget models IMO. From what I'm seeing, widget views in the deep-insights repo are using these, which is probably a sign that they are more widget related.

alonsogarciapablo:
:+1:  I left some comments and I will open a separate issues for those (as agreed offline).

I kind of liked `vis.dataviews.createSomething...` instead of `vis.dataviewFactory.createW...` but it's something we can change later too.

src/dataviews/category-dataview-model.js
viddo: > …widget views in the deep-insights repo are using these, which is probably a sign that they are more widget related.

Not necessarily, that widgets _utilize/depends on_ dataviews methods doesn't mean that the logic belongs there. I'll talk w/ @xavijam to decide where these makes sense conceptually.

viddo: I've talked with @xavijam and after his explanations I have a better idea: both the locked and search states are states which is required for the widget implementation, not the dataview itself (which only accept/reject categories really). So :+1: to move the logic.

viddo:
FYI, changed the name to `vis.dataviews`, as you proposed, before merging. :+1:

src/dataviews/category-dataview-model.js
alonsogarciapablo: Great!

PR #992, opened 2016-01-12
Fix image load infowindow

javierarce:
This PR fixes a problem loading custom infowindow templates with images. It also refactor some of the infowindow.js code, makes the loading of the image more pleasant and reviews the infowindow specs.

PS: I've also added several examples in the example folder.

russia


javierarce:
@xavijam: can you take a look, please?

xavijam:
Could we add an example in the v4 branch for testing porpuses? please :)

javierarce:
Sure! :)

javierarce:
I've added a basic first example: `examples/v4-infowindow.html`.

xavijam:
Does the hook appear with white background when image is being loaded?

javierarce:
Oh, yes, that's one of the things I fixed too.

src/geo/ui/infowindow.js
```diff
@@ -327,108 +329,126 @@ var Infowindow = View.extend({
...
+  },
+
   /**
    *  Get cover URL
    */
   _getCoverURL: function () {
     var content = this.model.get('content');
 
-    if (content && content.fields && content.fields.length > 0) {
+    if (this.$('.js-cover img').attr('src')) {
```
xavijam: _Totally optional: I'd create a var storing the src of the img, being used twice._

iriberri:
Awesome. I :revolving_hearts: you.

xavijam:
This is the new infowindow @iriberri :)

javierarce:
I've made several fixes to allow using custom loaders in the content of an infowindow. Sadly, I don't think we can use the same solution for the covers (because images could take longer to load than the fields and the infowindow template would be refreshed, without the special 'loading' flag that we add in the content)

javierarce:
@xavijam: :point_up:  that said, all yours :)

xavijam:
Checking it! :)

src/geo/ui/infowindow.js
```diff
@@ -164,36 +160,28 @@ var Infowindow = View.extend({
     }
   },
 
-  _renderLoader: function () {
+  _renderCoverLoader: function () {
+    var $loader = $('<div class="CDB-Loader js-loader"></div>');
```
xavijam: _Totally optional: You could create the element as $('<div>').addClass('CDB-Loader js-loader')_

javierarce: :cool:

xavijam:
I feel code looks great and it covers perfectly our new features, so :100:.
And also new examples are pure love :D.

javierarce:
Thanks for the review, the advices & the kind words! :)

PR #995, opened 2016-01-13
Removes references to the concept of dashboards from class names

alonsogarciapablo:
I also added a couple of comments to document what `src/windshaft/windshaft-map` and `src/windshaft/windshaft-map-instance` are and simplified the way that `statTag` is stored internally.

@viddo PTAL. Thanks!


viddo:
:shipit:

PR #998, opened 2016-01-13
Added back the "datasource" attribute to the examples for V4

alonsogarciapablo:
I basically removed a bunch of code from vis/vis.load that was there to make createVis work when no "datasource" attribute was present in the viz.json and added that attribute to the viz.json in the examples in the new v4 folder.

cc: @viddo


alonsogarciapablo:
ping @viddo!

src/vis/vis.js
```diff
@@ -403,19 +390,10 @@ var Vis = View.extend({
...
     if (cartoDBLayerGroup) {
       var endpoint;
-      var mapsApiTemplate;
-      var userName;
-      var statTag;
-      var forceCors = true;
       var configGenerator;
 
       var datasource = data.datasource;
       if (datasource) {
```
viddo: With this change isn't datasource always required? If so, perhaps should throw an error to indicate that it's missing if not set?

viddo:
:+1:

src/vis/vis.js
alonsogarciapablo: Yes, good catch!

PR #999, opened 2016-01-13
Auto-adds several infowindow CSS classes

javierarce:
This PR removes the need to add certain CSS classes. Instead, it uses the required .js- classes from the template to add them.


javierarce:
cc: @xavijam, can you take a look, please?

src/geo/ui/infowindow.js
```diff
@@ -178,8 +181,34 @@ var Infowindow = View.extend({
...
+    }
+    if (hasTitle) {
+      $infowindow.addClass('has-title');
+    }
+  },
+
   _renderScroll: function () {
-    if (this.$('.has-scroll').length === 0) return;
+    if (this.$('.js-content').length === 0) {
+      return;
```
xavijam: Shouldn't we remove has-scroll in case it is added?

xavijam:
For me it makes sense, so :+1:

src/geo/ui/infowindow.js
javierarce: Mmm… why would someone add it?

xavijam: Not really common, just wondering :dancer:

javierarce: :ok_hand:  :D

PR #1000, opened 2016-01-14
Including new CartoAssets as module

xavijam:
Please take a look @CartoDB/frontend @rochoa @piensaenpixel, does it make sense?


grunt/tasks/scss.js
```diff
@@ -14,6 +14,8 @@ module.exports = {
         files: [{
           expand: true,
           src: [
+            'node_modules/cartoassets/src/scss/utilities/*.scss',
```
matallo: I wouldn't use wildcards, but explicitely list the files, even better: have an entry file

themes/scss/infowindow/cartodb-infowindow-default.scss
```diff
@@ -1,5 +1,5 @@
-@import '../variables/sizes';
-@import '../variables/mixins';
+@import '../../../node_modules/cartoassets/src/scss/variables/sizes';
```
matallo: I wouldn't add non rendering scss on every file but just on the entry files, they are inherited

xavijam:
@CartoDB/frontend ready to be merged people :), more comments? If not, fasten your seatbelts.

PR #1001, opened 2016-01-14
Fix torque model views setup

### LINKED ISSUE #986: Setup torque layer model in the same way as map
Move/refactor the logic of [torque layer model](https://github.com/CartoDB/cartodb.js/blob/v4/src/geo/map/torque-layer.js#L27-L87), to be bridged outside the view, similar to how it's done in the [mapView](https://github.com/CartoDB/cartodb.js/blob/4a64eccccaff655879b0724594ccf584ca6de8fb/src/geo/map.js#L640-L657), i.e. the implementing view propagates the events according to its context making the model not having to know about views and these details.

viddo:
Resolves #986 

Following the same pattern as used for `MapView` <-> `Map` model, to keep the view and model in sync.

This simplifies the public-facing torque layer model quite a bit internally, avoiding those nasty binding and proxy hacks, should be easier to understand and maintain from here on.

While at it, could DRY the common logic of the two implementing torque layer views quite a bit (extracted to the new common base object).

@alonsogarciapablo can you review? cc @rochoa @javisantana


src/geo/torque-layer-view-base.js
```diff
@@ -0,0 +1,160 @@
...
+  _initialAttrs: function (layerModel) {
+    var extra = layerModel.get('extra_params');
+    return {
+      table: layerModel.get('table_name'),
+      user: layerModel.get('user_name'),
+      column: layerModel.get('property'),
+      blendmode: layerModel.get('torque-blend-mode'),
+      resolution: 1,
+      // TODO: manage time columns
+      countby: 'count(cartodb_id)',
```
javisantana: sql_api_*, tiler_*, animationDuration, countby, steps, blendmode, column and resolution could be removed safely (since we don't need backwards compatibility).

- sql_api_* is not being used
- tiler_* was replaced by maps_api_template
- the other ones go in the cartocss

javisantana:
sounds good, @alonsogarciapablo let me know if you don't understand this change so I explain you it in detail

src/geo/torque-layer-view-base.js
viddo: Oh great to know, will :fire: those then.

src/geo/torque-layer-view-base.js
```diff
@@ -0,0 +1,160 @@
...
+      stat_tag: layerModel.get('stat_tag'),
+      animationDuration: layerModel.get('torque-duration'),
+      steps: layerModel.get('torque-steps'),
+      sql: this._getQuery(layerModel),
+      visible: layerModel.get('visible'),
+      extra_params: {
+        api_key: extra ? extra.map_key : ''
+      },
+      attribution: layerModel.get('attribution'),
+      cartocss: layerModel.get('cartocss') || layerModel.get('tile_style'),
```
viddo: @javisantana considering the mention above, about some props that could be removed due to dropping backward compability, can the `tile_style` fallback be :fire: too?

javisantana: I guess yes, we should use cartocss

src/geo/torque-layer-view-base.js
```diff
@@ -0,0 +1,160 @@
...
+      tiler_domain: layerModel.get('tiler_domain'),
+      tiler_port: layerModel.get('tiler_port'),
+      maps_api_template: layerModel.get('maps_api_template'),
+      stat_tag: layerModel.get('stat_tag'),
+      animationDuration: layerModel.get('torque-duration'),
+      steps: layerModel.get('torque-steps'),
+      sql: this._getQuery(layerModel),
+      visible: layerModel.get('visible'),
+      extra_params: {
+        api_key: extra ? extra.map_key : ''
```
javisantana: map_key as well, should be api_key

alonsogarciapablo:
I left a minor/optional comment. Bye bye double binding! :clap:

viddo:
Reverted the last 3 commits, turns out it was not that easy, removing these causes the deep-insights example to not work (no errors or feedback so hard to debug). I think it's better to deal with cleaning up these in isolation.

PR #1003, opened 2016-01-15
Remove check for layergroup to setup WindshaftMap

viddo:
Turns out it’s not needed anymore, so remove the checks.

@alonsogarciapablo ping


alonsogarciapablo:
:+1:

PR #1004, opened 2016-01-18
And infowindow popups for WMS

ezgis504:
I am trying to mashup a handful of WMS layers into a Leaflet map. I need the map to include popup infowindows, but this functionality is current unavailable. There are a couple of different scripts out there, but none of them have worked for me. [Leaflet.wms] (https://github.com/heigeo/leaflet.wms) seems like a good one, but like I said it doesn't work for me.


Cartofante:

PR #1008, opened 2016-01-22
Flatten map.layers collection

alonsogarciapablo:
After this PR `map.layers` will work as a "flat" collection of layers. Before this, `CartoDB` layers were being grouped on this collection. This PR moves this responsibility to the mapView classes for Leaflet and GoogleMaps. All mapViews know listen to changes on the `map.layers` collection and handles `CartoDB` layers a bit differently (i.e: grouping layers).

As part of this PR I have extracted 3 factories to split things a bit: 1 factory that knows how to create a MapView for a given provider (`map-view-factory.js`), and a couple of factories that know how to create instances of LayerViews for each different type of layer view and provider (`leaflet-layer-view.js` and `gmaps-layer-view.js`). I've also tried to make the code a bit more clear by renaming some methods/internals.

Sorry for the huge PR!

@acanimal can you review this one please?

cc: @viddo @javisantana


src/geo/leaflet/leaflet-layer-view-factory.js
```diff
@@ -0,0 +1,49 @@
+var log = require('cdb.log');
```
alonsogarciapablo: @viddo Here's an example of a factory that I'm introducing here. Some other factories that we have in CartoDB.js provide a way to register new types and constructors but I've decided to go with this very simple alternative (no new types will need to be added at runtime). I'd like to hear you thoughts on this! Thx

src/geo/gmaps/gmaps-layer-view-factory.js
```diff
@@ -0,0 +1,47 @@
...
+var GMapsBaseLayerView = require('./gmaps-base-layer-view');
+var LeafletWMSLayerView = require('../leaflet/leaflet-wms-layer-view');
+
+var GMapsLayerViewFactory = function () {};
+
+GMapsLayerViewFactory.prototype.createLayerView = function (layerModel, mapModel) {
+  var layerType = layerModel.get('type').toLowerCase();
+  var LayerViewClass;
+
+  if (layerType === 'tiled') {
```
javisantana: arg :P

viddo:
For the moved/changed files that have linter issues fixed I think it's beneficial to remove them from the `.eslintignore` file

src/geo/map-view-factory.js
```diff
@@ -0,0 +1,33 @@
...
+      throw new Error('Google maps library should be included');
+    }
+  } else {
+    throw new Error(provider + ' provider is not supported');
+  }
+
+  return new MapViewClass({
+    el: el,
+    map: mapModel,
+    layerViewFactory: new LayerViewFactoryClass()
```
javisantana: don't totally understand why the factory should be passed but other things are done though inheritance. Does make sense to change the factory for a map implementation?

src/geo/leaflet/leaflet-layer-view-factory.js
viddo: The introduction of this factory is arguably better what we had before so :+1: for me under current circumstances. Anyway, since you asked for thoughts on this particularly: 

1. The `LeafletTorqueLayer` is actually one of those cases that IMHO could/should be _registered_ dynamically when that bridge-lib loads, instead of having this factory even have to know anything about it.

2. It's assumed all layers here are created in the same way. That seems fine for the current use-cases, but will not work the second there's a new type that can't adhere to this convention, which is why I prefer the approach of delegating the object creation entirely.

acanimal:
Nice for me.

src/geo/gmaps/gmaps-layer-view-factory.js
alonsogarciapablo: I decided to go with the simplest possible solution to raise a [discussion about factories](https://github.com/CartoDB/cartodb.js/pull/1008#discussion_r50563814). We were/are following different patterns in CartoDB.js (factory with map of types/constructors vs factory that provides a way to register mapping).

viddo: After the short offline talk on Fri and giving it a second thought, I think "the simplest possible solution" would be a simple map (as it was before), since these if-else's are just mapping type-->class anyway.

src/vis/vis.js
```diff
@@ -338,45 +331,28 @@ var Vis = View.extend({
...
+        layers.push(Layers.create(layerData.type, self, layerData));
       }
     });
 
-    // Map layers are resetted and the mapView adds the layers to the map
-    this.map.layers.reset(layers);
+    // TODO: This is PUBLIC. Remove dependency on this attribute from deep-insights.js
+    this.interactiveLayers = new Backbone.Collection(_.select(layers, function (layer) {
+      return layer.get('type') === 'CartoDB' || layer.get('type') === 'torque';
+    }));
```
viddo: I think we need to rethink this `interactiveLayers` object, it will be invalid as soon as a user add/change/remove a layer, but probably better to tackle that in a separate issue/PR.

src/geo/gmaps/gmaps-layer-view-factory.js
alonsogarciapablo: :+1:

src/geo/map-view-factory.js
alonsogarciapablo: I agree that it's kind of weird because a LeafletMapView will always have an instance of a LeafletLayerViewFactory. But, it makes the design a bit more flexible IMO:

- This factory can now be constructed elsewhere. For example, if we wanted modules (e.g: torque) to register new types of constructors on the fly (as they are loaded, as @viddo was suggesting).
- We can pass a fake factory to a generic MapView in out tests and make assertions on the fake factory (instead of having to spy on internals).

PR #1011, opened 2016-01-25
Fixes google maps API dependency

alonsogarciapablo:
Some class definitions expected the global "google" object to be defined. This PR introduces a condition in the `gmaps-layer-view-factory` that prevents those files from being required when "google" is not present.

@viddo perhaps there's a better way to do this...

### DIFF
--- a/grunt/tasks/jasmine.js
+++ b/grunt/tasks/jasmine.js
@@ -34,11 +34,7 @@ module.exports = {
         ],
         options: _.defaults({
           outfile: 'test/SpecRunner-cartodb.html',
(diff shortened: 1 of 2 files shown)


viddo:
How does this resolve the issue? From what I can see the gmaps lib is removed from the tests runtime, so how are this code-path tested? It's not hit anymore?

alonsogarciapablo:
Yes, I removed my previous change from `grunt/tasks/jasmine.js` because that's exactly what happens when you try to load cartodb.js and the Google Maps API is not present (and that particular test bundle is  just testing that files are being loaded correctly).

viddo:
Aha ok, so good to go, right?

PR #1012, opened 2016-01-26
Removed untested methods that shouldn't be needed with new viz.json

alonsogarciapablo:
I don't fully understand what those two methods are doing. They are not tested and examples work fine without them so :fire:. Also, we're re-doing the viz.json so things are going to change anyhow.

cc: @javisantana


javisantana:
Basically you can fire them but I would not remove them without adding the functionality at the right place

PR #1013, opened 2016-01-26
Pass options to Layers Factory instead of passing the whole vis object

alonsogarciapablo:
A little step towards cleaning up vis/vis and removing dependencies with the vis object. I wanted to remove `vis.https` but I then realized is being used by `cartodb.createLayer`. Still, layer factory doesn't depend on vis anymore so I think it's better than it was.

@acanimal what do you think? Thanks!

### DIFF
--- a/src/vis/vis/layers.js
+++ b/src/vis/vis/layers.js
@@ -6,32 +6,32 @@ var Layers = {
 
   _types: {},
 
-  register: function(type, creatorFn) {
+  register: function (type, creatorFn) {
     this._types[type] = creatorFn;
   },
 
-  create: function(type, vis, data) {
+  create: function (type, data, options) {
     if (!type) {
-      log.error("creating a layer without type");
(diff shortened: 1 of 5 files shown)


src/vis/vis/layers.js
```diff
@@ -6,32 +6,32 @@ var Layers = {
...
+    var LayerClass = this._types[type.toLowerCase()];
 
-    var c = {};
-    c.type = type;
-    _.extend(c, data, data.options);
-    return new t(vis, c);
+    var layerAttributes = {};
+    layerAttributes.type = type;
+    _.extend(layerAttributes, data, data.options);
+    return new LayerClass(layerAttributes, options);
```
acanimal: Much more better names :+1:

acanimal:
Fine for me. Step by step API is becoming clearer and with better source code.

PR #1018, opened 2016-01-28
Removed responsibilities from WindshaftMap 

### LINKED ISSUE #1016: Map should decide when to create new instances of the map in Windshaft
- [x] Map has a reference to the WindshaftMap and decides when to create new instances of the map based on events that are triggered by the layers.
- [x] LayerModels implement a method that the map uses to decide if changes on the layer should trigger a new request to Winsshaft. 
- [x] Dataviews hold a reference to the Map and WindshaftMap and fetch new data when map bounds change or a new instance of the map is created in Windshaft.

alonsogarciapablo:
Closes #1016.

Highlights:

- WindshaftMap is no longer the one who decides when new instances of maps must be created and is no longer responsible of "propagating" new URLs to layer view models and dataviews.
- Dataviews are now bound to the Map and WindshaftMap and can now decide if they should update their `urls` and `boundingBox` when a the map bounds change or a new instance of the map has been created. NOTE: This bindings are updating the `url` and `boundingBox` attributes which are bound internally. We could remove that "internal" level of indirection but I didn't want to change too many things at the same time.
- Map exposes a `reload` method that forces the creation of a new instance of the map. Also, the map  listens to changes on the collection of layers so that a new instance of the map is created when a layers are reseted/added/removed.
- Layers hold a reference to the Map and explicitly and can now decide when to invoke `map.reload`.
- Removed vis.interactiveLayers.
- Unified WindshaftMap and WindshaftMapInstance into a single Backbone.Model that exposes a `createInstance` method.

@viddo could you please review this? Thanks!


src/dataviews/dataviews-factory.js
```diff
@@ -69,12 +81,17 @@ module.exports = Model.extend({
...
 
   _indexOf: function (layerModel) {
-    var index = this._interactiveLayersCollection.indexOf(layerModel);
+    // We need to filter the layers to only select those that Windshaft knows about
+    // and be able to calculate the right index.
+    var interactiveLayers = this._layersCollection.select(function (layer) {
+      return layer.get('type') === 'CartoDB' || layer.get('type') === 'torque';
+    });
+
+    var index = interactiveLayers.indexOf(layerModel);
```
viddo: I was reading up on _why_ we would need the layer here, I think we could :fire: both this change and this entire filter-layer reference. 

But, instead of writing several paragraphs here, let's do a quick hangout to discuss it, let's summarize the decision afterwards if it makes sense.

src/geo/map.js
```diff
@@ -21,24 +21,78 @@ var Map = Model.extend({
...
+    this.layers.bind('reset', this._onLayersResetted, this);
+    this.layers.bind('add', this._onLayerAdded, this);
+    this.layers.bind('remove', this._onLayerRemoved, this);
     this.layers.bind('change:attribution', this._updateAttributions, this);
 
-    this.geometries = new Backbone.Collection();
+    if (this._dataviewsCollection) {
+
+      // When new dataviews are defined, a new instance of the map needs to be created
+      this._dataviewsCollection.bind('add', _.debounce(this._onDataviewAdded, 10), this);
```
viddo: As of Backbone v1 we can use the more concise way/recommended way of listening to events:
```js
this.listenTo(otherObj, 'eventName', handler);
```

This way we can skip using the last `this` arg since the context will always be the `otherObj`. See http://backbonejs.org/#Events-listenTo

Btw, since using the debounce, you probably want to bind the context on the handler to be sure it's using the correct context, i.e. `_.debounce(this._onDataviewAdded.bind(this), 10),`

src/geo/map.js
```diff
@@ -21,24 +21,78 @@ var Map = Model.extend({
...
+    this._updateAttributions();
+  },
+
+  _onDataviewAdded: function (layerModel) {
+    this.reload();
   },
 
-  _updateAttributions: function() {
+  reload: function (options) {
+    this._createWindshaftMapInstance(options);
```
viddo: If this is the only usage of this method, why not just inline the content of that function here instead?

viddo:
Some minor comment, but overall the explanation in the desc and code sounds & looks reasonable to me :+1:

src/dataviews/dataviews-factory.js
viddo: After talking: this can indeed be removed, but better to do it in a separate PR. Especially, due to other requirements with vector integration, that will move the ownership of filters to a layer instead. This will also have the nice side-effect of cleaning up filters collection and windshaft-map (serialized layer should include filters that instead of the custom `FiltersCollection.toJSON`).

src/geo/map.js
alonsogarciapablo: :+1:

src/geo/map.js
alonsogarciapablo: Done! Thanks!

PR #1021, opened 2016-01-28
Cast any category value to string

### LINKED ISSUE #1019: null Categories should be casted to 'null'
Right now, if a ```null``` category appears, it will fail trying to "stringify" a null value.

xavijam:
Including nulls and undefines. Fixes #1019.

RW: @viddo


viddo:
Perhaps we should have a test cover this, to avoid it to happen again?
Anyway, :shipit: for the solution

xavijam:
Yep, here we go! ^^^

viddo:
:ship: :it:

PR #1023, opened 2016-01-29
Vector layer cleanup

alonsogarciapablo:
This PR removes all the unnecessary stuf from `src/geo/leaflet/leaflet-cartodb-vector-layer-group-view.js` so that we can focus on the important things :)

cc: @fdansv PTAL (here's the [resulting file](https://github.com/CartoDB/cartodb.js/blob/f470a561be543cb3f96dd2d3925a8c4dc060bdf4/src/geo/leaflet/leaflet-cartodb-vector-layer-group-view.js)).

### DIFF
--- a/src/geo/leaflet/leaflet-cartodb-vector-layer-group-view.js
+++ b/src/geo/leaflet/leaflet-cartodb-vector-layer-group-view.js
@@ -54,49 +54,6 @@ var LeafletCartoDBLayerGroupView = L.CartoDBd3Layer.extend({
     var previousEvent;
     var eventTimeout = -1;
 
-
-    opts.featureOver = function (e, latlon, pxPos, data, layer) {
-      if (!hovers[layer]) {
-        self.trigger('layerenter', e, latlon, pxPos, data, layer);
-      }
-      hovers[layer] = 1;
-      _featureOver && _featureOver.apply(this, arguments);
-      self.featureOver && self.featureOver.apply(self, arguments);
-      // if the event is the same than before just cancel the event
-      // firing because there is a layer on top of it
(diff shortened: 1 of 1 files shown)


src/geo/leaflet/leaflet-cartodb-vector-layer-group-view.js
```diff
@@ -1,360 +1,42 @@
...
-   * Bind off event for wax interaction
-   */
-  _manageOffEvents: function (map, o) {
-    if (this.options.featureOut) {
-      return this.options.featureOut && this.options.featureOut(o.e, o.layer);
-    }
-  },
+  _onTileJSONChanged: function () {
+    var tilejson = this.model.get('urls');
+    this.setUrl(tilejson.tiles[0]);
```
alonsogarciapablo: @fdansv IMO the following chunk could/should be part of the implementation of `setURL`:

```

    this.options.styles = this.model.layers.pluck('cartocss');

    if (this.renderers.length === 0) {
      L.CartoDBd3Layer.prototype.onAdd.call(this, this.leafletMap);
    } else {
      this.setProvider({
        styles: this.model.layers.pluck('cartocss'),
        urlTemplate: tilejson.tiles[0]
      });
    }

    this.options.styles = this.model.layers.pluck('cartocss');

    if (this.renderers.length === 0) {
      L.CartoDBd3Layer.prototype.onAdd.call(this, this.leafletMap);
    } else {
      this.setProvider({
        styles: this.model.layers.pluck('cartocss'),
        urlTemplate: tilejson.tiles[0]
      });
    }
```

What do you think?

PR #1025, opened 2016-01-29
Implement dataview.remove

### LINKED ISSUE #1024: Implement dataview.remove()
Use-case is that we'll need to support removing widgets, which should have the side-effect of removing the underlying dataview (and internally filter, and whatever else might be necessary). See https://github.com/CartoDB/deep-insights.js/issues/90 for more details.

Final scope TBD

cc @alonsogarciapablo

viddo:
Resolves #1024 

Noteworthy:
- Changed to `listenTo` instead, since that allows to remove _all_ event listeners from the own object (even on other objects), upon calling `stopListening`, which comes handy to cleanup after itself.
- Use `destroy` event (in lieu of a `remove event for example) because it removes the model from the parent collection too w/o any additional code.
- Renamed `dataviews.createFoobarDataview` to `dataview.createFoobarModel` to clarify what's returned
- Some cleanup of build stuff (skip console output that's included in the index file)

@alonsogarciapablo can you review? Also, I'm probably missing things… should a removal need updates in other parts of the code base? E.g. Should the WindshaftMap be re-instantiated?

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -140,26 +140,6 @@ module.exports = function(grunt) {
     grunt.config('config.doWatchify', true); // required for browserify to use watch files instead
   });
 
-  grunt.event.once('connect.server.listening', function(host, port) {
-    grunt.log.writeln('Styleguide:');
-    grunt.log.writeln(' - http://' + host + ':' + port + '/themes/styleguide');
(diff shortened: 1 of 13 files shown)


alonsogarciapablo:
:+1:

PR #1026, opened 2016-01-29
Moves provider setting logic to layer

fdansv:
Ref https://github.com/CartoDB/d3.cartodb/pull/82

@alonsogarciapablo


alonsogarciapablo:
:+1:

src/geo/leaflet/leaflet-cartodb-vector-layer-group-view.js
```diff
@@ -22,18 +22,8 @@ var LeafletCartoDBLayerGroupView = L.CartoDBd3Layer.extend({
...
-
-    if (this.renderers.length === 0) {
-      L.CartoDBd3Layer.prototype.onAdd.call(this, this.leafletMap);
-    } else {
-      this.setProvider({
-        styles: this.model.layers.pluck('cartocss'),
-        urlTemplate: tilejson.tiles[0]
-      });
-    }
+    this.setUrl(tilejson.tiles[0]);
```
alonsogarciapablo: IMO `setURL` is something that is particular of the XYZProvider we're using (although we are not setting this up explicitly), so I'd expect this to be `this.provider.setURL(...)`.

PR #1027, opened 2016-01-29
Only add boundingBox if set

viddo:
cc @alonsogarciapablo

### DIFF
--- a/src/dataviews/category-dataview-model.js
+++ b/src/dataviews/category-dataview-model.js
@@ -26,9 +26,8 @@ module.exports = DataviewModelBase.extend({
     if (this.get('boundingBox')) {
       params.push('bbox=' + this.get('boundingBox'));
     }
-    if (this.get('enableFilter')) {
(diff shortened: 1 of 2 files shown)


src/dataviews/category-dataview-model.js
```diff
@@ -22,7 +22,15 @@ module.exports = DataviewModelBase.extend({
   ),
 
   url: function () {
-    return this.get('url') + '?bbox=' + this.get('boundingBox') + '&own_filter=' + (this.get('enableFilter') ? 0 : 1);
+    var params = [];
+    if (this.get('boundingBox')) {
+      params.push('bbox=' + this.get('boundingBox'));
+    }
+    if (this.get('enableFilter')) {
+      params.push('own_filter=' + (this.get('enableFilter') ? 0 : 1));
```
viddo: @alonsogarciapablo should this really be in a if-check? looks wrong compared to the prev code above…

alonsogarciapablo: I don't think so!

PR #1030, opened 2016-02-01
Update compass gem version matching with Editor compass version

xavijam:
CR: @javierarce


xavijam:
Tested grunt tasks and working properly. Tested generated CSS in local examples and working properly :dancer: .

javierarce:
:+1:

PR #1032, opened 2016-02-01
moved createLayer up one level and provided link to layer source options

csobier:
For [Docs issue#596](https://github.com/CartoDB/docs/issues/596#issuecomment-178009348)

Thanks for clarifying @alonsogarciapablo !  Are you okay with these suggested changes?

### DIFF
--- a/doc/api_methods.md
+++ b/doc/api_methods.md
@@ -363,6 +363,8 @@ layer.setParams({'test': 1, 'color': '#F00'}); // set more than one parameter at
 
 ## cartodb.CartoDBLayer.SubLayer
 
+This is a [subclass](https://en.wikipedia.org/wiki/Inheritance_(object-oriented_programming)#Subclasses_and_superclasses) of `cartodb.CartoDB.SubLayer`, which is the subLayer base source code.
+
 ### sublayer.set(_layerDefinition_)
 
 Sets sublayer parameters. Useful when more than one parameter needs to be changed.
@@ -497,6 +499,8 @@ maxHeight | Max height of the scrolled content (value must be a number).
 ---
 
 ## cartodb.CartoDBLayer.HttpSubLayer
+
+This is a [subclass](https://en.wikipedia.org/wiki/Inheritance_(object-oriented_programming)#Subclasses_and_superclasses) of `cartodb.CartoDB.SubLayer`, which is the subLayer base source code.
 
 ### sublayer.setURLTemplate(_urlTemplate_)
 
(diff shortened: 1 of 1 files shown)


csobier:
@alonsogarciapablo , we just got this inquiry from a Partner. Can you let me know so I can address this in the same branch?

"What is the difference between [cartodb.CartoDBLayer.SubLayer](http://docs.cartodb.com/cartodb-platform/cartodb-js/api-methods/#cartodbcartodblayersublayer) and [cartodb.CartoDBLayer.CartoDBSubLayer](http://docs.cartodb.com/cartodb-platform/cartodb-js/api-methods/#cartodbcartodblayercartodbsublayer)? Shouldn't they be on the same section?"

alonsogarciapablo:
:+1: 

`cartodb.CartoDB.SubLayer` is the base class for `cartodb.CartoDBLayer.CartoDBSubLayer` and `cartodb.CartoDBLayer.HttpSubLayer`.

csobier:
Hold on, can you clarify? There is no 'cartodb.CartoDB.SubLayer` function?

alonsogarciapablo:
Basically, `getSublayer` will return an object is an instance of either `cartodb.CartoDBLayer.CartoDBSubLayer` or `cartodb.CartoDBLayer.HttpSubLayer`, which are [subclasses](https://en.wikipedia.org/wiki/Inheritance_(object-oriented_programming)#Subclasses_and_superclasses) of `cartodb.CartoDB.SubLayer`. The goal of the section about `cartodb.CartoDB.SubLayer` is to describe the methods that are common to both classes (inherited methods). I can see how it's confusing, so we could:

1. Make it more clear that `cartodb.CartoDBLayer.CartoDBSubLayer` or `cartodb.CartoDBLayer.HttpSubLayer` are subclasses of `cartodb.CartoDB.SubLayer`.
2. Remove the section about `cartodb.CartoDB.SubLayer` and duplicate the documentation about the common methods both on `cartodb.CartoDBLayer.CartoDBSubLayer` or `cartodb.CartoDBLayer.HttpSubLayer`. This has the drawback that any change will need be changed twice.

alonsogarciapablo:
BTW, `cartodb.CartoDBLayer.CartoDBSubLayer` and `cartodb.CartoDBLayer.HttpSubLayer` are just two names that we're giving to those classes/concepts. They are not the real names in the source code.

csobier:
Okay @alonsogarciapablo , I added a note in the two sections, as discussed. Is this correct?

csobier:
This makes sense! What do you think?

alonsogarciapablo:
:clap: :clap: :clap:

csobier:
:dancer: Thank you! Users will flip over this.

PR #1036, opened 2016-02-03
Returning suffix and prefix values for Formula and Category dataviews models

xavijam:
Trying to fix https://github.com/CartoDB/deep-insights.js/issues/105, @javierarce and I realized that dataviewModel(s) are not returning ```suffix``` and ```prefix``` attributes when ```toJSON``` is called.

CR: @javierarce
cc @viddo @alonsogarciapablo

### DIFF
--- a/test/spec/dataviews/category-dataview-model.spec.js
+++ b/test/spec/dataviews/category-dataview-model.spec.js
@@ -132,6 +132,14 @@ describe('dataviews/category-dataview-model', function () {
     expect(this.model._fetch.calls.count()).toEqual(1);
   });
 
+  describe('toJSON', function () {
+    it('should return suffix and prefix values', function () {
+      var data = this.model.toJSON();
+      expect(data.suffix).toBeDefined();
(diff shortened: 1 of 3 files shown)


javierarce:
:+1:

test/spec/dataviews/category-dataview-model.spec.js
```diff
@@ -132,6 +132,15 @@ describe('dataviews/category-dataview-model', function () {
     expect(this.model._fetch.calls.count()).toEqual(1);
   });
 
+  describe('toJSON', function () {
```
alonsogarciapablo: Why not take this opportunity to test the whole method?

```
this.model.set({ // whatever the model needs });

expect(data.type).toEqual('formula');
expect(data.options).toEqual({
  column: 'column',
  operation: 'operation',
  suffix: 'suffix',
  prefix: 'prefix'
})
```

xavijam:
All yours @javisantana, :kiss:.

PR #1039, opened 2016-02-04
Hooked up the GeoJSONProvider

alonsogarciapablo:
This PR changes the fetch method of dataview models so that it binds the dataview to the layer's data provider (if present). I added some basic tests and left some TODOs, but I think this is is good to start with. Also, not all types of dataviews/operations are supported YET.

@fdansv can you please review? cc: @javisantana


src/dataviews/dataview-model-base.js
```diff
@@ -134,6 +134,39 @@ module.exports = Model.extend({
...
+  },
+
+  _fetchFromDataProvider: function (opts) {
+    var dataProvider = this.layer.getDataProvider();
+
+    // TODO: At the beginning, the dataProvider will not have any features
+    // loaded so we will have to listen to the `featuresChanged` event, but
+    // once the dataProvider has been initialized, we should implement and use
+    // dataProvider.getFeatures() instead of adding yet another binding.
+    dataProvider.bind('featuresChanged', function (features) {
```
fdansv: Aren't we subscribing to this event a bunch of times?

alonsogarciapablo: Yes, that's why I added the TODO. I could go ahead and fix it right now, but I think it's more important to move ahead with the filtering part.

fdansv:
Ha sorry, hadn't noticed the todo. You're right in that it's not that much of a priority at the moment. Apart from that, everything LGTM. :ship: :it:

src/dataviews/dataview-model-base.js
alonsogarciapablo: Good catch anyway!

javisantana: wouldn't be as easy as using once? http://backbonejs.org/#Events-once

alonsogarciapablo: There are a couple of different situations I can think of here when it comes to fetching from a data provider:

1. When the dataProvider doesn't have features yet :point_right: we need to wait for the 'featuresChanged' event.
2. When the dataProvider has some features loaded :point_right: we need to get the features from the layer using layer.getFeatures().

But, at the same time, we want the dataview to always be bound to the `featuresChanged` event so that every time there're new features, the dataview gets updated.

Here's some pseudocode about how I think it should work:

```
var dataProvider = this.layer.getDataProvider();

// Features are already there, so we can set them on the dataview
if (dataProvider.hasFeatures()){
   this.set(this.parse(dataProvider.getFeatures()));
}

// Remove the existing bindings to the featuresChanged event to prevent
// multiple bindings
dataProvider.off('featuresChanged');

// Add a new binding so that the dataview is updated every time
// the data provider has new features.
// Perhaps we shouldn't be doing this here since this is part of the fetch lifecycle?
dataProvider.on('featuresChanged', function (features) {
   this.set(this.parse(dataProvider.getFeatures()));
}, this);
```

PR #1041, opened 2016-02-05
fixed infinite loop with torque and update torque metadata

javisantana:
cc @alonsogarciapablo


alonsogarciapablo:
:es: if that fixes the problem

PR #1042, opened 2016-02-05
Fix multiple bindings between dataviews and dataproviders

alonsogarciapablo:
Fixes the issues reported [here](https://github.com/CartoDB/cartodb.js/pull/1039#discussion_r51879207).

With this change, widgets are being refreshed must faster (and UI is more responsive in general):

filtering

@rambo what do you think?

cc: @fdansv


fdansv:
Niiiiice ;)

PR #1048, opened 2016-02-05
1040 make dataviews reactive

### LINKED ISSUE #1040: Make dataviews reactive to attrs changes
Ditto https://github.com/CartoDB/deep-insights.js/issues/117

When dataview model have its attrs changed they're not reflected on data rendered, seems like.

viddo:
Fixes #1040 

Basically, renamed `_onFilterChanged ` to `_reloadMap` (since that's what it does), and use that as the side-effect on attrs changes.

@alonsogarciapablo review, please 
cc @javisantana

### DIFF
--- a/src/dataviews/dataview-model-base.js
+++ b/src/dataviews/dataview-model-base.js
@@ -60,10 +60,19 @@ module.exports = Model.extend({
 
     // Retrigger an event when the filter changes
     if (this.filter) {
-      this.listenTo(this.filter, 'change', this._onFilterChanged);
+      this.listenTo(this.filter, 'change', this._reloadMap);
     }
 
-    this.bind('change:operation', this._onFilterChanged, this);
+    this.bind('change:operation', this._reloadMap, this);
+  },
+
+  /**
+   * @protected
+   */
+  _reloadMap: function () {
+    this._map.reload({
+      sourceLayerId: this.layer.get('id')
+    });
   },
 
   _onNewWindshaftMapInstance: function (windshaftMapInstance, sourceLayerId) {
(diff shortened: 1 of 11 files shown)


javisantana:
@alonsogarciapablo a question: does it make sense to have a reloadMap in the new architecture?

src/dataviews/dataview-model-base.js
```diff
@@ -60,8 +60,19 @@ module.exports = Model.extend({
 
     // Retrigger an event when the filter changes
     if (this.filter) {
-      this.listenTo(this.filter, 'change', this._onFilterChanged);
+      this.listenTo(this.filter, 'change', this._reloadMap);
     }
+
+    this.bind('change:operation', this._reloadMap, this);
```
alonsogarciapablo: I don't think we should leave this here since not all types of dataviews have an `operation` attribute.

alonsogarciapablo:
@javisantana @viddo here's how `_onFilterChanged` works on the branch that implements filtering in the client:

```
  _onFilterChanged: function (filter) {
    var dataProvider = this.layer.getDataProvider();
    if (dataProvider) {
      dataProvider.applyFilter(this.get('column'), filter);
    } else {
      this._map.reload({
        sourceLayerId: this.layer.get('id')
      });
    }
  },
```

It will make sense to keep `_onFilterChange` IMO, so perhaps we can have the following in order to avoid conflicts:

```
initialize: function () {
  ...
  this.listenTo(this.filter, 'change', this._onFilterChanged);
  this.bind('change:operation', this._reloadMap, this);
  ...
},

_onFilterChanged: function () {
  this._reloadMap();
},

_reloadMap: function () {
  map.reload(...);
},
```

src/dataviews/dataview-model-base.js
viddo: Yes, you're right, I'll fix this one

PR #1050, opened 2016-02-08
Reload map on prefix+suffix change on a formula dataview

viddo:
@alonsogarciapablo missed some attrs in #1048  :disappointed:


alonsogarciapablo:
:+1:

PR #1054, opened 2016-02-08
Clarify dataview models' default attrs names

viddo:
Changed the until-now-private attrs to snake_case, and changed names to make intentions clearer.
I tried to put names that are semantic to the "outside world", since we'll need to expose/pass these attrs (+`enabled`) from the editor.

@alonsogarciapablo  can you review?

### DIFF
--- a/src/dataviews/dataview-model-base.js
+++ b/src/dataviews/dataview-model-base.js
@@ -10,8 +10,8 @@ module.exports = Model.extend({
     url: '',
     data: [],
     columns: [],
-    syncData: true,
-    syncBoundingBox: true,
+    sync_on_source_change: true,
+    sync_on_bbox_change: true,
     enabled: true
   },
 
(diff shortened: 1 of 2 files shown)


src/dataviews/dataview-model-base.js
```diff
@@ -10,8 +10,8 @@ module.exports = Model.extend({
     url: '',
     data: [],
     columns: [],
-    syncData: true,
-    syncBoundingBox: true,
+    sync_on_source_change: true,
```
alonsogarciapablo: `source` is a bit too genetic IMO. What about `sync_on_data_change`?

viddo: Yes, that sounds reasonable, wasn't sure when I did the change but figured you'd comment here as you just did. :kissing_heart:

viddo:
@alonsogarciapablo I'll take the single comment as a :+1: for the rest, so I go ahead and merge so I can update the dependent PRs.

PR #1055, opened 2016-02-08
Several d3+category fixes

fdansv:
* Uses grouping methods from crossfilter
* Fills in max, min and count values that were previously hardcoded
* Adds **Other** trailing category

Requires merge of https://github.com/CartoDB/d3.cartodb/pull/94
@alonsogarciapablo 
cat

### DIFF
--- a/src/dataviews/category-dataview-model.js
+++ b/src/dataviews/category-dataview-model.js
@@ -166,7 +166,7 @@ module.exports = DataviewModelBase.extend({
   },
 
   getCount: function () {
-    return this.get('categoriesCount');
+    return this.get('allCategoryNames').length;
   },
 
   isOtherAvailable: function () {
--- a/src/geo/data-providers/geojson/geojson-data-provider.js
+++ b/src/geo/data-providers/geojson/geojson-data-provider.js
@@ -17,16 +17,19 @@ GeoJSONDataProvider.prototype._dataGeneratorsForDataviews = {
   category: function (features, options) {
     var columnName = options.column;
     var numberOfCategories = 5;
-
-    // TODO: There's probably a more efficient way of doing this
-    var groups = _.groupBy(features, function (feature) { return feature.properties[columnName]; });
-    var groupCounts = _.map(Object.keys(groups), function (key) { return [key, groups[key].length]; });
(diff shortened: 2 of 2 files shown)


src/geo/data-providers/geojson/geojson-data-provider.js
```diff
@@ -17,31 +17,34 @@ GeoJSONDataProvider.prototype._dataGeneratorsForDataviews = {
   category: function (features, options) {
     var columnName = options.column;
     var numberOfCategories = 5;
-
-    // TODO: There's probably a more efficient way of doing this
-    var groups = _.groupBy(features, function (feature) { return feature.properties[columnName]; });
-    var groupCounts = _.map(Object.keys(groups), function (key) { return [key, groups[key].length]; });
-    var sortedGroups = _.sortBy(groupCounts, function (group) { return group[1]; }).reverse();
+    var sortedGroups = this.filter.getColumnValues(columnName);
```
alonsogarciapablo: IMO, we shouldn't change the context of `this` to get access to the internal `filter` attribute:

1. I'd expect `this.filter` to be an instance variable of the GeoJSONDataProvider in this context cause that is the convention we use in the rest of the app (I don't think we bind any function to something other than `this` anywhere else in the source code).
2. With [this binding](https://github.com/CartoDB/cartodb.js/pull/1055/files#diff-2d93068700d903fc11533b3222cfd924R81) we're basically changing the context of this function so that we can access `this._vectorLayerView.renderers[this._layerIndex].filter.getColumnValues`. This is knowing too much about how the d3 layer operates internally. Couldn't we get this data from the d3 layer using the public API?

I'm wondering also if we could change the implementation of the dataview/category widget so that it works when the features changed and it only returns the categories that are being rendered.

src/dataviews/category-dataview-model.js
```diff
@@ -166,7 +166,7 @@ module.exports = DataviewModelBase.extend({
   },
 
   getCount: function () {
-    return this.get('categoriesCount');
+    return this.get('allCategoryNames').length;
```
alonsogarciapablo: Are we sure this is going to work with regular raster layers?

PR #1057, opened 2016-02-08
Fixes wrong attribute name ownFilter -> enableFilter

alonsogarciapablo:
[This commit](https://github.com/CartoDB/cartodb.js/commit/ff3960e3ec6f93500f454da4b6cd6f3b1b8f1cfc) renamed `ownFilter` to `enableFilter`, but I think we forgot to replace one usage of the attribute. This PR fixes that and adds a couple of tests to test that behaviour.

@viddo can you please review? thx!


viddo:
:+1:

PR #1061, opened 2016-02-09
Fixes category (vector)

fdansv:
Reopening because the previous base branch was merged.

@alonsogarciapablo this should address the concerns regarding the layer bind and direct access to the filter's internal components. Once again, for it to work we need to merge https://github.com/CartoDB/d3.cartodb/pull/94

### DIFF
--- a/src/geo/data-providers/geojson/geojson-data-provider.js
+++ b/src/geo/data-providers/geojson/geojson-data-provider.js
@@ -17,16 +17,19 @@ GeoJSONDataProvider.prototype._dataGeneratorsForDataviews = {
   category: function (features, options) {
     var columnName = options.column;
     var numberOfCategories = 5;
-
-    // TODO: There's probably a more efficient way of doing this
-    var groups = _.groupBy(features, function (feature) { return feature.properties[columnName]; });
-    var groupCounts = _.map(Object.keys(groups), function (key) { return [key, groups[key].length]; });
-    var sortedGroups = _.sortBy(groupCounts, function (group) { return group[1]; }).reverse();
+    var sortedGroups = this.filter.getColumnValues(columnName);
+    var lastCat = {
+      category: 'Other',
+      value: sortedGroups.slice(6).reduce(function (p, c) {
+        return p+c.value
+      }, 0),
+      agg: true
+    }
 
     // TODO: Calculate harcoded values
     var data = {
       categories: [],
-      categoriesCount: 3,
+      categoriesCount: sortedGroups.length,
       count: 7446,
       max: 4580,
       min: 106,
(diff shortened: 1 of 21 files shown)


src/geo/data-providers/geojson/geojson-data-provider.js
```diff
@@ -75,7 +79,7 @@ GeoJSONDataProvider.prototype._dataGeneratorsForDataviews = {
 };
 
 GeoJSONDataProvider.prototype.generateDataForDataview = function (dataview, features) {
-  var generateData = this._dataGeneratorsForDataviews[dataview.get('type')];
+  var generateData = this._dataGeneratorsForDataviews[dataview.get('type')].bind(this);
```
alonsogarciapablo: I don't think we need to `bind(this).

fdansv: The context in the generator functions is `Window` otherwise

alonsogarciapablo: You're totally right

src/geo/data-providers/geojson/geojson-data-provider.js
```diff
@@ -15,33 +15,37 @@ var GeoJSONDataProvider = function (vectorLayerView, layerIndex) {
 // TODO: We can extract each "generator" to an individual file so that this file doesn't get too BIG
 GeoJSONDataProvider.prototype._dataGeneratorsForDataviews = {
   category: function (features, options) {
+    var filter = this._vectorLayerView.renderers[this._layerIndex].filter;
```
alonsogarciapablo: As discussed offline, it would be nice if the vector layer provided a method to get the filter for a give layer so that we don't have to know so many internal details. Something like:

```
getFilter: function (layerIndex) {
   return this.renderers[layerIndex].filter;
}
```

src/geo/data-providers/geojson/geojson-data-provider.js
```diff
@@ -15,33 +15,37 @@ var GeoJSONDataProvider = function (vectorLayerView, layerIndex) {
...
     var columnName = options.column;
     var numberOfCategories = 5;
-
-    // TODO: There's probably a more efficient way of doing this
-    var groups = _.groupBy(features, function (feature) { return feature.properties[columnName]; });
-    var groupCounts = _.map(Object.keys(groups), function (key) { return [key, groups[key].length]; });
-    var sortedGroups = _.sortBy(groupCounts, function (group) { return group[1]; }).reverse();
+    var sortedGroups = filter.getColumnValues(columnName);
```
alonsogarciapablo: I realized that `getColumnValues` [accepts two parameters](https://github.com/CartoDB/d3.cartodb/pull/94/files#diff-d2ddee121ba21440f473010082e89e72R84), but here we're just passing one. Why?

fdansv: The second parameter is optional, which is a limit on the number of values. For the category model we need the whole grouping, so it defaults to `Infinity`, but we'll probably need limits in other calculations.

src/geo/data-providers/geojson/geojson-data-provider.js
fdansv: Done in https://github.com/CartoDB/d3.cartodb/commit/e707b3d22a732a2952dfa798dd14531ba8b19966

PR #1062, opened 2016-02-09
Removed duplicated binding

alonsogarciapablo:
The exact same code that is being removed on this PR is present in [the "super" method](https://github.com/CartoDB/cartodb.js/blob/remove-duplicated-binding/src/dataviews/dataview-model-base.js#L121-L133) so this object was being bound to the "change:enabled" event twice. This was causing the dataview to fetch it's data twice when the dataview was re-enabled.

@viddo it's pretty trivial, but just in case I'm missing something... PTAL. thanks!


viddo:
:shipit:

PR #1064, opened 2016-02-09
Histogram widget with vector tiles

fdansv:


src/geo/data-providers/geojson/geojson-data-provider.js
```diff
@@ -11,7 +11,28 @@ var GeoJSONDataProvider = function (vectorLayerView, layerIndex) {
 };
 
 // TODO: We can extract each "generator" to an individual file so that this file doesn't get too BIG
-GeoJSONDataProvider.prototype._dataGeneratorsForDataviews = {
+GeoJSONDataProvider.prototype._dataGeneratorsForDataviews = {,
+  histogram: function (features, options) {
+    var filter = this._vectorLayerView.renderers[this._layerIndex].filter;
```
javisantana: we should create this in the api instead of accessing it to the internal stuff

fdansv: It's already on course https://github.com/CartoDB/d3.cartodb/commit/e707b3d22a732a2952dfa798dd14531ba8b19966

javisantana: that does not fix the problem you are still calling filter methods... layer should provide what you need

PR #1065, opened 2016-02-09
Ensure `submitBBox` is set to true after the first fetch in the histogram dataview

### LINKED ISSUE #1063: Histogram dataview is not submitting the bbox parameter
It looks like [this line](https://github.com/CartoDB/cartodb.js/blob/v4/src/dataviews/histogram-dataview-model.js#L40-Lundefined) is not working properly and `submitBBox` is not being set to true after the first fetch.

I'll open a PR soon.

alonsogarciapablo:
Fixes #1063.

I think the problem was that the following line was not doing anything:

`this.set.bind(this, 'submitBBox', true))`

@javierarce since I think you wrote the original code... makes sense? Thanks!


javierarce:
:+1: looks good to me

PR #1066, opened 2016-02-09
Fixing change about how to parse data within category dataview

xavijam:
Basically, the acceptedCategories data should only be added when the `enableFilter` is false (before the variable was `ownFilter`).

CR: @alonsogarciapablo


alonsogarciapablo:
:+1:

PR #1067, opened 2016-02-09
Fix problem with category filter when accepts a category where it was rejected

xavijam:
If a category was rejected, when it is accepted, we shouldn't include it in the acceptedCategories "array", it should be removed from the rejectedCategories, and that's it :dancer:.

CR: @alonsogarciapablo


alonsogarciapablo:
:+1:

PR #1074, opened 2016-02-10
Fixed `enableFilter` flag on the category dataview model

alonsogarciapablo:
This "flag" was being used incorrectly so I fixed it. Since `enableFilter` and `disableFilter` are being used from deep-insights.js, I will open a PR on that repo to fix it there too.

@xavijam what do you think?

### DIFF
--- a/test/spec/dataviews/category-dataview-model.spec.js
+++ b/test/spec/dataviews/category-dataview-model.spec.js
@@ -146,7 +146,7 @@ describe('dataviews/category-dataview-model', function () {
     expect(this.model._fetch.calls.count()).toEqual(1);
   });
 
-  fdescribe('.parse', function () {
+  describe('.parse', function () {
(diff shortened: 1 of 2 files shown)


test/spec/dataviews/category-dataview-model.spec.js
```diff
@@ -146,7 +146,7 @@ describe('dataviews/category-dataview-model', function () {
     expect(this.model._fetch.calls.count()).toEqual(1);
   });
 
-  describe('.parse', function () {
+  fdescribe('.parse', function () {
```
xavijam: fdescribe warning :)

xavijam:
Fine by me :)

test/spec/dataviews/category-dataview-model.spec.js
alonsogarciapablo: Good catch! Fixed!

PR #1076, opened 2016-02-10
Dataviews fetch from data provider when present

### LINKED ISSUE #1068: Dataviews should fetch from the dataProvider when present
Right now dataviews provide some methods that cause the data to be fetched from Windshaft. For example, category widgets use the `forceFetch` method on the category dataview model. This is making the Lock feature not work properly when layers have a GeoJSONDataProvider (because it's fetching data from Windshaft):

locking

cc: @javisantana

alonsogarciapablo:
Closes #1068.

@fdansv can you please take a look?

cc: @javisantana


javisantana:
:es:

src/geo/data-providers/geojson/geojson-data-provider.js
```diff
@@ -15,23 +15,34 @@ var GeoJSONDataProvider = function (vectorLayerView, layerIndex) {
 // TODO: We can extract each "generator" to an individual file so that this file doesn't get too BIG
 GeoJSONDataProvider.prototype._dataGeneratorsForDataviews = {
   category: function (features, options) {
```
alonsogarciapablo: There're a couple of things I don't like about this function:

1. It acts differently depending on the `filterEnabled` option: If this option is true, it uses the `features` to generate the data that the dataview needs; but the option is false, it uses the filter to get all the data (no filters applied) and generate the data for the dataview (two very different responsibilities).
2. We're directly passing the dataview attributes through, which are subject to change....
3. This file is getting too big and we're not testing these methods, which are starting to get complicated.

Will try to solve these issues in a separate PR.

cc: @fdansv @javisantana

javisantana: from my point of view `features` is something totally internal that should not be passed. The dataprovider for the layer already knows where and how to get the data. The filterEnabled thing is something dataprovider needs in the same way windshaft needs to know it

agree on tests

fdansv:
All looks good to me, agree that we need to talk about improving the generators.

PR #1078, opened 2016-02-11
Changes over category dataview model

xavijam:
- Checking if search results are already selected or not.
- Adding methods to calculate filter categories over current data.

Both changes from this [Deep-insights PR](https://github.com/CartoDB/deep-insights.js/pull/151/files#diff-9174d99f35e6d799c1c69c256b44b3c3R126).

CR: @alonsogarciapablo

### DIFF
--- a/test/spec/dataviews/category-dataview-model.spec.js
+++ b/test/spec/dataviews/category-dataview-model.spec.js
@@ -163,7 +163,6 @@ describe('dataviews/category-dataview-model', function () {
     expect(this.model._fetch.calls.count()).toEqual(1);
   });
 
-
   describe('filters over data', function () {
     beforeEach(function () {
       this.model._data.reset([{ name: 'one' }, { name: 'buddy' }, { name: 'neno' }]);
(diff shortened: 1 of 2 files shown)


test/spec/dataviews/category-dataview-model.spec.js
```diff
@@ -146,6 +146,34 @@ describe('dataviews/category-dataview-model', function () {
     expect(this.model._fetch.calls.count()).toEqual(1);
   });
 
+  describe('filters over data', function () {
```
alonsogarciapablo: Since we're unit testing two different methods, I would change this to:

```
describe('numberOfAcceptedCategories, function () {
  ...
});

describe('numberOfRejectedCategories, function () {
  ...
});
```

xavijam: Ok!, but let me do that inside a describe that says the purpose of those functions.

xavijam:
Moved last thing from https://github.com/CartoDB/deep-insights.js/pull/151/files#diff-9174d99f35e6d799c1c69c256b44b3c3R126, ready to go @alonsogarciapablo?

xavijam:
retest this please

test/spec/dataviews/category-dataview-model.spec.js
alonsogarciapablo: :+1:  Thanks!

alonsogarciapablo:
:+1:

PR #1082, opened 2016-02-11
Fixes bounding box issue

### LINKED ISSUE #1070: Category dataview is not submitting the bounding box the first time
That's why the widget category is showing categories for features that are not being displayed in the map:

screen shot 2016-02-09 at 17 38 06

alonsogarciapablo:
Fixes #1070, by reverting this commit.

@viddo PTAL! Thanks!

NOTE to @viddo : this might have some effects on the deep-insights.js test suite since I remember you mentioning something related (there's some history about it on slack)

### DIFF
--- a/src/dataviews/dataview-model-base.js
+++ b/src/dataviews/dataview-model-base.js
@@ -48,6 +48,7 @@ module.exports = Model.extend({
     }
 
     this._initBinds();
+    this._updateBoundingBox();
   },
 
   _initBinds: function () {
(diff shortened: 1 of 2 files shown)


viddo:
https://github.com/CartoDB/cartodb.js/commit/3bf1a8b85d43bee75d179e644f3562bb24c1ea32#diff-805364e485d740e26e33839755aab753 found the original commit, should have documented the rationale in the commit… lesson learned for next time!

Anyway, I'm looking through the offline logs to figure out _why_ we changed this in the first place, let me see if I can find a solution that works with this PR for DI too before merging.

viddo:
This change is fine, it's the usage of cartodb.js in DI that arguably could be wrong. Added the optional `bounds` attr to make sure `vis.map.getViewBounds()` returns a non-null value (implicitly called through `this._updateBoundingBox` in the changes of this PR.

viddo:
retest this please

viddo:
I've updated this PR since it was not up-to-date with v4, now it passes

PR #1095, opened 2016-02-15
Fixed dataview naming

### LINKED ISSUE #1094: Invalid naming of dataviews
Dataviews are named incorrectly. This can be seen when looking the URLs they are requesting:
screen shot 2016-02-15 at 15 01 41

acanimal:
CR @viddo cc @alonsogarciapablo 

close #1094 

Added `type` attribute to the allowed dataview properties, otherwise the `id` is now well defined and always is named `undefined-xxx`.

### DIFF
--- a/src/dataviews/dataview-model-base.js
+++ b/src/dataviews/dataview-model-base.js
@@ -186,7 +186,6 @@ module.exports = Model.extend({
   {
     ATTRS_NAMES: [
       'type',
-      'column',
       'sync_on_data_change',
       'sync_on_bbox_change',
       'enabled'
--- a/src/dataviews/dataview-model-base.js
+++ b/src/dataviews/dataview-model-base.js
@@ -34,7 +34,7 @@ module.exports = Model.extend({
     }
 
     if (!attrs.id) {
-      this.set('id', attrs.type + '-' + this.cid);
+      this.set('id', this.defaults.type + '-' + this.cid);
     }
 
     this.layer = opts.layer;
--- a/src/dataviews/dataview-model-base.js
+++ b/src/dataviews/dataview-model-base.js
@@ -185,7 +185,7 @@ module.exports = Model.extend({
   // Class props
   {
     ATTRS_NAMES: [
-      'type',
+      'id',
       'sync_on_data_change',
       'sync_on_bbox_change',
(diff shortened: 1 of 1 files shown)


src/dataviews/dataview-model-base.js
```diff
@@ -185,6 +185,7 @@ module.exports = Model.extend({
   // Class props
   {
     ATTRS_NAMES: [
+      'type',
```
viddo: The type should be a static attr and set by a concrete implementation (e.g. see the [category dataview](https://github.com/CartoDB/cartodb.js/blob/3b6dcfd9e68e952142b387d03d3bc61ba4e2b13c/src/dataviews/category-dataview-model.js#L18)), it's not supposed to be changeable (which it would be by adding the attr name here).

Using the type+cid was a temporary hack, now that widgets are created (in editor) when persisted we should probably pass through the id (i.e. add it here instead of type), so server logs actually can be traceable back to the widget definition sources. cc @rochoa 

If the ids are not important for Windshaft, or if widgets are created ad-hoc (sales?), the proper place would be to change the [initialize](https://github.com/CartoDB/cartodb.js/blob/3b6dcfd9e68e952142b387d03d3bc61ba4e2b13c/src/dataviews/dataview-model-base.js#L37), to get the type from `this.defaults.type`.

rochoa: IDs are not important in windshaft, if they can be used in JSON keys it will do OK. But it probably makes a lot of sense to, as you are proposing, to use the IDs from the editor so it's easier to trace up to the origin.

viddo:
Hey, why closing this PR instead of fixing it?

acanimal:
I though it was not necessary due the comments.

src/dataviews/dataview-model-base.js
acanimal: @viddo I place on dataviewModelBase because is a common attrbitute, like `enabled`.

viddo:
@rochoa FYI, discussed & solved offline w/ @acanimal (see latest changes), as mentioned in the previous comments. The temporary widget id (`type`+`cid`) should be sufficient for now, since the URL also contains the mapstat id.

PR #1097, opened 2016-02-15
fixes updating bins

javisantana:
I'm pretty sure I'm missing something but changing `bins` should not reinstanciate the map since bins is something you set when you fetch.

cc @alonsogarciapablo @viddo

### DIFF
--- a/src/dataviews/histogram-dataview-model.js
+++ b/src/dataviews/histogram-dataview-model.js
@@ -50,7 +50,8 @@ module.exports = DataviewModelBase.extend({
       this.set('submitBBox', true);
     }, this);
     this.listenTo(this.layer, 'change:meta', this._onChangeLayerMeta);
-    this.on('change:column change:bins', this._reloadMap, this);
+    this.on('change:column', this._reloadMap, this);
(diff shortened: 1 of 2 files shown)


acanimal:
Right now bins are ignored and setted with the fetch result, but this is going to change in while (see https://github.com/CartoDB/deep-insights.js/issues/164 and https://github.com/CartoDB/deep-insights.js/issues/71)

acanimal:
In addition I think the same excercise must be done with other dataviews. Has sense to update the map if user changes the formula operation or simple refresh? Has sense to update the map if user updates the category aggregation property?

javisantana:
@acanimal bins are not ignored in general, only the first time

javisantana:
ok, merging

alonsogarciapablo:
I think we actually need to re-instantiate the map when the `bins` of a dataview. @viddo introduced that binding [recently](https://github.com/CartoDB/cartodb.js/commit/99daba6947803ee7c48ee43402b36da8d3a01422) so that widgets gets refreshed when the editor changes the `bins` attribute of a dataview.

PR #1102, opened 2016-02-16
Enable/disable dataview depending on the layer visibility

### LINKED ISSUE #1100: Dataviews continues requesting data on hidden layers
When layer visibility is set to false the widgets are toggled off but dataviews continues requesting data.

screen shot 2016-02-16 at 10 02 49

acanimal:
CR @javisantana cc @alonsogarciapablo @viddo 

close #1100 

Dataviews listen for layers visibility changes and enables/disables accordingly. Otherwise we can have hidden layers and dataviews continue requestin data on server.

This is also required for https://github.com/CartoDB/deep-insights.js/issues/162


javisantana:
:es:

it lacks of the fetch part but that could be done in a different pr

PR #1104, opened 2016-02-16
Map instantiation bug

xavijam:
There used to be several map instantiations when widget dataviews are added to the dataviewsCollection. It generated problems in the map rendering, more requests to the API than expected, etc.

So, from now on, when a dashboard (deep-insights) is created, it will skip the map instantiation until dataviews collection is generated (@viddo idea's, and totally makes sense).

CR: @viddo
cc @javisantana @alonsogarciapablo @javierarce @acanimal 

Deep-insights [PR](https://github.com/CartoDB/deep-insights.js/pull/196).

### DIFF
--- a/src/geo/map.js
+++ b/src/geo/map.js
@@ -28,8 +28,6 @@ var Map = Model.extend({
 
     this._windshaftMap = options.windshaftMap;
     this._dataviewsCollection = options.dataviewsCollection;
-
-    this._initBinds();
   },
 
   _initBinds: function () {
@@ -47,10 +45,14 @@ var Map = Model.extend({
     this.layers.bind('change:attribution', this._updateAttributions, this);
 
     if (this._dataviewsCollection) {
-
       // When new dataviews are defined, a new instance of the map needs to be created
       this.listenTo(this._dataviewsCollection, 'add', _.debounce(this._onDataviewAdded.bind(this), 10));
     }
+  },
+
+  instantiateMap: function () {
+    this._initBinds();
+    this.reload();
   },
 
   _onLayersResetted: function () {
--- a/src/vis/vis.js
+++ b/src/vis/vis.js
@@ -262,7 +262,6 @@ var Vis = View.extend({
     this._applyOptionsToVizJSON(data, options);
 
     this._dataviewsCollection = new DataviewCollection();
-    this._dataviewsCollection.on('add reset remove', _.debounce(this._invalidateSizeOnDataviewsChanges, 10), this);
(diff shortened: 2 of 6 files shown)


src/geo/map.js
```diff
@@ -47,12 +45,16 @@ var Map = Model.extend({
...
 
     if (this._dataviewsCollection) {
-
       // When new dataviews are defined, a new instance of the map needs to be created
       this.listenTo(this._dataviewsCollection, 'add', _.debounce(this._onDataviewAdded.bind(this), 10));
     }
   },
 
+  instantiateMap: function () {
+    this._initBinds();
```
xavijam: Don't [bind map layers changes](https://github.com/CartoDB/cartodb.js/blob/v4/src/geo/map.js#L35) until map is going to be instantiated.

src/vis/vis.js
```diff
@@ -424,14 +423,26 @@ var Vis = View.extend({
...
+    }
+
     return this;
   },
 
+  instantiateMap: function (invalidateMapSize) {
+    this._dataviewsCollection.on('add reset remove', _.debounce(this._invalidateSizeOnDataviewsChanges, 10), this);
+    this.map.instantiateMap();
+
+    if (invalidateMapSize) {
```
xavijam: Only invalidate map size when it is requested. In CartoDB.js visualizations is not needed, because the map size doesn't change. For DI is a different story.

src/vis/vis.js
```diff
@@ -424,14 +423,26 @@ var Vis = View.extend({
...
     });
 
+    if (!options.skipMapInstantiation) {
+      this.instantiateMap();
+    }
+
     return this;
   },
 
+  instantiateMap: function (invalidateMapSize) {
```
viddo: This argument should not really be necessary, see #1105 which contains the proposed change

src/geo/map.js
viddo: The _initBinds seems to a bunch of binds, not only for dataviews but also for layers, are we sure we want to postpone _all_ of them?

xavijam: Yes, if not, when the layers collection is generated and single layers are being added will fire as many `reload` methods as layers has the final map. That means `X` map instantiations.

xavijam: Those methods are needed when the map is already instantiated and ready, and there is any change.

viddo: Alright :+1:

viddo:
LGTM

PR #1105, opened 2016-02-16
Improve map instantiation mechanism

viddo:
Code-review proposal for #1104 

Make the public method to not require any argument, and explain what it’s intended for.

@xavijam I figured doing a PR was easiest way to showcase this, instead of a comment. :)

### DIFF
--- a/src/vis/vis.js
+++ b/src/vis/vis.js
@@ -443,7 +443,7 @@ var Vis = View.extend({
     this.mapView.invalidateSize();
   },
 
-  _instantiateMap: function (invalidateMapSize) {
+  _instantiateMap: function () {
     this._dataviewsCollection.on('add reset remove', _.debounce(this._invalidateSizeOnDataviewsChanges, 10), this);
     this.map.instantiateMap();
   },
--- a/src/vis/vis.js
+++ b/src/vis/vis.js
@@ -428,19 +428,24 @@ var Vis = View.extend({
     });
 
     if (!options.skipMapInstantiation) {
-      this.instantiateMap();
+      this._instantiateMap();
     }
 
     return this;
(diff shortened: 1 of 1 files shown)


src/vis/vis.js
```diff
@@ -428,19 +428,24 @@ var Vis = View.extend({
...
+   * Only expected to be called if {skipMapInstantiation} flag is set to true when vis is created.
+   */
+  instantiateMap: function () {
+    this._instantiateMap();
+    this.mapView.invalidateSize();
+  },
+
+  _instantiateMap: function (invalidateMapSize) {
     this._dataviewsCollection.on('add reset remove', _.debounce(this._invalidateSizeOnDataviewsChanges, 10), this);
     this.map.instantiateMap();
```
xavijam: I feel we have missed the condition

src/vis/vis.js
```diff
@@ -428,19 +428,24 @@ var Vis = View.extend({
...
-  instantiateMap: function (invalidateMapSize) {
+  /**
+   * Force a map instantiation.
+   * Only expected to be called if {skipMapInstantiation} flag is set to true when vis is created.
+   */
+  instantiateMap: function () {
+    this._instantiateMap();
+    this.mapView.invalidateSize();
+  },
+
```
xavijam: I like the idea but for me it is a bit confusing to have two methods with almost the same name.

xavijam:
I'm not against or in favour, but I don't see the problem of adding a public method with an argument.

xavijam:
If we add another library, using this public method, but not needing to invalidate the map size, we will need to create a condition with an argument, but it is ok, let's follow this :+1:.

PR #1106, opened 2016-02-17
Fixing problem with map centering for Deep insights

xavijam:
@javisantana @AbelVM 

First part of https://github.com/CartoDB/deep-insights.js/issues/190.
This really closes the problem -> https://github.com/CartoDB/deep-insights.js/pull/199


src/vis/vis.js
```diff
@@ -317,15 +317,9 @@ var Vis = View.extend({
       provider: data.map_provider
     };
 
-    // if the boundaries are defined, we add them to the map
-    if (data.bounding_box_sw && data.bounding_box_ne) {
```
xavijam: There is no more uses of those variables in the current code and viz.json doesn't provide them, only bounds (and center + zoom) related to map centering.

javisantana:
I think we should not recenter the map on new layers added, this problem should be address in DI.js when the map size changes because of the panel, something around here https://github.com/CartoDB/deep-insights.js/blob/master/src/dashboard-view.js#L45

```js
sidebarVIew.on('resize', function() {
   mapView.invalidateSize()
})
```

xavijam:
What about in [create-dashboard.js](https://github.com/CartoDB/deep-insights.js/blob/master/src/create-dashboard.js#L78) try:

```js
...
var widgets = new WidgetsCollection();
if (widgets.size() > 0) {
  vis.resizeMap();
}
vis.instantiateMap();
```

And then in CartoDB.js a function with:

```js
...
resizeMap: function () {
     var c = this.mapConfig;
     if (c.view_bounds_sw && c.view_bounds_ne) {
       this.map.setBounds([
         c.view_bounds_sw,
         c.view_bounds_ne
       ]);
     } else {
       this.map.setCenter(c.center);
     }
   },
},
...
```

xavijam:
retest this please

src/vis/vis.js
```diff
@@ -435,15 +427,23 @@ var Vis = View.extend({
...
-    this.mapView.invalidateSize();
-  },
-
-  _instantiateMap: function () {
     this._dataviewsCollection.on('add reset remove', _.debounce(this._invalidateSizeOnDataviewsChanges, 10), this);
     this.map.instantiateMap();
   },
 
+  invalidateMapSize: function () {
+    this.mapView.invalidateSize();
```
javisantana: I don't think we should reset the map to the initial state, it should be "reset" to the current center/zoom, right?

xavijam: It has to reset to the origin position. After a talk, we are going to rename the function to something like `centerMapToOrigin`, because not only invalidates, it centers the map to the initial bounds/center.

xavijam:
Approved by the CTO.

javisantana:
:es:

Issue #926, opened 2015-12-11
histogram widget filter is sending wrong bounds


javisantana:
When you filter using the last bucket in a histogram bucket the filter sends wrong bounds, for example:

the buckets goes from 400 to 500, you filter it and the filter sent to the tiler is [400, 499.9999] so rows with 500 are kept out.

how to reproduce:
- go here: http://cartodb.github.io/widgets-examples/airbnb_madrid_oct_2015_listings.html
- filter like in the image: 
  <img width="1231" alt="screen shot 2015-12-11 at 11 45 12 am" src="https://cloud.githubusercontent.com/assets/30147/11749586/aef6c098-9ffc-11e5-94ca-5393eccfb5d9.png">
- notice the values sent in the map instanciation

javierarce:
I'm afraid those values are calculated in the backend:

https://github.com/CartoDB/deep-insights.js/blob/master/src/widgets/histogram/content-view.js#L224

The frontend picks the values from the data model using the lower and upper index.

rochoa:
The problem is the zoomed histogram returns with:
```
{
  "bin_width": 66.8333333333333,
  "bins_count": 6,
  "bins_start": 99,
  ...
}
```

So ... `66.8333333333333 * 6 + 99 = 499.9999999999998`.

More on this:

```
$ node
> // https://gist.github.com/rochoa/5b57e57c9694741c4887
> // Following code was extracted from almost-equal
undefined
> // (c) 2013 Mikola Lysenko. MIT License
undefined
> // MIT License (MIT)
undefined
> // See https://github.com/scijs/almost-equal/blob/93af191fe4075ee2dea94d29a26c796c26bb9cb6/almost_equal.js
undefined
>
undefined
> var PRECISION_FLOAT_EPSILON = 1.19209290e-7;
undefined
> var PRECISION_DOUBLE_EPSILON = 2.2204460492503131e-16;
undefined
>
undefined
> function almostEqual(a, b, absoluteError, relativeError) {
...     absoluteError = absoluteError || PRECISION_DOUBLE_EPSILON;
...     relativeError = relativeError || PRECISION_DOUBLE_EPSILON;
...
...     var d = Math.abs(a - b);
...     if(d <= absoluteError) {
...         return true;
...     }
...     if(d <= relativeError * Math.min(Math.abs(a), Math.abs(b))) {
...         return true;
...     }
...
...     return a === b;
... }
undefined
[...]

PR #1107, opened 2016-02-17
Rounded end for last bin

acanimal:
CR @javisantana 

workaround for #926 
close https://github.com/CartoDB/deep-insights.js/issues/165

This is a workaround to fix a problem with last bucket selection.

### DIFF
--- a/src/dataviews/histogram-dataview-model.js
+++ b/src/dataviews/histogram-dataview-model.js
@@ -82,12 +82,10 @@ module.exports = DataviewModelBase.extend({
       }, buckets[i]);
     }
 
-    // FIXME - Round the end of last bin due https://github.com/CartoDB/cartodb.js/issues/926
+    // FIXME - Update the end of last bin due https://github.com/CartoDB/cartodb.js/issues/926
     var lastBucket = buckets[numberOfBins - 1];
     if (lastBucket.end < lastBucket.max) {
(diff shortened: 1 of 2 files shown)


javisantana:
shouldn't that be done server side?
Anyways, I don't follow the logic, in some cases `lastBucket.end` could be really different than max, for example when `lastBucket.end` is 0.0001 and `lastBucket.max` is 0.000999999999 `lastBucket.end` will be 1

acanimal:
Yes the "good" solution would be to solve at server side.

From the sample I saw, the issue is done when the end is lesser than the max value, just the opposite you pointed. In the abel sample we have the bucket computed `lastBucket.end = 0.000999999999` (which is computed at client side width*numbin) and `lastBucket.max is 0.001`. In that case case I change we made end to be the same as max.

For other cases I round the end using ceil.

javisantana:
My question is, why ceil? shouldn't be something according the `bin_width`? In other words, shoudn't it round to the closest bin_width multiplo?

javisantana:
could you add some tests to check the corner cases?

acanimal:
I put to ensure end is "a bit bigger" and fix the round error but I'm thinking it has no sense. The issue always is done when end less than max, like:

- max: 0.01
- end: 0.009999999

javisantana:
:es:

PR #1109, opened 2016-02-17
Ensure bucket has values

acanimal:
CR @xavijam 

Fix issue when buckets has no data.

### DIFF
--- a/src/dataviews/histogram-dataview-model.js
+++ b/src/dataviews/histogram-dataview-model.js
@@ -84,7 +84,7 @@ module.exports = DataviewModelBase.extend({
 
     // FIXME - Update the end of last bin due https://github.com/CartoDB/cartodb.js/issues/926
     var lastBucket = buckets[numberOfBins - 1];
-    if (lastBucket.end < lastBucket.max) {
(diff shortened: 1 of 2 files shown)


xavijam:
I don't have much context, I encourage you to choose another person with better perspective. I checked @javisantana was on charge of the previous PR.
But in any case, I miss a test checking this problem.

acanimal:
I simply modified to ensure lastBucket has a values. Was mea-culpa. Test always helps but IMO for this case I think it is not much useful. Test are useful to check some logic and here the test will simply *check when there is no bin data `parser` do not crashes`, which is something that should do.

xavijam:
IMHO, if something is broken, we didn't notice and none of our tests warned us, we should try to prevent ourselves to fail again.

acanimal:
Added test "for if the flies" :smile: 
When creating test after coding is hard to make all the tests fine, mainly the negative test.

xavijam:
"For if the flies" is valid to me :dancer: :D :+1:

xavijam:
restarted tests in travis

xavijam:
\o/

PR #1110, opened 2016-02-17
Added done trigger after map instantiation

xavijam:
cc @AbelVM @javisantana


javisantana:
:es:

to take into account for the doc

Issue #1020, opened 2016-01-28
Associate filters directly to map.layers


alonsogarciapablo:
Right now filters "belong" to dataviews, which are related to layers. So we're doing things like:

```
   var filtersFromVisibleLayers = [];
    if (dataviews) {
      filtersFromVisibleLayers = dataviews.chain()
        .filter(function (dataview) { return dataview.layer.isVisible(); })
        .map(function (dataview) { return dataview.filter; })
        .compact() // not all dataviews have filters
        .value();
    }
```

We're also having to [set a `layerIndex` on filters](https://github.com/CartoDB/cartodb.js/blob/v4/src/dataviews/dataviews-factory.js#L56) so they can be grouped together when [we serialize them to send them to Windshaft](https://github.com/CartoDB/cartodb.js/blob/v4/src/windshaft/filters/collection.js#L12-L21).

All this would be easier if filters were associated to layers directly. Plus, it conceptually makes more sense IMO.

cc: @viddo

viddo:
:kissing_heart: :bow:

PR #1113, opened 2016-02-18
Di 164 hide layers

acanimal:
CR @alonsogarciapablo 

close https://github.com/CartoDB/deep-insights.js/issues/169 (from DI).

Please Pable review this PR. It solves a problem when there are two layers with widgets and you hide one of them. The widgets are hidden and the second layer widgets works until you make some filter.

Check if my reasoning is correct.
I realize we have generating an array of filters for winshaft map with the filters in the same position as the layer index and holes of no layer of filter is empty.

Supposing we have two layer and we hide one, new info is send to server with only one layer (not two). Later if you filter on a widget of the second layer an array with two positions is sent, the first empty (hole) for filters of the first layer and the second for filter in the layer two), generating an error on server `{"errors":["Cannot read property 'amount-spent-histogram-layer2' of undefined"]}`.

### DIFF
--- a/src/windshaft/filters/collection.js
+++ b/src/windshaft/filters/collection.js
@@ -2,14 +2,36 @@ var _ = require('underscore');
 var Backbone = require('backbone');
 
 module.exports = Backbone.Collection.extend({
+  initialize: function (filters, layers) {
+    this.layers = layers;
+    Backbone.Collection.prototype.initialize.apply(this, filters);
+  },
+
   toJSON: function () {
     var json = {};
     var activeFilters = this.getActiveFilters();
     if (activeFilters.length) {
       json.layers = [];
       _.each(activeFilters, function (filter) {
-        json.layers.push(filter.toJSON());
+        var index = filter.get('layerIndex');
+        if (index >= 0) {
+          if (json.layers[index]) {
+            _.extend(json.layers[index], filter.toJSON());
+          } else {
+            json.layers[index] = filter.toJSON();
+          }
+        } else {
+          throw new Error('layerIndex missing for filter ' + JSON.stringify(filter.toJSON()));
+        }
       });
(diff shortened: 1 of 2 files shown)


alonsogarciapablo:
Existing code would regenerate would group filters by layer in the resulting JSON object. It will return something like:

```
{
  "layers": [
    // filters for layer 0
    {
      "layer_0_histogram_widget_1": {
        "min": 0,
        "max": 1000
      },
      "layer_0_histogram_widget_2": {
        "min": 10,
        "max": 9000
      }
    },
    // filters for layer 1
    {
      "layer_1_histogram_widget_1": {
        "min": 999,
        "max": 10009
      }
    }
  ]
}
```
If I'm understanding the new code correctly, it will generate something like:

```
{
  "layers": [
    // all filters
    {
      "layer_0_histogram_widget_1": {
        "min": 0,
        "max": 1000
      },
      "layer_0_histogram_widget_2": {
        "min": 10,
        "max": 9000
      }
      "layer_1_histogram_widget_1": {
        "min": 999,
        "max": 10009
      }
    }
  ]
}
```
I haven't tested it out, but I'm not sure Windshaft is prepared to work with all filters grouped in one "layer"... Is it actually working?

acanimal:
Understood, then seems the problem is we are sending filters for layers that are hidden, that is info not sent to server, and windshaft fails, becaus have info about one layer and we are sending filter for two layers.

acanimal:
@alonsogarciapablo Review again Pablo. The issue was `this.client.instantiateMap()` receives a `mapConfig` which only the visible layers but a property `filters` array with a position for each layer (in the case of a hidden layer that position was empty but exists, which make the server return an error).

alonsogarciapablo:
I think we should fix this issue by implementing the changes described in #1020.

We can merge this PR for the time being and start working on #1020.

PR #1114, opened 2016-02-18
Changes to the histogram dataview model

### LINKED ISSUE #1069: Windshaft map should not be refreshed when histogram is zoomed in
The widget histogram is [changing some attributes](https://github.com/CartoDB/deep-insights.js/blob/master/src/widgets/histogram/content-view.js#L420) on its internal histogram dataview model after the first fetch, and when the user zooms in and out. The problem is that dataviews are now [listening to changes on some internal attributes (like bins)](https://github.com/CartoDB/cartodb.js/blob/v4/src/dataviews/histogram-dataview-model.js#L51) to refresh the map when the dataview changes. This binding is causing the map to be refreshed when the internal state of the dataview changes (more specifically, when the `bins` property is changed).

cc: @javisantana @fdansv @viddo @javierarce

alonsogarciapablo:
- Adds two public methods: enableFilter and disableFilter, that will be used by deep-insights. This is consistent with the category-dataview-model.
- Histrogram Dataview models now sets the `start`, `end` and `bins` properties after the first fetch.
- `.url` method only includes `start`, `end` and `bins` when `own_filter` is 1.

This is my second attempt to fix #1069 in a nicer way :) :) :). Checkout CartoDB/deep-insights.js#220 to checkout the changes in deep-insights.js.

@javisantana since you suggested following a different approach... what do you think? thanks!

### DIFF
--- a/src/dataviews/histogram-dataview-model.js
+++ b/src/dataviews/histogram-dataview-model.js
@@ -18,17 +18,18 @@ module.exports = DataviewModelBase.extend({
     if (this.get('column_type')) {
       params.push('column_type=' + this.get('column_type'));
     }
-    if (_.isNumber(this.get('start'))) {
-      params.push('start=' + this.get('start'));
-    }
-    if (_.isNumber(this.get('end'))) {
-      params.push('end=' + this.get('end'));
-    }
-    if (_.isNumber(this.get('bins'))) {
-      params.push('bins=' + this.get('bins'));
-    }
     if (_.isNumber(this.get('own_filter'))) {
       params.push('own_filter=' + this.get('own_filter'));
+    } else {
+      if (_.isNumber(this.get('start'))) {
+        params.push('start=' + this.get('start'));
+      }
+      if (_.isNumber(this.get('end'))) {
+        params.push('end=' + this.get('end'));
+      }
+      if (_.isNumber(this.get('bins'))) {
+        params.push('bins=' + this.get('bins'));
+      }
     }
     if (this.get('boundingBox') && this.get('submitBBox')) {
       params.push('bbox=' + this.get('boundingBox'));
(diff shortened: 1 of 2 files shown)


src/dataviews/histogram-dataview-model.js
```diff
@@ -18,17 +18,18 @@ module.exports = DataviewModelBase.extend({
...
-    if (_.isNumber(this.get('start'))) {
-      params.push('start=' + this.get('start'));
-    }
-    if (_.isNumber(this.get('end'))) {
-      params.push('end=' + this.get('end'));
-    }
-    if (_.isNumber(this.get('bins'))) {
-      params.push('bins=' + this.get('bins'));
-    }
     if (_.isNumber(this.get('own_filter'))) {
```
javisantana: I don't exactly get this, own_filter and setting up bins (or even start and end) could make sense from the dataview point of view

src/dataviews/histogram-dataview-model.js
```diff
@@ -48,12 +49,29 @@ module.exports = DataviewModelBase.extend({
     // BBox should only be included until after the first fetch, since we want to get the range of the full dataset
     this.once('change:data', function () {
       this.set('submitBBox', true);
+
+      var data = this.getData();
+      if (data && data.length > 0) {
+        this.set({
```
javisantana: be careful with calling `set` on the initialize method, if you want to do this sometimes is better to change attributes directly. The reason is Backbone keeps changes for later so if you do `set('other_attr', 'test')` later you will get `change:bins` triggered

src/dataviews/histogram-dataview-model.js
javisantana: I understand why you did this, just want to point it

PR #1122, opened 2016-02-19
Explicitly fetch data from data provider when map bounds change

### LINKED ISSUE #216: Errror in Firefox (v32 and v33) in torque visualizations
[TrackJS](https://my.trackjs.com/messages/filtered?message=Index+or+size+is+negative+or+greater+than+the+allowed+amount)

"Index or size is negative or greater than the allowed amount"

alonsogarciapablo:
Partially fixes CartoDB/deep-insights.js#216.

@acanimal could you please take a look? Thanks!


alonsogarciapablo:
I've added one commit more where I've turned the geojson-data-provider into a factory that is used to return specific data providers for each type of dataview.

cc: @acanimal @fdansv

acanimal:
Fine for me. IMO it look more clear with the provider factory.

PR #1124, opened 2016-02-22
Filters has a reference to the Layer instead a `layerIndex` property

### LINKED ISSUE #1020: Associate filters directly to map.layers

acanimal:
CR @alonsogarciapablo 

close #1020 

- Filters has a reference to the Layer instead had a `layerIndex` property.
- WindshaftMap and WindshaftFilterCollection are updated to compute the list of filters grouped by layer.

### DIFF
--- a/src/windshaft/filters/collection.js
+++ b/src/windshaft/filters/collection.js
@@ -14,24 +14,32 @@ module.exports = Backbone.Collection.extend({
   },
 
   toJSON: function () {
-    var json = {
-      layers: []
-    };
-    var l, f, filters, filtersJson;
-    for (l = 0; l < this.layers.length; l++) {
-      var layer = this.layers[l];
-      if (layer.isVisible()) {
-        filters = this.getActiveFiltersForLayer(layer);
-        filtersJson = {};
-        if (filters.length) {
-          for (f = 0; f < filters.length; f++) {
-            _.extend(filtersJson, filters[f].toJSON());
+    var json = {};
+    var activeFilters = this.getActiveFilters();
+    if (activeFilters.length) {
+      json.layers = [];
+      var l, f, filters, filtersJson;
+      for (l = 0; l < this.layers.length; l++) {
+        var layer = this.layers[l];
+        if (layer.isVisible()) {
+          filters = this.getActiveFiltersForLayer(layer);
+          filtersJson = {};
+          if (filters.length) {
+            for (f = 0; f < filters.length; f++) {
+              _.extend(filtersJson, filters[f].toJSON());
+            }
           }
+          json.layers.push(filtersJson);
         }
-        json.layers.push(filtersJson);
       }
     }
     return json;
+  },
+
+  getActiveFilters: function () {
+    return this.filter(function (filter) {
+      return !filter.isEmpty();
+    });
   },
 
   getActiveFiltersForLayer: function (layer) {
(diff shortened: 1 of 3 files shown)


src/windshaft/filters/collection.js
```diff
@@ -2,44 +2,39 @@ var _ = require('underscore');
...
+   * @param  {Array.<Filter>} filters Array with the filters to add to the collection.
+   * @param  {Array.<LayerModel>} layers  Array of LayerModel in the same order passed to the WindshaftMap.
+   * @returns {WindshaftFilterCollection}
+   */
   initialize: function (filters, layers) {
     this.layers = layers;
     Backbone.Collection.prototype.initialize.apply(this, filters);
   },
 
   toJSON: function () {
```
alonsogarciapablo: Can we add some tests for this method?

acanimal: I updated the tests. They already cover use cases with more than one layer and with one hidden layer.

alonsogarciapablo: If there were tests for this and everything was working fine... why do we need to change the tests? :astonished:

acanimal: Because output is a bit different. Now we always have an object with filters per layer (although no filter was applied on the layer). In summary, if our map has three layers without filters we get three objects. 

Before we only generate output if there were any "active filter". If there were no active filter we get an empty array. That is the only difference with current change.

If you think it is important I can change it.

alonsogarciapablo: IMO it should continue working as it used to work before, unless there a reason to change it, which doesn't seem to be the case...

src/windshaft/filters/collection.js
```diff
@@ -12,28 +18,21 @@ module.exports = Backbone.Collection.extend({
     var activeFilters = this.getActiveFilters();
     if (activeFilters.length) {
```
acanimal: @alonsogarciapablo  I tried to generate the same previous output but there is a case not as easy as can seem. Previously when we worked with layerIndex we generate filters for a given layerIndex and fill the "layer filter group" gaps. Now the process is different, we create a "layer filter group" and add all the layer filters.

For the next test https://github.com/CartoDB/cartodb.js/pull/1124/files#diff-1bff0c96a4fa609337cd94e943c2bfe5R109 is not as easy. Previously given three layers that only the first one has filters the out was `{ layers: [{ filters_layer_1}] }`. With the new approach it is `{ layers: [{ filters_layer_1}, {}, {}] }` and it is not easy to know if the latests "filter groups" can be removed. 

It is the only case, for other cases, for example when layer one and three has filter the output is the same: `{ layers: [{ filters_layer_1}, {}, {filter_layer_3}] }`

alonsogarciapablo:
:+1:

PR #1125, opened 2016-02-22
Debounce map instantiation through map.reload

### LINKED ISSUE #1096: debounce map instanciation
One of the problems with the map instanciation is when you have something like:

```
layer.setSQL('....')
//code
layer.setSQL('....')
```

In this case you instanciate the map twice but the user only sees one update. To avoid that a debounce was added so the instanciation only happens one time if the consecutive calls are done really fast.

That was implemented here: https://github.com/CartoDB/cartodb.js/blob/develop/src/geo/layer_definition.js#L120 but looks like with the refactor we removed it. 

This is really important, specially right now with the widgets where the initial setup is really hard.

At programming level is a hard part, the previous class had really **a lot** of work to avoid race conditions.

cc @alonsogarciapablo

alonsogarciapablo:
Fixes #1096.

@javisantana what do you think?

### DIFF
--- a/src/geo/map.js
+++ b/src/geo/map.js
@@ -21,7 +21,7 @@ var Map = Model.extend({
     provider: 'leaflet'
   },
 
-  RELOAD_DEBOUNCE_TIME: 100,
+  RELOAD_DEBOUNCE_TIME: 10,
 
   initialize: function (attrs, options) {
     options = options || {};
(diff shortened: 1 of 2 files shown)


javisantana:
Some questions and comments:

- 0.5s of delay for every layer change sounds like too much, it should be something like 10ms (right now is 4ms)
- it still has a problem: if a request arrive and another one is being done the tiles and widgets are still loaded (they will be there for a short amount of time). Not really important right now

alonsogarciapablo:
Decreased delay to 10s.

javisantana:
:es:

PR #1129, opened 2016-02-23
Cancel old requests that comes from dataviewModels

xavijam:
Fixes part of https://github.com/CartoDB/deep-insights.js/issues/236.
No sure if it could break something from D3 integration.

CR: @javisantana
@viddo @alonsogarciapablo @fdansv @rochoa

PS: Deep insights PR also coming.

### DIFF
--- a/src/dataviews/dataview-model-base.js
+++ b/src/dataviews/dataview-model-base.js
@@ -171,12 +171,36 @@ module.exports = Model.extend({
     return this.previous('data');
   },
 
+  sync: function (method, model, options) {
+    var self = arguments[1];
+    if (this._xhr) {
+      this._xhr.abort();
+    }
+    this._xhr = Model.prototype.sync.apply(this, arguments);
+    this._xhr.always(function () {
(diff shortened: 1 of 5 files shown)


xavijam:
before you say, test is coming too.

src/dataviews/dataview-model-base.js
```diff
@@ -173,10 +176,22 @@ module.exports = Model.extend({
...
+    if (opts.success) {
+      var successCallback = opts && opts.success;
+    }
+
     return Model.prototype.fetch.call(this, _.extend(opts, {
-      error: function () {
-        this.trigger('error');
+      success: function () {
+        successCallback && successCallback(arguments);
+        this.trigger('loaded', this);
```
xavijam: I forgot to mention that we are using this "custom" event due to we need any event to listen in Deep-insights elements. Before we used to use "sync" event, but it doesn't work pretty good because sync doesn't mean the request has been done/finished.

javisantana:
:es:

PR #1130, opened 2016-02-23
Using current window protocol for widget fetch url

xavijam:
Spotted while we were testing widgets with namedmaps, where all widgets requests were under `http`.

Do I forget something? Good solution?

@alonsogarciapablo @viddo 
cc @juanignaciosl


alonsogarciapablo:
:+1: Good enough for me.

PR #1132, opened 2016-02-24
Fixes infowindows and tooltips

alonsogarciapablo:
In this PR I changed how we're creating new infowindows and tooltips right now. Instead of listening to the "newLayerView" event and adding infowindows/tooltips when a new layerview is added, we now work directly with the array of layerModels.

In an attempt to make things a bit more simpler and clean up the vis#load method, I've extracted the logic to manage infowindows and tooltips to what I've called an InfowindowManager and a TooltipManager. I'm open to better names.

@javisantana could you please tell me what you think about this approach? Thanks!

### DIFF
--- a/src/vis/infowindow-manager.js
+++ b/src/vis/infowindow-manager.js
@@ -55,9 +55,7 @@ InfowindowManager.prototype._bindFeatureClickEvent = function (layerView, layerM
     }
     var cartoDBId = data.cartodb_id;
 
-debugger;
     layerView.model.fetchAttributes(layerIndex, cartoDBId, function (attributes) {
-debugger;
       // Old viz.json doesn't contain width and maxHeight properties
       // and we have to get the default values if there are not defined.
       var extra = _.defaults(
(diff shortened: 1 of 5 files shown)


src/vis/infowindow-manager.js
```diff
@@ -0,0 +1,103 @@
...
+};
+
+InfowindowManager.prototype.manage = function (mapView, map) {
+  this._mapView = mapView;
+  this._map = map;
+
+  this._map.layers.bind('reset', function (layers) {
+    layers.each(this._addInfowindowForLayer, this);
+  }, this);
+  this._map.layers.bind('add', this._addInfowindowForLayer, this);
```
javisantana: the remove? I think I saw the "remove" event on the view, not sure it is enough

alonsogarciapablo: Are you refering to [this](https://github.com/CartoDB/cartodb.js/pull/1132/files#diff-7d5ed7ef009eb2aac7940c8a0a158332R48)?

alonsogarciapablo: What would be the behaviour you'd expect when a layer is removed from `map.layers`?

PR #1135, opened 2016-02-24
Not adding leaflet layer each time map is instantiated

xavijam:
Seems like our very own [weird-zoom-bug](https://github.com/CartoDB/deep-insights.js/issues/249) is due to a problem of adding the same layer each time the map is instantiated. For example, when a filter is applied. Seems like @javisantana and I have cleaned a little bit the old code.

CR: @javisantana 
cc @alonsogarciapablo 

No more :scream: faces showing a Deep-Insights bug.

### DIFF
--- a/src/geo/leaflet/leaflet-cartodb-layer-group-view.js
+++ b/src/geo/leaflet/leaflet-cartodb-layer-group-view.js
@@ -211,6 +211,8 @@ var LeafletCartoDBLayerGroupView = L.TileLayer.extend({
     if (tilejson) {
       this.tilejson = tilejson;
       this.setUrl(this.tilejson.tiles[0]);
+      // TODO: Is this necessary?
+      this.ok && this.ok();
       // manage interaction
       this._reloadInteraction();
     } else {
--- a/src/geo/leaflet/leaflet-cartodb-layer-group-view.js
+++ b/src/geo/leaflet/leaflet-cartodb-layer-group-view.js
@@ -211,10 +211,10 @@ var LeafletCartoDBLayerGroupView = L.TileLayer.extend({
     if (tilejson) {
       this.tilejson = tilejson;
       this.setUrl(this.tilejson.tiles[0]);
-      // TODO: Is this necessary?
-      this.ok && this.ok();
       // manage interaction
       this._reloadInteraction();
+      // TODO: Is this necessary?
+      this.ok && this.ok();
     } else {
       this.error && this.error('URLs have not been fetched yet');
     }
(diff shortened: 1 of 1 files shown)


javisantana:
Since the instanciation is done outside of the layer the code is simplified, I remember when I had to do that hack.

:es: \o/ \o/ \o/ \o/ :es:

src/geo/leaflet/leaflet-cartodb-layer-group-view.js
```diff
@@ -219,23 +207,15 @@ var LeafletCartoDBLayerGroupView = L.TileLayer.extend({
...
    * do not collide with leaflet _update
    */
-  __update: function (done) {
-    var self = this;
-    this.fire('updated');
-    this.fire('loading');
-    var map = this.options.map;
-
-    var tilejson = self.model.get('urls');
+  __update: function () {
```
alonsogarciapablo: Is there a reason why this has to be called `__update`? Otherwise, I would call it `onTileJSONChanged`...

xavijam: done!

alonsogarciapablo:
Simpler! :shipit:

src/geo/leaflet/leaflet-cartodb-layer-group-view.js
```diff
@@ -215,27 +202,19 @@ var LeafletCartoDBLayerGroupView = L.TileLayer.extend({
...
+  _onTileJSONChange: function () {
+    var tilejson = this.model.get('urls');
     if (tilejson) {
-      self.tilejson = tilejson;
-      self.setUrl(self.tilejson.tiles[0]);
+      this.tilejson = tilejson;
+      this.setUrl(this.tilejson.tiles[0]);
       // manage interaction
-      self._reloadInteraction();
-      self.ok && self.ok();
```
xavijam: I've removed this because ok function only triggers an event called `tileOk`. Is it necessary? Rings the bell @javisantana?
I've checked rest of code and nobody in CartoDB.js is listening that event.
http://www.gifbay.com/gif/bola_del_desierto-143119/

alonsogarciapablo: Can we remove the `ok` function then?

xavijam: @javisantana?

xavijam: Oh wait, I've seen we use it within CartoDB editor, let's keep it for the time being :school: .

javisantana: I remember I added it because of something but I don't remember that something. I'd drop it

PR #1136, opened 2016-02-24
Enforce client-side rendering with "vector" attribute in the viz.json

alonsogarciapablo:
This change replaces the current mechanism that we use to determine when to render client-side. With this change, viz.json accepts a new `vector` option that can be used to enforce client-side rendering. I added an example that uses this new option(`examples/v4/leaflet-layergroup-vector.html`).

@javisantana yay or nay?

### DIFF
--- a/examples/v4/leaflet-layergroup-vector.html
+++ b/examples/v4/leaflet-layergroup-vector.html
@@ -276,6 +276,7 @@
               "order": 3,
               "type": "tiled"
             }
+
           ],
           "overlays":[
             {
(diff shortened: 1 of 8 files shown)


examples/v4/leaflet-layergroup-vector.html
```diff
@@ -0,0 +1,329 @@
...
+      function main() {
+      	var vizjson = {
+          "id":"2b13c956-e7c1-11e2-806b-5404a6a683d5",
+          "version":"0.1.0",
+          "title":"european_countries_e 0",
+          "likes":0,
+          "description":null,
+          "scrollwheel":true,
+          "legends":true,
+          "vector":true,
```
alonsogarciapablo: :point_up: this is the new option.

javisantana:
:es:

PR #1139, opened 2016-02-24
Disable filter and unset range on bins, start, and end change on a histogram dataview

viddo:
Extracted from https://github.com/CartoDB/deep-insights.js/blob/96c5b72846d5c6c13c97c10260f435ba60f6b538/src/widgets/histogram/content-view.js#L314-L315 since
this should apply for all histograms, including time-series (which is also why start and end are included)

Required to solve https://github.com/CartoDB/deep-insights.js/issues/244 (PR: https://github.com/CartoDB/deep-insights.js/pull/253)

@alonsogarciapablo can you review?

### DIFF
--- a/src/dataviews/histogram-dataview-model.js
+++ b/src/dataviews/histogram-dataview-model.js
@@ -61,7 +61,7 @@ module.exports = DataviewModelBase.extend({
     }, this);
     this.listenTo(this.layer, 'change:meta', this._onChangeLayerMeta);
     this.on('change:column', this._reloadMap, this);
-    this.on('change:bins change:start change:end', this._refreshAndResetFilter, this);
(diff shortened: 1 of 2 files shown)


src/dataviews/histogram-dataview-model.js
```diff
@@ -135,6 +135,12 @@ module.exports = DataviewModelBase.extend({
         this.trigger('histogram_sizes', this);
       }
     }, this);
+  },
+
+  _refreshAndResetFilter: function () {
+    this.refresh();
```
alonsogarciapablo: [`refresh` is actually just fetching](https://github.com/CartoDB/cartodb.js/blob/v4/src/dataviews/dataview-model-base.js#L160-L162) so perhaps we could use `fetch` here.

alonsogarciapablo:
Left a minor comment :+1:

Issue #1137, opened 2016-02-24
Add ` addCursorInteraction` method to the docs?


csobier:
Trying to help @michellechandra  confirm her [tutorial](http://docs.cartodb.com/tutorials/google_driving_directions/) steps... 

Found [` addCursorInteraction`](https://github.com/CartoDB/cartodb.js/blob/3798323b953318576fac7ef00dede0aeec81e978/src/vis/vis.js#L1502) and also the [`remove cursor interaction`](https://github.com/CartoDB/cartodb.js/blob/3798323b953318576fac7ef00dede0aeec81e978/src/vis/vis.js#L1513) method.  Doesn't `featureClick` do the same thing? 
- Why would you use one method over the other?
- Should `add & remove CursorInteraction` be added to the API docs?

@alonsogarciapablo , please advise.

alonsogarciapablo:
I'm not familiar with `addCursorInteraction` and `removeCursorInteraction` at all so I'm not sure if they are actually needed in the tutorial. I'd try to run the example in the tutorial without the usage of `addCursorInteraction` and see if everything still works fine (cursor changes when hovering over a feature).... Otherwise, we'll have to ask @javisantana, which probably knows about the story behind these two methods....

michellechandra:
Testing with my tutorial code and finding that the mouse pointer changes to a finger point with the addCursorInteraction method (missing if I remove and just have the featureClick event.) Prob worth documenting!

csobier:
As per @javisantana, this is an internal object that the user shouldn't know about. Closing as "wontfix" and deleting the related PR.  Michelle will update her tutorial accordingly.

PR #1140, opened 2016-02-24
added addCursorInteraction and removeCursorInteraction to cartodb.Vis…

csobier:
… API methods

Fixes: #1137. @javisantana , I took a guess here as to where to put this and used the examples and resources that I found. Let me know if this is correct, or if there is more to the story that I should know about.

### DIFF
--- a/doc/api_methods.md
+++ b/doc/api_methods.md
@@ -141,7 +141,7 @@ addCursorInteraction: function(map, layer) {
     });
 {% endhighlight %}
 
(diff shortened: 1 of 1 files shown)


javisantana:
why did you write the functions code in the doc. They use internal objects that the user shouldn't know about

csobier:
Ah, that is what we were trying to figure out in #1137.  I'll ask Michelle to see if she can find an alternative in her tutorial.

csobier:
Just curious as to why there was an example in the [CartoDB.js/Examples](https://github.com/CartoDB/cartodb.js/blob/develop/examples/cursor_interaction.html) folder if this was supposed to be internal though? I will delete this PR and close the related issue as "wontfix".  Thanks!

PR #1141, opened 2016-02-25
Remove the possibility to edit start/end values of a histogram

viddo:
After discussion w/ @javisantana regarding problematic use-cases related to changing start/end values for a time-series histogram, we concluded that it’s better it works like a normal histogram, i.e. the data range gets set on the initial fetch (calculated server-side based on the dataset), and can’t be changed after that.

I'll update DI and editor after this is merged (to remove the possibility to change the start/end dates).

cc @alonsogarciapablo


javisantana:
I'd add a ticket to do this as enhancement

PR #1143, opened 2016-02-25
Ensure featureClick and featureOver events are bound once

alonsogarciapablo:
There was a problem with how we were binding featureClick and featureOver events to layerView. We're basically binding the same events multiple times and when one of these two events, handlers were being executed multiple times (one per layer). This PR fixes it.

@javisantana :es: please

### DIFF
--- a/src/vis/tooltip-manager.js
+++ b/src/vis/tooltip-manager.js
@@ -53,8 +53,8 @@ TooltipManager.prototype._addTooltipOverlay = function (layerView, layerModel) {
 
 TooltipManager.prototype._bindFeatureOverEvent = function (layerView) {
   var tooltipView = layerView.tooltipView;
-  layerView.unbind('featureOver');
-  layerView.bind('featureOver', function (e, latlng, pos, data, layerIndex) {
+
+  var onFeatureOver = function (e, latlng, pos, data, layerIndex) {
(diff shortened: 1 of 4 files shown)


src/vis/tooltip-manager.js
```diff
@@ -51,15 +51,24 @@ TooltipManager.prototype._addTooltipOverlay = function (layerView, layerModel) {
   }
 };
 
-TooltipManager.prototype._bindFeatureOverEvent = function (layerView, layerModel) {
+TooltipManager.prototype._bindFeatureOverEvent = function (layerView) {
   var tooltipView = layerView.tooltipView;
-  layerView.bind('featureOver', function (e, latlng, pos, data, layer) {
+  layerView.unbind('featureOver');
```
javisantana: be careful with this, you could be removing `featureOver` events from other objects.
You should do something like:
```
var f = function() {....}
layerView.unbind('featureOver', f);
layerView.bind('featureOver', f);
```

alonsogarciapablo: Good point! Please take a look at [35711fed22060b20243d1d6ac71133cee0fba8d0](https://github.com/CartoDB/cartodb.js/commit/35711fed22060b20243d1d6ac71133cee0fba8d0)

PR #1145, opened 2016-02-25
Forcing dataview model fetch new data although sync_on_data_change attribute is disabled

### LINKED ISSUE #1144: If widget has sync_on_data_change disabled, any widget form change doesn't refresh the widget
@xavijam commented on [Wed Feb 24 2016](https://github.com/CartoDB/cartodb/issues/6665)

### STP
- Create a widget (no matter the type).
- Go to style tab within the widget edit/options view.
- Select No for `Unfiltered` attribute.
- Go to data tab and change any other attribute.
- Widget doesn't change when new data arrives.

cc @viddo :joy:

xavijam:
Basically, if a widget has disabled sync_on_data_change, it will not change if there is an update of its 'column'.
We need to test this but I'd like your opinion before doing that. How does this approach sound?

- [ ] Create necessary test(s) for this behaviour.

CR: @alonsogarciapablo

Fixes #1144

### DIFF
--- a/src/dataviews/category-dataview-model.js
+++ b/src/dataviews/category-dataview-model.js
@@ -42,7 +42,7 @@ module.exports = DataviewModelBase.extend({
     this._data = new CategoriesCollection();
     this._searchModel = new SearchModel();
 
-    this.on('change:column change:aggregation change:aggregation_column', this._reloadMap, this);
+    this.on('change:column change:aggregation change:aggregation_column', this._reloadMapAndForceFetch, this);
   },
 
   // Set any needed parameter when they have changed in this model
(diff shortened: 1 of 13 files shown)


viddo:
I think it would be better (simpler) to have a `.forceFetch()` method of some sort, since the requirements for _editing_ dataview (through a widget) is not the same as for a _using_ them. Having a bunch of implicit behaviors based on various flags states makes it really hard to reason about what's going on.

alonsogarciapablo:
We'd need to something like:

```
// Somewhere in the histogram dataview...
this.on('change:column', this._reloadMapAndForceFetch, this);
```

In the dataview-model-base.js:

```
  _reloadMapAndForceFetch: function () {
    this._map.reload({
      sourceLayerId: this.layer.get('id'),
      forceFetch: true
    });
  }
```

We would need to propagate that "forceFetch" (as we do with the `sourceLayerId`) so that we can finally do something like:

```
// dataview-model-base -> onChangeBinds
    this.on('change:url', function (model, attribute, options) {
      if (options.forceFetch || this._shouldFetchOnURLChange()) {
        this.fetch();
      }
    }, this);
```

xavijam:
@alonsogarciapablo ^^^ that approach?

alonsogarciapablo:
Yes, that can work.

xavijam:
Could you take a look now @alonsogarciapablo?

alonsogarciapablo:
LGTM! :clap:

PR #1156, opened 2016-02-29
Reduces usage of getValues

fdansv:
cc @alonsogarciapablo


src/geo/data-providers/geojson/category-geojson-data-provider.js
```diff
@@ -13,9 +13,11 @@ CategoryGeoJSONDataProvider.prototype.getData = function () {
   var filterEnabled = options.filterEnabled;
   var numberOfCategories = 5;
   var filter = this._vectorLayerView.getFilter(this._layerIndex);
-  var features = this._getFeatures();
```
alonsogarciapablo: We should get rid of `GeoJSONDataProviderBase#getFeatures` if we're not using it anymore.

fdansv: righto!

PR #1160, opened 2016-03-01
Updating dataview boundingBox although url has not been set yet

xavijam:
CR: @alonsogarciapablo

### DIFF
--- a/src/dataviews/dataview-model-base.js
+++ b/src/dataviews/dataview-model-base.js
@@ -152,7 +152,7 @@ module.exports = Model.extend({
   },
 
   _onChangeBinds: function () {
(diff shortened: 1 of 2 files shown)


src/dataviews/dataview-model-base.js
```diff
@@ -151,6 +152,7 @@ module.exports = Model.extend({
   },
 
   _onChangeBinds: function () {
+    this.stopListening(this._map);
```
alonsogarciapablo: `this.stopListening(this._map, 'change:center', this._updateBoundingBox)`?

xavijam:
Closing since we have another solution

PR #1162, opened 2016-03-01
Dataviews get bounding box from map object

### LINKED ISSUE #1158: [vector] Result data (using widgets) in vector rendering differs with raster data
### STR

- Open populated-places example in DI (raster way).
- Open populated-places example in DI with vector enabled (in another window).
- Compare results from widgets, they are totally different in all widgets.

untitled

Assign it accordingly @rochoa @fdansv @alonsogarciapablo

alonsogarciapablo:
Partially fixes #1158. The bounding box that widgets were using to get information from Windshaft was not right.

@xavijam please review. Thanks!

cc: @fdansv

### DIFF
--- a/src/dataviews/dataview-model-base.js
+++ b/src/dataviews/dataview-model-base.js
@@ -17,11 +17,15 @@ module.exports = Model.extend({
   },
 
   url: function () {
-    var params = [];
-    if (this.get('boundingBox')) {
-      params.push('bbox=' + this.get('boundingBox'));
-    }
+    var params = [
+      'bbox=' + this._getBoundingBoxFilterParam()
+    ];
     return this.get('url') + '?' + params.join('&');
+  },
+
+  _getBoundingBoxFilterParam: function () {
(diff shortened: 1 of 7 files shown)


src/dataviews/dataview-model-base.js
```diff
@@ -127,6 +130,10 @@ module.exports = Model.extend({
         silent: silent,
         forceFetch: forceFetch
       });
+
+      if (this.get('sync_on_data_change')) {
+        this._newDataAvailable = true;
```
xavijam: What is this?

alonsogarciapablo: That flag is used [here](https://github.com/CartoDB/cartodb.js/pull/1162/files#diff-805364e485d740e26e33839755aab753R161).

alonsogarciapablo:
Ping!

xavijam:
:dancer: :+1: :shipit:

PR #1164, opened 2016-03-01
updated refreshTime descrip

csobier:
Tiny, minor description update. @michellechandra , can you sanity check this description based on the Stack Exchange message we discussed?


michellechandra:
@csobier :+1:

PR #1166, opened 2016-03-02
Fixing problem with first request in histogram dataview

xavijam:
We forgot to use `submitBbox` attribute in order to get the data without any filter (first load) and then get it with all filter parameters.

CR: @alonsogarciapablo


xavijam:
Also, seems like in vector rendering, we are not taking into account this thing either.

alonsogarciapablo:
:+1: Good catch!

Issue #1142, opened 2016-02-25
Update torque layer animation range on histogram 


viddo:
The possibility to change start/end values of a histogram (time-series use-case) was removed in #1141, since it had non-trivial implications on how to sync it with the related torque layer.

It should be allowed to change, but need to think through how to should work, in particular figure out how the torque layer should be updated.

viddo:
cc @javisantana, feel free to prioritize to a milestone if you think it should be done anytime soon

nobuti:
There has been no activity on this issue for several months and that's why we are closing it (automatically). If you think this still needs to be addressed please reopen it. The good news is that we are now working on the next version of CARTO.js, stay tuned.

PR #1173, opened 2016-03-04
Possibility to set start or/and end params for histogram initialization

xavijam:
Not sure if we should reset histogram widget if any of those params have changed (in Deep-insights).

CR: @alonsogarciapablo 
cc @AbelVM

### DIFF
--- a/src/dataviews/histogram-dataview-model.js
+++ b/src/dataviews/histogram-dataview-model.js
@@ -148,7 +148,9 @@ module.exports = DataviewModelBase.extend({
     ATTRS_NAMES: DataviewModelBase.ATTRS_NAMES.concat([
       'column',
       'column_type',
-      'bins'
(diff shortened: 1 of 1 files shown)


xavijam:
cc @viddo too

alonsogarciapablo:
I'm not sure about the implications of this change... :(

AFAIK `start` and `end` are some internal attributes that the histogram uses once the first histogram is fetched.

I think @javierarce can help us here too...

viddo:
See [last PR #1141](https://github.com/CartoDB/cartodb.js/pull/1141) and the pending issue #1142, which explains why start/end was removed here. Adding them back won't solve that, it's not that easy unfortunately.

xavijam:
ok, it was a good addition for a PoC, but let's face it in the proper way, not this easy one.

Issue #1172, opened 2016-03-03
Make loop: boolean an option for CreateVis


michellechandra:
It is possible to pass loop: boolean as an option to createLayer for Torque map, this option should also be available to maps created using createVis but does not seem to work.

**Sample Code**
```
<!DOCTYPE html>
<html>
  <head>
    <title>Pause Torque</title>
    <meta name="viewport" content="initial-scale=1.0, user-scalable=no" />
    <meta http-equiv="content-type" content="text/html; charset=UTF-8"/>
    <link rel="shortcut icon" href="http://cartodb.com/assets/favicon.ico" />
    <style>
      html, body, #map {
        height: 100%;
        padding: 0;
        margin: 0;
      }
    </style>
  <link rel="stylesheet" href="http://libs.cartodb.com/cartodb.js/v3/3.15/themes/css/cartodb.css" />
  </head>
  <body>
    <div id="map"></div>
 
    <!-- include cartodb.js library -->
   <script src="http://libs.cartodb.com/cartodb.js/v3/3.15/cartodb.js"></script>
 
    <script>
      function main() {
        
        cartodb.createVis('map', 'http://andrew.cartodb.com/api/v2/viz/b5cece38-4af4-11e3-bfb4-3085a9a9563c/viz.json', { loop: false });
      }
      window.onload = main;
    </script>
  </body>
</html>
```

michellechandra:
@csobier

alonsogarciapablo:
Closing this for now since it might not be relevant in `v4`.

PR #1174, opened 2016-03-04
added loop to createLayer options

csobier:
For [Docs Issue#717](https://github.com/CartoDB/docs/issues/717).  @michellechandra , please review and advise.
- I added this an an option to createLayer and linked to the Torque.js doc. (Loop is already documented in the Torque API methods). 
- This is only applicable to createLayer, not createVis, right?
- Does this also need to be included in the [MapConfig, Torque Layer Options](http://docs.cartodb.com/cartodb-platform/maps-api/mapconfig/#torque-layer-options)?


michellechandra:
@csobier - looks good, here is an issue I opened re: loop not part of createVis: https://github.com/CartoDB/cartodb.js/issues/1172

csobier:
@alonsogarciapablo , can I add this update to the docs if this looks OK to you?  (Also, please advise if this should also be added to the Map Configs- Torque layer doc?)

alonsogarciapablo:
:+1: If that's how it works. I realized you're using \`\`\`true\`\`\` and \`\`\`false\`\`\`, but \`true\` would work too (simpler)

csobier:
Thanks!  I will merge once I figure out what is happening with the code (I didn't write it like that, something is happening programmatically with the code in the docs. I've asked @matallo to look into this).

xavijam:
retest this please

csobier:
Thanks for the fix @xavijam !

PR #1175, opened 2016-03-04
Layers API for the editor

### LINKED ISSUE #1169: Layers API
**WARNING: This is a WIP. I will remove this comment once I complete the description based on the feedback from @viddo @javierarce  @xavijam**

Here's an initial proposal for the API that the editor needs to work with CartoDB.js layers:

**Creation of new layers**
```
var layer = map.createCartoDBLayer(attrs);
var layer = map.createTorqueLayer(attrs);
var layer = map.createTileLayer(attrs);
// ... (we probably have more types (color, wms, etc.))
```

**Deletion of an existing layer**

```
layer.remove();
```

**Update of an existing layer**

```
// Update a layer
layer.update(attrs)
```

**Showing/hiding a layer**

```
layer.show();
layer.hide();
```

alonsogarciapablo:
Fixes #1169.

Here're some of the highlights of this PR:

- Introduced `map.createCartoDBLayer(...)`, `map.createTorqueLayer(...)`, etc.
- ~~I extracted a ModuleLoader class that takes care of loading modules for specific layer models (only torque layers have to load external modules at the moment).~~
- ~~Modules are now loaded right before the layer view is generated by the MapView. Before, this process was being done right after the viz.json was loaded. With this change it's now possible to add a Torque layer using `map.createTorqueLayer({...})` and ensure that the module will be loaded (and vis#load is a bit cleaner now \o/).~~
- Removed mechanism to load torque dinamically as suggested by @javisantana in [this comment](https://github.com/CartoDB/cartodb.js/pull/1175#issuecomment-193209657).
- Renamed src/geo/map/map-layer.js to src/geo/map/layer-model-base.js

@viddo could you please CR! Thanks!

cc: @javisantana @javierarce @xavijam


javisantana:
From my point of view having a module loader is something we don't need anymore. The original idea for the module loader was to avoid to load the cartocss parser all the times but now that we have d3.cartodb all the time does not make any sense to have all that complexity to load a few kb.

Loading modules dinamically is also pretty bad since browserify can't do the work is supposed to do

alonsogarciapablo:
@javisantana good point. I'm all about :fire: code so, even if I'm really happy with our shiny new `ModuleLoader`, I will :fire: it and make sure torque is always loaded. I think it'll simplify code/tests overall.

I'll comment here once it's done.

alonsogarciapablo:
:fire: all module loading related stuff. Torque is now required via browserify.

@viddo ready for CR again.

grunt/tasks/_browserify-bundles.js
```diff
@@ -49,30 +41,11 @@ module.exports = {
...
-    options: {
-      external: sharedModules
-    }
-  },
-  'cartodb.mod-torque-srcSpecs': {
-    src: torqueSpecs,
-    dest: '<%= config.tmp %>/cartodb.mod.torque-srcSpecs.js',
     options: {
-      external: sharedModules
+      require: sharedModules
```
viddo: I think we could remove this altogether, since the module loading code was removed, since there's no need to load things in a separate bundle, right?

src/geo/cartodb-layer-group-base.js
```diff
@@ -1,15 +1,13 @@
 var $ = require('jquery');
 var Backbone = require('backbone');
-var MapLayer = require('./map/map-layer');
 var util = require('cdb.core.util');
 
-var CartoDBLayerGroupBase = MapLayer.extend({
+var CartoDBLayerGroupBase = Backbone.Model.extend({
```
viddo: Is there some reason for not use the default core model? i.e.
 ```js
var Model = require('../core/model');`
var CartoDBLayerGroupBase = Model.extend({
```

File: src/geo/cartodb-layer-common.js [hunk 3]
```diff
@@ -97,6 +97,7 @@ CartoDBLayerCommon.prototype = {
     return this;
   },
 
+  // TODO: This can be removed
```
viddo: So, can the method be removed? Otherwise I'd say the comment should be removed, or clarify what's pending for it to be removed or such.

src/vis/vis.js
```diff
@@ -215,46 +187,20 @@ var Vis = View.extend({
...
     if (typeof (data) === 'string') {
       var url = data;
-
       Loader.get(url, function (data) {
         if (data) {
-          self.load(data, options);
+          this.load(data, options);
         } else {
-          self.throwError('error fetching viz.json file');
+          this.throwError('error fetching viz.json file');
```
viddo: Is this really correct? The code still looks like its executed in a async closure for which `this` might not be the `this` of the load function.

viddo: Nevermind this comment, just saw the `.bind(this)` below… (will delete comment…)

src/vis/vis.js
```diff
@@ -416,6 +362,10 @@ var Vis = View.extend({
     // Lastly: reset the layer models on the map
     this.map.layers.reset(layerModels);
 
+    // Global variable for easier console debugging / testing
+    window.vis = this;
+    window.map = map;
```
viddo: While I understand the intents of exposing this (like explained in the comment just above), setting these directly on window could possibly overwrite someone's vars (that's using cartodb.js for a standalone app), do we really want to expose them here?

viddo:
:shipit:

grunt/tasks/_browserify-bundles.js
alonsogarciapablo: Yes. Makes sense! Thanks!

src/geo/cartodb-layer-group-base.js
alonsogarciapablo: > Is there some reason for not use the default core model? i.e.

No. But there isn't a reason for using it either, so not a big deal IMO.

grunt/tasks/_browserify-bundles.js
alonsogarciapablo: :+1: Good catch!

src/geo/cartodb-layer-common.js
alonsogarciapablo: Fair point. Removing method and tests.

PR #1177, opened 2016-03-04
[vector] Gets style from meta when map is named

### LINKED ISSUE #1176: [vector] Named maps should get cartocss from metadata
As otherwise it is not possible to fetch styles.

fdansv:
Fixes #1176 

Needs merge of https://github.com/CartoDB/Windshaft/pull/457 for it to work. Merging won't break non-named maps though.

### DIFF
--- a/src/geo/leaflet/leaflet-cartodb-vector-layer-group-view.js
+++ b/src/geo/leaflet/leaflet-cartodb-vector-layer-group-view.js
@@ -28,15 +28,10 @@ var LeafletCartoDBVectorLayerGroupView = CartoDBd3Layer.extend({
     }, this);
 
     layerModel.layers.bind('change:meta', function (child, meta) {
-      if (this.model.get('type') === 'namedmap') {
-        var index = child.collection.indexOf(child);
-        if (child.collection.at(0).get('type') === 'tiled') {
-          index++;
-        }
-        this.options.styles[index] = meta.cartocss;
-        if (this.options.styles.indexOf(undefined) === -1) {
-          this.setUrl(this.model.get('urls').tiles[0]);
-        }
+      var index = layerModel.layers.indexOf(child);
+      this.options.styles[index] = meta.cartocss;
+      if (this.options.styles.indexOf(undefined) === -1) {
+        this.setUrl(this.model.get('urls').tiles[0]);
       }
     }, this);
 
@@ -71,7 +66,6 @@ var LeafletCartoDBVectorLayerGroupView = CartoDBd3Layer.extend({
   _onTileJSONChanged: function () {
     var tilejson = this.model.get('urls');
     this.options.styles = this.model.layers.pluck('cartocss');
-    this.setUrl(tilejson.tiles[0]);
   },
 
   onAdd: function (map) {
--- a/src/geo/leaflet/leaflet-cartodb-vector-layer-group-view.js
+++ b/src/geo/leaflet/leaflet-cartodb-vector-layer-group-view.js
@@ -31,7 +31,7 @@ var LeafletCartoDBVectorLayerGroupView = CartoDBd3Layer.extend({
       if (this.model.get('type') === 'namedmap') {
         var index = child.collection.indexOf(child);
         this.options.styles[index - 1] = meta.cartocss;
-        if (this.options.styles.filter(function (p) { return typeof p === 'undefined'; }).length === 0) {
+        if (this.options.styles.indexOf(undefined) === -1) {
           this.setUrl(this.model.get('urls').tiles[0]);
         }
       }
--- a/src/geo/leaflet/leaflet-cartodb-vector-layer-group-view.js
+++ b/src/geo/leaflet/leaflet-cartodb-vector-layer-group-view.js
@@ -30,7 +30,10 @@ var LeafletCartoDBVectorLayerGroupView = CartoDBd3Layer.extend({
     layerModel.layers.bind('change:meta', function (child, meta) {
       if (this.model.get('type') === 'namedmap') {
         var index = child.collection.indexOf(child);
-        this.options.styles[index - 1] = meta.cartocss;
+        if (child.collection.at(0).get("type") === 'tiled') {
+          index ++
+        }
+        this.options.styles[index] = meta.cartocss;
         if (this.options.styles.indexOf(undefined) === -1) {
           this.setUrl(this.model.get('urls').tiles[0]);
         }
--- a/src/geo/leaflet/leaflet-cartodb-vector-layer-group-view.js
+++ b/src/geo/leaflet/leaflet-cartodb-vector-layer-group-view.js
@@ -30,9 +30,6 @@ var LeafletCartoDBVectorLayerGroupView = CartoDBd3Layer.extend({
     layerModel.layers.bind('change:meta', function (child, meta) {
       var index = layerModel.layers.indexOf(child);
       this.options.styles[index] = meta.cartocss;
-      if (this.options.styles.indexOf(undefined) === -1) {
-        this.setUrl(this.model.get('urls').tiles[0]);
-      }
     }, this);
 
     layerModel.layers.each(function (layer) {
@@ -66,6 +63,7 @@ var LeafletCartoDBVectorLayerGroupView = CartoDBd3Layer.extend({
   _onTileJSONChanged: function () {
     var tilejson = this.model.get('urls');
     this.options.styles = this.model.layers.pluck('cartocss');
+    this.setUrl(this.model.get('urls').tiles[0]);
   },
 
   onAdd: function (map) {
--- a/src/geo/leaflet/leaflet-cartodb-vector-layer-group-view.js
+++ b/src/geo/leaflet/leaflet-cartodb-vector-layer-group-view.js
@@ -63,7 +63,7 @@ var LeafletCartoDBVectorLayerGroupView = CartoDBd3Layer.extend({
   _onTileJSONChanged: function () {
     var tilejson = this.model.get('urls');
     this.options.styles = this.model.layers.pluck('cartocss');
-    this.setUrl(this.model.get('urls').tiles[0]);
+    this.setUrl(tilejson.tiles[0]);
   },
 
   onAdd: function (map) {
--- a/src/geo/leaflet/leaflet-cartodb-vector-layer-group-view.js
+++ b/src/geo/leaflet/leaflet-cartodb-vector-layer-group-view.js
@@ -30,8 +30,8 @@ var LeafletCartoDBVectorLayerGroupView = CartoDBd3Layer.extend({
     layerModel.layers.bind('change:meta', function (child, meta) {
       if (this.model.get('type') === 'namedmap') {
         var index = child.collection.indexOf(child);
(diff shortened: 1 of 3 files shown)


fdansv:
@alonsogarciapablo

src/windshaft/windshaft-map.js
```diff
@@ -71,10 +71,12 @@ var WindshaftMap = Backbone.Model.extend({
 
         // TODO: Revisit this (will layerIndex work for NamedMaps??)
         // Should we move it somewhere else?
-        _.each(layers, function (layer, layerIndex) {
+        _.each(options.layers, function (layer, layerIndex) {
```
alonsogarciapablo: Why `options.layers` instead of `layers`?

src/windshaft/windshaft-map.js
```diff
@@ -71,10 +71,12 @@ var WindshaftMap = Backbone.Model.extend({
...
           if (layer.get('type') === 'torque') {
             layer.set('meta', this.getLayerMeta(layerIndex));
             layer.set('urls', this.getTiles('torque'));
+          } else if (layer.get('type') === 'CartoDB' || layer.get('type') === 'mapnik') {
```
alonsogarciapablo: In the context of CartoDB.js, CartoDB layers should have the `CartoDB` type so I don't understand why we are checking for `mapnik`...

src/geo/leaflet/leaflet-cartodb-vector-layer-group-view.js
```diff
@@ -27,6 +27,17 @@ var LeafletCartoDBVectorLayerGroupView = CartoDBd3Layer.extend({
       this.setCartoCSS(index, style);
     }, this);
 
+    layerModel.layers.bind('change:meta', function (child, meta) {
+      if (this.model.get('type') === 'namedmap') {
+        var index;
+        child.collection.models.filter(function (c, i) { if (child.cid === c.cid) index = i; });
```
alonsogarciapablo: I think you could use Backbone's `collection.indexOf(model)` here.

alonsogarciapablo:
I'm missing some tests in this PR.

src/windshaft/windshaft-map.js
fdansv: Layers filters out some types of layer, so the indices are tampered.

src/geo/leaflet/leaflet-cartodb-vector-layer-group-view.js
fdansv: that makes so much more sense

src/geo/leaflet/leaflet-cartodb-vector-layer-group-view.js
```diff
@@ -27,6 +27,17 @@ var LeafletCartoDBVectorLayerGroupView = CartoDBd3Layer.extend({
...
+        this.options.styles[index - 1] = meta.cartocss;
+        if (this.options.styles.filter(function (p) { return typeof p === 'undefined'; }).length === 0) {
```
alonsogarciapablo: This conditional is really hard to understand for me. We could extract one or two variables with good names to make it a bit easier. Or even extract a method with that check:

```
// ...
if (this._somethingIsTrue()) {
  // ...
}
// ...

File: test/spec/geo/leaflet/leaflet-cartodb-vector-layer-group-view.spec.js [hunk 5]
```diff
@@ -75,4 +76,28 @@ describe('src/geo/leaflet/leaflet-cartodb-vector-layer-group-view.js', function
...
+    this.layerGroupModel.layers = new Backbone.Collection([
+      cartoDBLayer1,
+      cartoDBLayer2
+    ]);
+    var view = new LeafletCartoDBVectorLayerGroupView(this.layerGroupModel, this.leafletMap); // eslint-disable-line
+    this.layerGroupModel.set('type', 'namedmap');
+    this.layerGroupModel.set('urls', {tiles: []});
+    cartoDBLayer1.set('meta', {cartocss: 'whatever'});
+    _.defer(function () {
+      expect(L.CartoDBd3Layer.prototype.setUrl).not.toHaveBeenCalled();
```
alonsogarciapablo: Would this work too?

```
var view = new LeafletCartoDBVectorLayerGroupView(this.layerGroupModel, this.leafletMap); // eslint-disable-line
spyOn(view, 'setURL');
// ...
expect(view.setURL).toHaveBeenCalled();
```

alonsogarciapablo:
From what I understand it seems that `view.setURL` is going to be called twice with this change (and tiles are going to be fetched twice):

1. when the new instance of the map has been created.
2. when the all the (sub)layers have the new metadata.

Couldn't we avoid that somehow?

src/geo/leaflet/leaflet-cartodb-vector-layer-group-view.js
```diff
@@ -27,6 +27,19 @@ var LeafletCartoDBVectorLayerGroupView = CartoDBd3Layer.extend({
...
     }, this);
 
+    layerModel.layers.bind('change:meta', function (child, meta) {
+      if (this.model.get('type') === 'namedmap') {
+        var index = child.collection.indexOf(child);
+        if (child.collection.at(0).get('type') === 'tiled') {
+          index++;
+        }
+        this.options.styles[index] = meta.cartocss;
+        if (this.options.styles.indexOf(undefined) === -1) {
```
alonsogarciapablo: Will this work if a second instance of the map is created (all styles in this.options.styles will have been set after the previous map instance has been created)?

src/geo/leaflet/leaflet-cartodb-vector-layer-group-view.js
```diff
@@ -27,6 +27,19 @@ var LeafletCartoDBVectorLayerGroupView = CartoDBd3Layer.extend({
       this.setCartoCSS(index, style);
     }, this);
 
+    layerModel.layers.bind('change:meta', function (child, meta) {
+      if (this.model.get('type') === 'namedmap') {
+        var index = child.collection.indexOf(child);
+        if (child.collection.at(0).get('type') === 'tiled') {
```
alonsogarciapablo: I'm not sure where `child.collection` comes from (I don't see it in the Backbone docs), but we could use `layerModel.layers` instead, which surely doesn't have any `tiled` layer.

src/geo/leaflet/leaflet-cartodb-vector-layer-group-view.js
```diff
@@ -27,6 +27,19 @@ var LeafletCartoDBVectorLayerGroupView = CartoDBd3Layer.extend({
       this.setCartoCSS(index, style);
     }, this);
 
+    layerModel.layers.bind('change:meta', function (child, meta) {
+      if (this.model.get('type') === 'namedmap') {
```
alonsogarciapablo: Why only for named maps? It would be nice if we didn't have to check this and both named and anonymous maps worked the same way (i.e: set the URL to the provider once all styles have been (re)seted?)

test/spec/geo/leaflet/leaflet-cartodb-vector-layer-group-view.spec.js
```diff
@@ -75,4 +76,22 @@ describe('src/geo/leaflet/leaflet-cartodb-vector-layer-group-view.js', function
       done();
     });
   });
+  it('should call setUrl when all named map styles have been added', function () {
```
alonsogarciapablo: The description for this test case is no longer right.

test/spec/geo/leaflet/leaflet-cartodb-vector-layer-group-view.spec.js
```diff
@@ -75,4 +76,22 @@ describe('src/geo/leaflet/leaflet-cartodb-vector-layer-group-view.js', function
...
+    var cartoDBLayer1 = new CartoDBLayer();
+    var cartoDBLayer2 = new CartoDBLayer();
+    L.CartoDBd3Layer.prototype.setUrl = jasmine.createSpy();
```
alonsogarciapablo: We're not using this anymore, so we can :fire: it.

test/spec/geo/leaflet/leaflet-cartodb-vector-layer-group-view.spec.js
```diff
@@ -75,4 +76,22 @@ describe('src/geo/leaflet/leaflet-cartodb-vector-layer-group-view.js', function
...
+    this.layerGroupModel.layers = new Backbone.Collection([
+      cartoDBLayer1,
+      cartoDBLayer2
+    ]);
+    var view = new LeafletCartoDBVectorLayerGroupView(this.layerGroupModel, this.leafletMap); // eslint-disable-line
+    this.layerGroupModel.set('type', 'namedmap');
+    this.layerGroupModel.set('urls', {tiles: []});
+    cartoDBLayer1.set('meta', {cartocss: 'whatever'});
+    cartoDBLayer2.set('meta', {cartocss: 'whatever2'});
+    expect(view.options.styles.indexOf(undefined)).toEqual(-1);
```
alonsogarciapablo: Why not assert that:

```
expect(view.options.styles[0]).toEqual('wathever');
expect(view.options.styles[1]).toEqual('wathever2');
```

to be more specific?

test/spec/geo/leaflet/leaflet-cartodb-vector-layer-group-view.spec.js
```diff
@@ -75,4 +76,22 @@ describe('src/geo/leaflet/leaflet-cartodb-vector-layer-group-view.js', function
...
+    L.CartoDBd3Layer.prototype.setUrl = jasmine.createSpy();
+    LeafletCartoDBVectorLayerGroupView.prototype._onTileJSONChanged = function () {
+      this.options.styles = [undefined, undefined];
+    };
+    this.layerGroupModel.layers = new Backbone.Collection([
+      cartoDBLayer1,
+      cartoDBLayer2
+    ]);
+    var view = new LeafletCartoDBVectorLayerGroupView(this.layerGroupModel, this.leafletMap); // eslint-disable-line
+    this.layerGroupModel.set('type', 'namedmap');
```
alonsogarciapablo: I don't think we need this anymore, so we can probably :fire: it.

test/spec/geo/leaflet/leaflet-cartodb-vector-layer-group-view.spec.js
alonsogarciapablo: And we can remove `var L = window.L` if we remove this, right?

If we have to spy on `setURL` I would do it on the view:

```
spyOn(view, setURL).and...
```

src/windshaft/windshaft-map.js
xavijam: this change provokes that torque layers doesn't receive the proper meta and torque widget is not working at all.

xavijam:
@alonsogarciapablo @fdansv what happened here? There were comments after the merge.

fdansv:
o/ As discussed with Pablo the other day, I merged prematurely, I'm sorry :(

Shall we revert the merge?

xavijam:
No, it is not necessary.

PR #1178, opened 2016-03-04
Docs 716 other important stuff reorg

csobier:
For [Docs Issue#716](https://github.com/CartoDB/docs/issues/716). @alonsogarciapablo , are you okay with this reshuffle? I think it makes more sense from a usability perspective.


xavijam:
:+1:

PR #1180, opened 2016-03-07
Checking attributions

xavijam:
cc @javisantana @alonsogarciapablo

### DIFF
--- a/src/geo/leaflet/leaflet_cartodb_layergroup.js
+++ b/src/geo/leaflet/leaflet_cartodb_layergroup.js
@@ -27,7 +27,7 @@ L.CartoDBGroupLayerBase = L.TileLayer.extend({
     sql_api_domain:     "cartodb.com",
     sql_api_port:       "80",
     sql_api_protocol:   "http",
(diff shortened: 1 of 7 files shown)


xavijam:
Good thing is we have those problems already fixed in v4.

src/geo/leaflet/leaflet_cartodb_layergroup.js
```diff
@@ -142,6 +142,9 @@ L.CartoDBGroupLayerBase = L.TileLayer.extend({
     });
   },
 
+  getAttribution: function() {
+    return cdb.core.sanitize.html(this.options.attribution);
```
javisantana: should we santize the attribution in this case? I think we should only do this if this was a view method

xavijam: Yeah, probably this is the most important addition, because leaflet adds the attribution when the layer is added ([in the `onAdd` method](https://github.com/CartoDB/cartodb.js/blob/develop/vendor/leaflet.js#L8131)), and it uses this method `getAttribution`.

xavijam:
Tested in staging \0/

PR #1181, opened 2016-03-08
Changing way to get all the data (non filtered) from the histogram widget

xavijam:
Histogram widget needs to know all the data from the beginning and it was requested within the proper histogram dataview model, using a parameter called `submitBBox` in order to know if the data should return filtered or not.

Bad thing is vector rendering doesn't have any property to know all the data out of the bounds, so that described technique doesn't make the trick. Turns out category widget needs also all the data from the beginning, in order to know the amount of categories, and it solved that problem creating another model (called `rangeModel`) which request all the data independently.

I've tried the same thing with the histogram dataview, a different model requesting histogram data without any filter.

How do you like it @alonsogarciapablo?

cc @fdansv @javisantana and @javierarce.

It doesn't fix 1179, but it is part of it, Deep-insights PR is coming too.


src/geo/data-providers/geojson/histogram-geojson-data-provider.js
```diff
@@ -27,7 +27,7 @@ HistogramGeoJSONDataProvider.prototype.getData = function () {
     start = typeof options.start === 'number' ? options.start : filter.getMin(columnName);
     width = (end - start) / options.bins;
     values = filter.getValues(false, columnName);
-    bins = d3.layout.histogram().bins(numberOfBins)(values.map(function (f) {
+    bins = d3.layout.histogram().range([start, end]).bins(numberOfBins)(values.map(function (f) {
```
xavijam: We weren't applied the necessary range in order to calculate properly the buckets.
Boy scout fixing!

src/dataviews/histogram-dataview/histogram-data-model.js
```diff
@@ -0,0 +1,70 @@
...
+    url: '',
+    data: []
+  },
+
+  url: function () {
+    return this.get('url') + '?' + 'bins=' + this.get('bins');
+  },
+
+  initialize: function () {
+    this.bind('change:url change:bins', function () {
```
alonsogarciapablo: Do we really need the `change:bins` binding? I don't think it's being set externally.

src/dataviews/histogram-dataview/histogram-data-model.js
```diff
@@ -0,0 +1,70 @@
...
+  },
+
+  setUrl: function (url) {
+    if (!url) {
+      throw new Error('url not specified');
+    }
+    this.set('url', url);
+  },
+
+  setBins: function (num) {
```
alonsogarciapablo: Is this going to be used externally?

src/dataviews/histogram-dataview-model.js
```diff
@@ -75,6 +85,10 @@ module.exports = DataviewModelBase.extend({
     return this._data.toJSON();
   },
 
+  getUnfilteredDataModel: function () {
```
alonsogarciapablo: Is this going to be used externally? (is not being used in this PR)

src/dataviews/histogram-dataview/histogram-data-model.js
alonsogarciapablo: When the histogram is zoomed, bins are updated but just for internal usage, I don't think we need re-fetch this model...

src/geo/data-providers/geojson/histogram-geojson-data-provider.js
alonsogarciapablo: cc: @fdansv

fdansv: wow NICE catch, I can't believe I didn't consider that. :clap:

test/spec/dataviews/histogram-dataview-model.spec.js
```diff
@@ -18,6 +20,47 @@ describe('dataviews/histogram-dataview-model', function () {
     });
   });
 
+  it('should not listen any url change from the beginning', function () {
+    expect(this.model.listenTo.calls.count()).toBe(5);
```
alonsogarciapablo: `expect(this.model.listenTo.calls.count()).toBe(5);` is not very descriptive IMO and it makes this test case not very laser focused :)

What about?
```
var callback = jasmine.createSpy('callback');
this.model.bind('change:url', callback);
this.model.set('url', 'http://example.com');
expect(callback).not.toHaveBeenCalled();
```

src/dataviews/histogram-dataview/histogram-data-model.js
xavijam: Yes it is possible, if user updates dataview bins, this "unfilteredData" model should update its bins and get the data again.

src/dataviews/histogram-dataview/histogram-data-model.js
xavijam: It is going to be used externally, yes, from histogram content-view (Deep-insights).

alonsogarciapablo: Forget about this.

src/dataviews/histogram-dataview/histogram-data-model.js
alonsogarciapablo: :+1:

alonsogarciapablo:
Left some minor comments, but it looks :+1: overall!

Thanks @xavijam! :clap:

test/spec/dataviews/histogram-dataview-model.spec.js
xavijam: Much better! :)

xavijam: Wait, we can't do that, with that test the callback will always be triggered.

test/spec/dataviews/histogram-dataview-model.spec.js
```diff
@@ -18,6 +20,47 @@ describe('dataviews/histogram-dataview-model', function () {
     });
   });
 
+  it('should not listen any url change from the beginning', function () {
+    this.model.set('url', 'http://cartodb.com');
```
xavijam: What about this one @alonsogarciapablo?
If there is a url change, a fetch should happen (for rest of dataview models).

src/dataviews/histogram-dataview-model.js
```diff
@@ -45,24 +43,36 @@ module.exports = DataviewModelBase.extend({
...
+    this._unfilteredData.bind('change:data', function (mdl, data) {
+      this.set({
+        start: mdl.get('start'),
+        end: mdl.get('end'),
+        bins: mdl.get('bins')
+      }, { silent: true });
+      this._onChangeBinds();
+    }, this);
+
+    this.once('change:url', function () {
```
alonsogarciapablo: Won't this be unset by `this.stopListening(this, 'change:url', null);`?

src/dataviews/histogram-dataview-model.js
xavijam: Yes, it is gonna be used in Deep-insights.

src/dataviews/histogram-dataview-model.js
xavijam: No, `_initBinds` method happens (line 43) before this new "binding", line 43 with the DataviewModelBase initialize.

test/spec/dataviews/histogram-dataview-model.spec.js
alonsogarciapablo: Ok

src/dataviews/histogram-dataview-model.js
alonsogarciapablo: :+1:

xavijam:
Waiting to merge until Deep-insights histogram code is ready, stay tuned.

src/geo/data-providers/geojson/histogram-geojson-data-provider.js
```diff
@@ -18,16 +18,16 @@ HistogramGeoJSONDataProvider.prototype.getData = function () {
     end = filter.getMax(columnName);
     start = filter.getMin(columnName);
     values = filter.getValues();
-    bins = d3.layout.histogram().bins(numberOfBins)(values.map(function (f) {
+    bins = d3.layout.histogram().range([start, end])(values.map(function (f) {
       return f.properties[options.column];
     }));
-    width = (end - start) / options.data.length;
+    width = (end - start) / bins.length;
```
xavijam: After some debugging, I feel we missed to set the range and remove the number of bins, because in the zoomed histogram the number of bins are calculated by the "provider", in this case D3 will calculate them by default with Sturges formula.

xavijam:
Merging, DI pull request almost ready to go, ok @alonsogarciapablo?

src/geo/data-providers/geojson/histogram-geojson-data-provider.js
alonsogarciapablo: Not very familiar with this part of the code. Probably @fdansv can say something here.

xavijam: No no, this has to be like this, it was not working properly before :(.

fdansv: Yeah that makes a lot more sense :ship: :it:

PR #1183, opened 2016-03-09
Do not add cartodb_id column in interactivity column if there is no infowindow or tooltip field names required

### LINKED ISSUE #1182: An interactivity object is always being passed in the layer configuration
https://team.cartodb.com/u/saleiva/editor/ff2dddfa-03bf-11e5-a4d8-0e43f3deba5a/embed?vector=true
vs
https://team.cartodb.com/u/saleiva/viz/ff2dddfa-03bf-11e5-a4d8-0e43f3deba5a/public_map

If you open both mapconfigs being sent to the tiler, you'll notice that in a few layers we are "automatically" adding an `interactivity` array only containing `cartodb_id`. The SQL query<sup>1</sup> being applied to that layer doesn't return a cartodb_id column, so the instance fails, as it's trying to get a column that doesn't exist.

This comes from https://github.com/CartoDB/d3.cartodb/issues/134

@xavijam @alonsogarciapablo 

1 - `SELECT ST_Transform(ST_Segmentize(ST_MakeLine(a.the_geom, b.the_geom)::geography, 100000)::geometry,3857) as the_geom_webmercator FROM saleiva.tuenti_employees_locations_2015 a, populated_places b WHERE b.name = 'Madrid'`

xavijam:
Fixes #1182 
CR: @alonsogarciapablo 

cc @javisantana @fdansv

### DIFF
--- a/src/geo/map/cartodb-layer.js
+++ b/src/geo/map/cartodb-layer.js
@@ -74,11 +74,14 @@ var CartoDBLayer = LayerModelBase.extend({
   },
 
   getInteractiveColumnNames: function() {
-    return _.uniq(
-      ['cartodb_id']
-        .concat(this.getInfowindowFieldNames())
-         .concat(this.getTooltipFieldNames())
+    var fieldNames = _.union(
+      this.getInfowindowFieldNames(),
+      this.getTooltipFieldNames()
     );
+    if (fieldNames.length) {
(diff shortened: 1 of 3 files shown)


alonsogarciapablo:
:+1: (after test suite is red - probably some lint errors)

fdansv:
Niiice!!!

PR #1185, opened 2016-03-10
Fixes issue with indexes when setting the metadata of layer models from Windshaft response

alonsogarciapablo:
We basically now ignore any possible `tiled` layer at the bottom of layers that comes from Windshaft response when getting the metadata for each of the local layer models.

@xavijam @fdansv PTAL, thanks!


xavijam:
This is music to my ears :) 

:shipit:

PR #1187, opened 2016-03-11
Prevents tilejson listener from invalidating cartocss

fdansv:
@alonsogarciapablo After you fixed the passing of metadata, the only thing that was broken was the tilejson listener reassigning styles by plucking the client styles. Since now we're relying on meta to always pass the styles, removing that style setter fixes the issue. Named maps are visible again in vector.


alonsogarciapablo:
:+1:

PR #1196, opened 2016-03-15
Points d3.cartodb to latest commit

fdansv:
Includes nice fixes such as excluding empty `GeometryCollection` from entering the filter and performing contains checks to lines and polygons.

@xavijam


xavijam:
Could you take the short "hash", we are using that everywhere, just to be consistent.

fdansv:
> Could you take the short "hash"?, we are using that everywhere, just to be consistent.

Oui

xavijam:
:heart:  :+1:

PR #1198, opened 2016-03-16
Extracts a VizJSON class with some methods to manipulate viz.json objects

alonsogarciapablo:
This PR completes the first task of #804.

These changes will allow us to easily extract loading and manipulation of the vizjson (based on `createVis`  options) to `createVis`, which is the next step in #804.

@viddo could you please take a look? Thanks!


viddo:
:+1:

PR #1199, opened 2016-03-17
Use the main branch instead of commit for dependency

viddo:
As discussed with @javisantana, will change the internal libs to not use commits by main branches for reference instead.

@fdansv can you confirm that the gh-pages is the correct “main” branch of d3.cartodb?

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "underscore": "1.8.3",
     "cartoassets": "CartoDB/CartoAssets#master",
     "d3": "3.5.8",
-    "d3.cartodb": "CartoDB/d3.cartodb#7816870"
+    "d3.cartodb": "CartoDB/d3.cartodb#gh-pages"
   },
   "devDependencies": {
     "browserify": "11.2.0",


fdansv:
That is correct :)

alonsogarciapablo:
Could you please explain the motivation behind this? Thx!

viddo:
@alonsogarciapablo the primary motivation is that an update wanted in the editor that requires a fix in an implicit library that we own (e.g. d3.cartodb) should not require a release in each library in between, which is the case with the current system of commits (which @fdansv had trouble with as I understood it).

By changing the dependency to be more laxed (point to the master branch instead of a specific commit) a npm install will always install latest versions available. I'll update DI and editor to follow the same pattern. The editor will continue to use the shrinkwrap however, since we actually deploy things to production there.

Once we publish the new v4 we should set the versions to point to allow minor versions, only require explicit release on intermediate libraries on mayor/non-backward compatible changes.

Does that answer your question?

alonsogarciapablo:
Thanks fro the explanation.

> The editor will continue to use the shrinkwrap however, since we actually deploy things to production there.

Not sure I understand that sentence. What do you mean?

viddo:
What I mean was that we're using a shrinkwrap since some time back (introduced due to [other reasons](https://github.com/CartoDB/cartodb/pull/6620)), so even with this change one will still need to update and release on the editor to get new versions there.

viddo:
Also, as I pointed out in the offline meeting yesterday (although I think you weren't present by then), this approach has the downside that any change that's would not be considered "patch" by semver standard, might cause cartodb.js and/or DI to break after a npm install, but @javisantana considered this to be preferable to having to do individual releases since the new libs code is still WIP and not yet published. 
On the editor we're a little more protected since the shrinkwrap allows to control what versions are used regardless of what's stated in the package.json.

alonsogarciapablo:
Thanks for the explanation @viddo. I agree that having to recursively update dependencies in all projects is a PITA given how many changes we're doing at the moment.

PR #1204, opened 2016-03-18
Fixes bug with indexes when generating the TileJSON for Named Maps.

alonsogarciapablo:
I spent some time to (hopefully) clarify the code a bit and added some more tests to (hopefully) avoid more issues with indexes & infowindows :)

@xavijam can you please CR? Thanks!

### DIFF
--- a/src/geo/cartodb-layer-group-anonymous-map.js
+++ b/src/geo/cartodb-layer-group-anonymous-map.js
@@ -5,10 +5,7 @@ var CartoDBLayerGroupAnonymousMap = CartoDBLayerGroupBase.extend({
     type: 'layergroup'
   },
 
-  // Returns the index of the CartoDB layer in relation to all layers when the map is
-  // an "Anonymous Map". For example, if there are two CartoDB layers and layer #0 is
-  // hidden, this method would return -1 for #0 and 0 for layer #1.
-  _getIndexOfVisibleLayer: function (layerIndex) {
+  _convertToWindshaftLayerIndex: function (layerIndex) {
     var layers = {};
     var i = 0;
     this.layers.each(function (layer, index) {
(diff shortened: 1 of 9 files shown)


src/geo/cartodb-layer-group-base.js
```diff
@@ -31,36 +31,51 @@ var CartoDBLayerGroupBase = Backbone.Model.extend({
...
+        tiles: urls.tiles,
+        formatter: function (options, data) { return data; }
+      };
     }
+  },
 
-    return {
-      tilejson: '2.0.0',
-      scheme: 'xyz',
-      grids: urls.grids[this._getIndexOfVisibleLayer(layerIndex)],
```
alonsogarciapablo: @xavijam: the problem was basically that this method was returning the wrong index. `urls.grids` is an array that contains one entry for each of the visible layers. In the case of Named Maps `this._getIndexOfVisibleLayer` was taking into account that Named Maps have a tiled layer at the "bottom" and the wrong grid was being included in the TileJSON. The new method `_convertToMapnikLayerIndex` only takes into account visible "mapnik" layers, which will (hopefully) match the index for the given `layerIndex` in `url.grids`.

src/geo/cartodb-layer-group-anonymous-map.js
```diff
@@ -5,10 +5,7 @@ var CartoDBLayerGroupAnonymousMap = CartoDBLayerGroupBase.extend({
     type: 'layergroup'
   },
 
-  // Returns the index of the CartoDB layer in relation to all layers when the map is
-  // an "Anonymous Map". For example, if there are two CartoDB layers and layer #0 is
-  // hidden, this method would return -1 for #0 and 0 for layer #1.
-  _getIndexOfVisibleLayer: function (layerIndex) {
+  _convertToWindshaftLayerIndex: function (layerIndex) {
```
xavijam: key function!

xavijam:
:clap: :clap:

xavijam:
Go ahead! Infowindows will never break again :smile_cat:

PR #1211, opened 2016-03-23
Extract vizjson loading to create-vis function

alonsogarciapablo:
This PR completes the second task of #804.

The create-vis function now fetches the viz.json file from a URL when given one.

@viddo could you please CR? Thanks!

### DIFF
--- a/src/api/create-vis.js
+++ b/src/api/create-vis.js
@@ -1,34 +1,70 @@
 var _ = require('underscore');
 var Vis = require('../vis/vis');
+var Loader = require('../core/loader');
+var Promise = require('./promise');
 
-var createVis = function(el, vizjson, options, callback) {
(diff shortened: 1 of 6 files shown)


src/api/create-vis.js
```diff
@@ -1,34 +1,79 @@
...
+  var fn = args[args.length - 1];
 
   if (_.isFunction(fn)) {
     callback = fn;
   }
+  if (typeof el === 'string') {
+    el = document.getElementById(el);
+  }
+
+  var DEFAULT_OPTIONS = {
```
viddo: Considering this is a intended to be constant, why not move it outside the function?

viddo:
:+1:

src/api/create-vis.js
alonsogarciapablo: :+1:

PR #1219, opened 2016-03-29
Change SQL API to support node-like callbacks

viddo:
I.e. 1st argument is always an error, and 2nd is the successful data.

This enabled to know when a `describeXXX` method fails, e.g.

```js
sql.describeGeom('SELECT * FROM foobar', 'the_geom', function(err, data) {
  if (err) throw err;

  console.log(data) // => { simplified_geometry_type: "line", … 
})
```

@xavijam can you review?
cc @alonsogarciapablo


xavijam:
:+1:

PR #1220, opened 2016-03-30
Define required options for a view (proposal)

viddo:
Right now we're doing a lot of these:

``` js
// some-view.js
module.exports = cdb.core.View.extend({
  initialize: function(options) {
    if (!options.someObj) throw new Error('someObj is required');
  }
  //…
}
```

I propose we leverage the existing options object to define required vars as such:

``` js
module.exports = cdb.core.View.extend({
  options: {
    someObj: cdb.core.View.requires(Object)
  },
  //…
```

One can access the options on `this.options.someObj`, but for more terse code we could set it up that the defined vars on options gets set as `this._someObj` ofc, really a matter of how this would be implemented. See the spec for more details.

Similar to [prop types in React components](https://facebook.github.io/react/docs/reusable-components.html#prop-validation)

@xavijam @javierarce opinions/thoughts?

### DIFF
--- a/test/spec/core/view.spec.js
+++ b/test/spec/core/view.spec.js
@@ -3,6 +3,7 @@ var _ = require('underscore');
 var Backbone = require('backbone');
 var templates = require('cdb.templates');
 var View = require('../../../src/core/view');
+var Model = require('../../../src/core/model');
 
 describe('core/view', function() {
   var TestView;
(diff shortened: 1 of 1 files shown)


xavijam:
I like it man! :clap:

xavijam:
Could we do it also for Models and Collections?

javierarce:
:+1:

viddo:
> Could we do it also for Models and Collections?

I suppose

xavijam:
Let's start using this one, then we will migrate these changes to model and collection. Sounds good?

viddo:
You mean that you'll implement it?

xavijam:
Yep, no problem, open a ticket and assign it to me ;).

viddo:
Why not just continue on this PR/branch?

On Wed, Mar 30, 2016 at 2:51 PM, Javier Álvarez Medina <
notifications@github.com> wrote:

> Yep, no problem, open a ticket and assign it to me ;).
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/CartoDB/cartodb.js/pull/1220#issuecomment-203417970>
>



-- 

*Nicklas Gummesson*, software engineer
*CartoDB* — Map Your World's Data

Live sessions to learn CartoDB <http://cartodb.com/webinars/>

alonsogarciapablo:
I like the idea :+1:.

Not sure if `options` is the best keyword given that we're already (over)using it a lot internally and we're actually defining things that are required (i.e: not optional)... Perhaps we can find a better word (i.e: more like `dependencies`?)

Also, I guess we could hide the `View.requires`, right?

```
module.exports = cdb.core.View.extend({
  requiredDeps: {
    someObj: Object
  },
  //…
});
```

nobuti:

Cartofante:

PR #1221, opened 2016-04-01
[WIP] changed public api

javisantana:
Change visualization public API to be used from DI.js

### DIFF
--- a/src/vis/vis.js
+++ b/src/vis/vis.js
@@ -63,7 +63,7 @@ var Vis = View.extend({
     var layers = [];
 
     // flatten layers (except baselayer)
-    var layers = _.map(this.getLayers().slice(1), function (layer) {
+    var layers = _.map(this.getLayerViews().slice(1), function (layer) {
       if (layer.getSubLayers) {
         return layer.getSubLayers();
       }
--- a/src/vis/vis.js
+++ b/src/vis/vis.js
@@ -624,18 +624,34 @@ var Vis = View.extend({
   // public methods
   //
 
-  // get the native map used behind the scenes
+  /**
+   * @return the native map used behind the scenes {L.Map} or {google.maps.Map}
+   */
   getNativeMap: function () {
     return this.mapView.getNativeMap();
   },
 
   // returns an array of layers
-  // TODO: Rename to getLayerViews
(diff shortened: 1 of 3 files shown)


src/vis/vis.js
```diff
@@ -624,20 +624,36 @@ var Vis = View.extend({
...
+  /**
+   * @return the native map used behind the scenes {L.Map} or {google.maps.Map}
+   */
   getNativeMap: function () {
     return this.mapView.getNativeMap();
   },
 
   // returns an array of layers
-  // TODO: Rename to getLayerViews
-  getLayers: function () {
```
alonsogarciapablo: This method is being used [by the _setLayerOptions method](https://github.com/CartoDB/cartodb.js/blob/api/src/vis/vis.js#L66), which is not being used. I think we should remove `setLayerOptions` since no tests fail for that reason.

That method is also being used [here](https://github.com/CartoDB/cartodb.js/blob/v4/src/vis/overlays.js#L297), [here](https://github.com/CartoDB/cartodb.js/blob/v4/src/vis/overlays.js#L300), and [here](https://github.com/CartoDB/cartodb.js/blob/v4/src/vis/overlays.js#L308). We can just rename these usages to getLayerViews for now.

javisantana: Ok, will do, thanks

javisantana:
done @alonsogarciapablo

src/vis/vis.js
```diff
@@ -624,20 +624,36 @@ var Vis = View.extend({
...
     return _.compact(this.map.layers.map(function (layer) {
       return self.mapView.getLayerViewByLayerCid(layer.cid);
     }));
   },
 
+  /**
+   * @return Array of {LayerModel}
+   */
+  getLayers: function() {
+    return _.clone(this.map.layers.models);
```
alonsogarciapablo: Out of curiosity... whey are we using `_.clone` here?

alonsogarciapablo:
:es: (although you might want to add some very basic tests for those new methods given that they are public, just to ensure that a `vis` object responds to those).

javisantana:
yep, adding them

alonsogarciapablo:
:es:

PR #1223, opened 2016-04-05
cartodblayer.update expects an ID as the source attribute

alonsogarciapablo:
It expected an analysis node before.

I started to override the `update` method for CartoDB layers to check if the given source attribute is a valid analysis node, but it required CartoDB layers to know about the collection analysis and that required several changes. This would probably be easier if we had some API objects, so I'll think about it later.

@viddo please take a look. Thanks!


viddo:
Looks legit :+1:

PR #1224, opened 2016-04-05
Updating the "sources" of an analysis refreshes the map

alonsogarciapablo:
Also, this PR removes the method to update individual analysis nodes since it was not currently being used anywhere.

@viddo


viddo:
:shipit:

PR #1228, opened 2016-04-07
Leaflet needs to be required before wax

alonsogarciapablo:
For some reason, in the new editor, wax was being included before Leaflet in the `editor3.js` file. This was causing a JS error (when trying to use `L.TileLayer`, `L` was undefined). This little change fixes that. Perhaps, there's a better way to ensure Leaflet is required before wax?

@xavijam @javierarce?


xavijam:
:+1:

PR #1230, opened 2016-04-12
Reload map when type of analyses change

alonsogarciapablo:
@xavijam

### DIFF
--- a/src/analysis/analysis-model.js
+++ b/src/analysis/analysis-model.js
@@ -15,7 +15,15 @@ module.exports = Model.extend({
 
     this._camshaftReference = opts.camshaftReference;
     this._map = opts.map;
-    this.bind('change:type', this._reloadMap, this);
+    this.bind('change:type', function () {
+      this.unbind(null, null, this);
+      this._initBinds();
+      this._reloadMap();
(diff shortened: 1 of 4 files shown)


xavijam:
:dancer: :dancers:  thanks!

alonsogarciapablo:
Anyway, if you're changing the `type` of an analysis, I asume you're also changing some of the params of that analysis, so [this already existing binding](https://github.com/CartoDB/cartodb.js/pull/1230/files#diff-c5288cba30a7a70973afdc0e3816deb5R20) should cause the map to be reloaded...

xavijam:
So I realized that, If we change the type of the analysis, this solution doesn't work because new type params are not binded, so we should re-bind new type params when it has changed.

alonsogarciapablo:
:+1: (after adding a new test for the last commit)

alonsogarciapablo:
:+1:

alonsogarciapablo:
retest this please

alonsogarciapablo:
retest this please

xavijam:
retest this please

xavijam:
:clap:

PR #1231, opened 2016-04-13
Keep listening type changes when it was changed once

xavijam:
So, silly me, I was unbinding previous listeners included type, so if type had changed twice, second one won't listen any change :(.

CR: @alonsogarciapablo


alonsogarciapablo:
:+1:

PR #1232, opened 2016-04-13
Auto style

fdansv:
cc @javisantana


javisantana:
@alonsogarciapablo should be the one to review this.

I love that image but it would look much better with an explanation of the solution :tmcw:

fdansv:
Sure:
- Dataviews generate a custom CartoCSS for its corresponding layer that is saved in the model as tempStyle.
- Hitting auto-style triggers the map to re-instantiate, which is when `DataviewsCollection` gathers any temporary styles and includes them in the mapconfig.
- If it's a named map, it will use the `styles` query parameter that was added recently to the API. Otherwise, it will just overwrite the `cartocss` property in the corresponding layer.
- To restore the style, on named maps we just don't pass a `styles` parameter, and for anonymous maps we store the initial state of the style, which we go back to.

javisantana:
some tests would be welcomed as well.

In the other hand, doing:

```
map.layers.at(2).set('cartocss', '....')
```

would work if it's a named map?

src/dataviews/dataview-model-base.js
```diff
@@ -96,6 +96,14 @@ module.exports = Model.extend({
     }
   },
 
+  _onStyleChanged: function (style) {
```
alonsogarciapablo: It doesn't look we're using this method... are we?

src/dataviews/dataview-model-base.js
```diff
@@ -96,6 +96,14 @@ module.exports = Model.extend({
...
+    if (this._dataProvider) {
+      this._dataProvider.applyFilter(filter);
```
alonsogarciapablo: `filter` is not defined in this method...

src/windshaft/layergroup-config.js
```diff
@@ -7,13 +7,14 @@ LayerGroupConfig.generate = function (options) {
...
-  _.each(layers, function (layer) {
+  _.each(layers, function (layer, index) {
+    var provisionalCartoCSS = options.styles[index]
     if (layer.isVisible()) {
       var layerConfig = {
         type: layer.get('type').toLowerCase(),
         options: {
           sql: layer.get('sql'),
-          cartocss: layer.get('cartocss'),
+          cartocss: provisionalCartoCSS || layer.get('initialStyle') || layer.get('cartocss'),
```
alonsogarciapablo: will `initialStyle` ever be different than `cartocss`? do we really need it?

fdansv: AFAIK `cartocss` is the style that the layer has currently. `initialStyle` is the style the layer is initialised with.

src/dataviews/dataview-model-base.js
fdansv: Yes, sorry, I left that method there by accident

alonsogarciapablo:
Some general comments:

- Why are temporary styles set on the dataview models instead of setting them on the layer model directly? Right now `cartocss` is an attribute of the layer, so IMHO it makes sense to set that  `temporary_cartocss` (or however we wanna call it) on layer models.
- `src/windshaft/namedmap-config.js` and `src/windshaft/layergroup-config.js` classes know about the layers that need to be serialized. I we set temp styles on layer models (as suggested above), those classes could know how to include those styles in the Map Config (we could get rid of [this](https://github.com/CartoDB/cartodb.js/pull/1232/files#diff-e97f11285d65b93bffa13d114e6fa71cR23) - IMO that shouldn't be a responsibility of the collection of dataviews).
- I'm missing some tests here... These are important to better understand the PR, but also to ensure this won't break in the future (a lot of this stuff has changed in the `analysis` branch so we should make sure we don't forget this... and tests are a good reminder)
- Is there a corresponding PR on deep-insights.js? Would be nice to review both PRs together :)

Thanks!

src/windshaft/layergroup-config.js
alonsogarciapablo: I can see that, but I'm not sure why/when would we want to use the initial style of the layer... For instance, if `cartocss` is changed... why should `initialStyle` take precedence over `cartocss`?

fdansv: Yes, you're right, this should be handled differently

fdansv:
> Why are temporary styles set on the dataview models instead of setting them on the layer model directly? Right now cartocss is an attribute of the layer, so IMHO it makes sense to set that temporary_cartocss (or however we wanna call it) on layer models.

I rectified this and now we just don't use temporary styles, we rather store a copy of the original style in the layer model itself, which is referenced later in `restoreCartoCSS`

> src/windshaft/namedmap-config.js and src/windshaft/layergroup-config.js classes know about the layers that need to be serialized. I we set temp styles on layer models (as suggested above), those classes could know how to include those styles in the Map Config (we could get rid of this - IMO that shouldn't be a responsibility of the collection of dataviews).

Styles are now set with the layers directly, which also allows us to change a named map's cartocss without using widgets.

src/geo/map/cartodb-layer.js
```diff
@@ -31,6 +45,17 @@ var CartoDBLayer = LayerModelBase.extend({
     return this.get('visible');
   },
 
+  getGeometryType: function () {
```
javisantana: this method basically works because of the cartodb wizards work but you can't guess geometry type from style

src/dataviews/dataviews-collection.js
```diff
@@ -7,14 +7,26 @@ module.exports = Backbone.Collection.extend({
...
     // If a histogram model applies the histogram sizes, rest should remove/disable
     // the sizes applied before.
-    this.bind('change:histogram_sizes', function (m, isSizesApplied) {
-      if (isSizesApplied) {
+    this.bind('change:autoStyle', function (m, autoStyle) {
+      if (autoStyle) {
         this.each(function (mdl) {
-          if (mdl !== m && mdl.get('histogram_sizes')) {
-            mdl.set('histogram_sizes', false);
+          if (mdl !== m && mdl.layer.get('layer_name') === m.layer.get('layer_name') && mdl.get('autoStyle')) {
```
javisantana: I don't see why cartodb.js needs to know if there is some autostyling, that is managed at different level, cartodb.js only knows (or should know) about cartocss for each layer

src/geo/leaflet/leaflet-cartodb-vector-layer-group-view.js
```diff
@@ -24,7 +24,7 @@ var LeafletCartoDBVectorLayerGroupView = CartoDBd3Layer.extend({
 
     layerModel.layers.bind('change:cartocss', function (child, style) {
       var index = child.get('order') - 1;
-      this.setCartoCSS(index, style);
+      this.setCartoCSS(index, style, true);
```
javisantana: I see this change but not a change in the setCartoCSS signature

src/dataviews/dataview-model-base.js
alonsogarciapablo: Are we using this?

src/geo/map/cartodb-layer.js
```diff
@@ -14,7 +14,17 @@ var CartoDBLayer = LayerModelBase.extend({
     options = options || {};
 
     this._map = options.map;
-    this.bind('change:visible change:sql change:cartocss change:source', this._reloadMap, this);
+    if (attrs && attrs.cartocss) {
+      this.set('initialStyle', attrs.cartocss);
+    }
+    this.bind('change:visible change:sql change:source', this._reloadMap, this);
+    this.bind('change:cartocss', this._setCartoCSS);
```
alonsogarciapablo: It looks like `_setCartoCSS` is basically doing something when cartocss has changed. I'd call it `onCartoCSSChanged` or someting along those lines...

src/dataviews/histogram-dataview-model.js
```diff
@@ -141,9 +141,9 @@ module.exports = DataviewModelBase.extend({
 
   _onChangeBinds: function () {
     DataviewModelBase.prototype._onChangeBinds.call(this);
-    this.bind('change:histogram_sizes', function (mdl, isSizesApplied, d) {
-      if (isSizesApplied) {
-        this.trigger('histogram_sizes', this);
+    this.bind('change:autoStyle', function (mdl, isAutoStyle, d) {
```
alonsogarciapablo: AFAIK we're no longer changing this... so is this really needed? Is anybody listening to the `autoStyle` event that is being triggered below 👇 ?

src/geo/map/cartodb-layer.js
alonsogarciapablo: Is this something we could get from the Maps API (as part of the response)?

javisantana: We could but I don't think it should. cartodb.js does not need to know anything about the geometry type, that's something tied to the editor not cartodb.js

javisantana: and I'd remove this method, basically guess the geometry type from the style set. Nothing avoids me to set a style with marker-width rendering polygons (we actually do it in bubble maps).

fdansv: > guess the geometry type from the style set.

please elaborate?

alonsogarciapablo: AFIK this is used [here](https://github.com/CartoDB/deep-insights.js/pull/304/files#diff-a7270c805528a9245fa1351d6719bcb2R7) to determine the cartocss that should be generated, based on the type of geometry... If getting the type of geometry is not doable, perhaps we can generate a more generic cartocss such as:

```
// points
#layer_name['mapnik::geometry_type'=1] {
  ...
}
// lines
#layer_name['mapnik::geometry_type'=2] {
  ...
}
// polygons
#layer_name['mapnik::geometry_type'=3] {
  ..
}
```

alonsogarciapablo:
👍  👍  👍

PR #1233, opened 2016-04-13
Fixes return value of the _getSourceId method

javierarce:
The `source` of a layer is not a model anymore, so the method should change.

### DIFF
--- a/src/dataviews/dataview-model-base.js
+++ b/src/dataviews/dataview-model-base.js
@@ -246,7 +246,7 @@ module.exports = Model.extend({
   _getSourceId: function () {
     var source = this.layer.get('source');
     if (source) {
-      return source.get('id');
+      return source;
     }
 
     return this.layer.get('id');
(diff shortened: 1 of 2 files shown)


src/dataviews/dataview-model-base.js
```diff
@@ -246,7 +246,7 @@ module.exports = Model.extend({
   _getSourceId: function () {
     var source = this.layer.get('source');
     if (source) {
-      return source.get('id');
+      return source;
     }
 
     return this.layer.get('id');
```
alonsogarciapablo: Now that I see it, we could do:

```
return this.layer.get('source') || this.layer.get('id');
```

alonsogarciapablo:
Left a suggestion. :+1:

src/dataviews/dataview-model-base.js
javierarce: True, changing it.

PR #1234, opened 2016-04-15
Update metadata of analysis models from map instantiation response

alonsogarciapablo:
With this PR, analysis models will get the following attributes once a response from Windshaft is received: `query`, `url`, and `status`.

I took some time to change the way the metadata of models is updated when a new instance of the map is created on Windshaft. I extracted that responsibility to a class (see `src/vis/model-updater` - I'm open to better names) and I was able to remove the dependency between the different models and the Windshaft object.

@javierarce could you please take a look? Thanks!


javierarce:
It seems that the last change introduced a bug… I've reverted the change.

javierarce:
👍  code LGTM

PR #1236, opened 2016-04-19
Dataviews can be associated to an analysis via the sourceId attribute

### LINKED ISSUE #1212: Dataviews must support Analysis as a type of source
Right now dataviews are directly linked to a layer (to the SQL of that layer basically). When the  source of a layer is changed to an analysis the resulting SQL for that layer changes, dataviews pointing to that layer will break (because the specified `column` won't exist in that new SQL).

alonsogarciapablo:
Closes #1212.

@javierarce can you please take a look? Thanks!

### DIFF
--- a/src/analysis/analysis-model.js
+++ b/src/analysis/analysis-model.js
@@ -25,7 +25,7 @@ module.exports = Model.extend({
       this._reloadMap();
     }, this);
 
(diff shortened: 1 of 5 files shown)


javierarce:
👍

javierarce:
👍  LGTM

PR #1237, opened 2016-04-21
Analyses polling

### LINKED ISSUE #1229: Integrate analysis statuses/urls from Maps API
Windshaft is now returning the ~statuses~ and URLs to get the actual status of each of the analysis nodes present in a visualization (*). This issue captures all the changes that we need to do in CartoDB.js.

**Requirements:**

- [x] If all nodes are ready -> proceed normally (fetch tiles, update URLs of dataviews, update analysis queries...).
- ~~If any of the nodes of the analyses is not ready yet (waiting, running, failed), CartoDB.js **should NOT**~~:
  - [ ] ~~fetch the new tiles (cause they won't be ready yet) (@saleiva: have you thought about this state? tiles won't be ready because analyses are being processed and new tiles cannot be fetched yet... should we display some sort of global loader?)~~.
  - [ ] ~~set the URLs of dataviews (widgets should change to "loading" state)~~.
- [x] CartoDB.js should use the URLs of each of the analysis nodes, to check periodically if "everything" is ready. Each analysis model should pull and update it's state from the Maps API.
- [ ] ~~Dataviews associated to an analysis node should listen to their state and update their URLs as soon as the associated analysis is ready (analysis of type "source" are always ready)~~.
- [x] CartoDB.js should fetch the tiles once all analyses are ready again.

cc: @javisantana @rochoa 

(*) Example of a response with analyses:

```
{
  "layergroupid": "7357a96cb13323a64251f93f15649572:0",
  "metadata": {
    "layers": [
      {
        "type": "mapnik",
        "meta": {
          "stats": [],
          "cartocss": "#points['mapnik::geometry_type'=1] {\n  marker-fill-opacity: 1;\n  marker-line-color: #FFF;\n  marker-line-width: 0.5;\n  marker-line-opacity: 1;\n  marker-placement: point;\n  marker-type: ellipse;\n  marker-width: 8;\n  marker-fill: #FABADA;\n  marker-allow-overlap: true;\n}\n#lines['mapnik::geometry_type'=2] {\n  line-color: #000000;\n  line-width: 2;\n  line-opacity: 1;\n}\n#polygons['mapnik::geometry_type'=3] {\n  polygon-fill: #FABADA;\n  polygon-opacity: 1;\n  line-color: #FFF;\n  line-width: 0.5;\n  line-opacity: 1;\n}"
        }
      }
    ],
    "dataviews": {},
    "analyses": [
      {
        "nodes": {
          "a1": {
            "query": "select * from analysis_trade_area_5eeaad22776ff8936767c173f4f92f8aa17c8ebc",
            "url": {
              "http": "http://cdb.localhost.lan:8181/api/v1/map/7357a96cb13323a64251f93f15649572:0/analysis/0/node/a1"
            },
            "status": "running"
          },
          "a0": {
            "query": "select * from arboles limit 10",
            "url": {
              "http": "http://cdb.localhost.lan:8181/api/v1/map/7357a96cb13323a64251f93f15649572:0/analysis/0/node/a0"
            },
            "status": "ready"
          }
        }
      }
    ]
  },
  "last_updated": "1970-01-01T00:00:00.000Z"
}
```

alonsogarciapablo:
Closes #1229:

- When a new map instance is created and analyses that are the source of a layer or a dataview are not "ready" or "failed", polling starts.
- Every time a new map instance is created, ongoing pollings are cleared.
- Loader is displayed while tiles are loading and analysis are not done.

Also, I renamed src/vis/vis.js to src/vis/vis.view.js and introduced a vis model in src/vis/vis.js so that we can start moving stuff out of src/vis/vis.view.js.

~~@xavijam~~ @viddo can you please CR? Thanks!


src/analysis/analysis-model.js
```diff
@@ -18,6 +26,14 @@ module.exports = Model.extend({
     this._initBinds();
   },
 
+  url: function () {
+    var url = this.get('url');
+    if (this.get('apiKey')) {
+      url += '?api_key=' + this.get('apiKey');
```
viddo: I propose to provide the query strings through the backbone expected data prop on options arg to fetch instead of building the full URL yourself, e.g.:

```js
fetch: function(options) {
  options = options || {};
  options.data = options.data || {};
  
  var apiKey = this.get('apiKey');
  if (apiKey) {
    options.data.api_key = apiKey;
  }
  
  return Model.prototype.fetch.call(this, options);
}
```

test/spec/api/create-vis.spec.js
```diff
@@ -242,21 +242,21 @@ describe('src/api/create-vis', function () {
     it('should instantiate map', function (done) {
       this.vis = createVis('domId', fakeVizJSON, {});
 
-      _.defer(function () {
+      setTimeout(function () {
         expect(mapInstantiationRequestDone()).toEqual(true);
         done();
-      });
+      }, 25);
```
viddo: I strongly recommend against using timeouts in tests (including `_.defer`), since they cause the tests to fail in-deterministically eventually. I'd recommend to find some other way to delay the assertion, e.g. jasmine.clock.

src/analysis/analysis-poller.js
```diff
@@ -0,0 +1,59 @@
+var _ = require('underscore');
+var BackbonePoller = require('backbone-poller');
+
+var AnalysisPoller = {
```
viddo: I'm feeling wary about having this object as a global/singleton, since separated vis object on the same page will share the cache and thus interfere with each other on poll/reset calls. What's the reason for not having these methods on a function prototype and create an instance when the vis is created instead?

viddo:
Some minor details, overall 👍

test/spec/api/create-vis.spec.js
alonsogarciapablo: I tried to avoid this timeout but we're using `_.debounce` in src/geo/map.js and jasmine.clock didn't work :(

src/analysis/analysis-poller.js
alonsogarciapablo: Good point. We don't have separated `vis` objects right now, but I'll change it. Thanks!

PR #1247, opened 2016-04-27
Poll all analysis nodes

alonsogarciapablo:
We were only polling analyses that were the source of a layer or dataview. This removes that limitation and polls all analyses instead. When an analysis that is the source of a layer or dataview becomes "ready", the map is reloaded...

cc: @viddo


viddo:
Looks legit, seems to work fine on editor too 👍

PR #1252, opened 2016-05-03
Discards plain layers as basemaps

fdansv:
Fixes https://github.com/CartoDB/d3.cartodb/issues/177 and a couple more.

cc @rochoa ok?


alonsogarciapablo:
👍

PR #1253, opened 2016-05-04
Updated camshaft

xavijam:
I feel it is better to set the proper version of the reference instead of ^0.X.0, what do you think @alonsogarciapablo?


alonsogarciapablo:
👍

PR #1254, opened 2016-05-04
Use latest browserify versions to rely on factor-bundle plugin

rochoa:
cc @xavijam


xavijam:
Not much to say, :+1:.

PR #1256, opened 2016-05-09
Adds method to histogram dataview model to obtain its shape

fdansv:
Following the same logic we used in Pecan: https://github.com/CartoDB/cartodb-postgresql/blob/master/scripts-available/CDB_DistType.sql cc @ohasselblad 

@javisantana @alonsogarciapablo mind having a look? :)


fdansv:
(cr'd by @javisantana and @rochoa offline)

PR #1259, opened 2016-05-10
Map config includes layer's sql_wrap option if present

### LINKED ISSUE #1258: add support for query_wrapper
the maps api supports query_wrapper so we should be able to send that attribute in the mapconfig. In the `layer.update` method we should allow to change that param

https://github.com/CartoDB/Windshaft-cartodb/pull/441

alonsogarciapablo:
Closes #1258.

cc: @javisantana @rochoa


javisantana:
thx mate :*

PR #1260, opened 2016-05-10
Ensure polling doesn't start for analysis nodes that are not the source of a layer or dataview

alonsogarciapablo:
Basically, this is currently needed because the editor is "creating" (via `vis.analysis.analyse`) analysis nodes of type `source` when new layers are added and those nodes are not the source of a layer or dataview and so, they don't have a URL yet...

@viddo can you take a look please? thanks!


viddo:
Code looks legit, but the tests don't seem to pass

alonsogarciapablo:
Retest this please

alonsogarciapablo:
Just realised this is not right. "Intermediate" nodes should actually be polled and they are not the source of layers of dataviews. I'll need to take a different approach, so closing this.

PR #1262, opened 2016-05-11
Polling should not start for analysis nodes that don't have a URL yet

alonsogarciapablo:
Closes CartoDB/cartodb#7319.

@viddo, this is the replacement for CartoDB/cartodb.js#1260. Basically, analysis nodes might be created/defined using `vis.analysis.analyse`, but they might not have been sent to the Maps API for some reason (eg: they are not linked to a layer/dataview yet) so they should not be polled.

cc: @xavijam @juanignaciosl


viddo:
👍

PR #1265, opened 2016-05-13
Do not try to include 'interactivity' and 'attributes' options in the map config for layers that don't have infowindow

alonsogarciapablo:
cc: @javisantana


javisantana:
dale candela :es:

PR #1268, opened 2016-05-19
Torque layer reloading

### LINKED ISSUE #1266: calling update on a torque layer reload the tiles
When updating a torque layer using the `update` method, changing cartocss triggers a full map reload.

It also fails to update the style.

steps to reproduce:

1) run this code
```
layer.update("Map { -torque-frame-count: 256; -torque-animation-duration: 30; -torque-time-attribute: "cartodb_id"; -torque-aggregation-function: "count(1)"; -torque-resolution: 2; -torque-data-aggregation: linear; }#layer { marker-width: 10; marker-fill: #29a846; marker-fill-opacity: 1; marker-allow-overlap: true; marker-line-width: 2; marker-line-color: #FFFFFF; marker-line-opacity: 1; }");

layer.update("Map { -torque-frame-count: 256; -torque-animation-duration: 30; -torque-time-attribute: "cartodb_id"; -torque-aggregation-function: "count(1)"; -torque-resolution: 2; -torque-data-aggregation: linear; }#layer { marker-width: 10; marker-fill: #F00; marker-fill-opacity: 1; marker-allow-overlap: true; marker-line-width: 2; marker-line-color: #FFFFFF; marker-line-opacity: 1; }"); // change marker-fill
```

tiles are reloaded, the style is not update and it should.

Take into account if `torque-time-attribute` was the change property it should reload map tiles

alonsogarciapablo:
(Hopefully) closes #1266.

@javisantana: yay or nay?

### DIFF
--- a/src/geo/map/torque-layer.js
+++ b/src/geo/map/torque-layer.js
@@ -21,7 +21,13 @@ var TorqueLayer = LayerModelBase.extend({
 
   ATTRIBUTES_THAT_TRIGGER_MAP_RELOAD: ['visible', 'sql', 'source', 'cartocss'],
 
-  TORQUE_TIME_ATTRIBUTE_PROP: '-torque-time-attribute',
+  TORQUE_LAYER_CARTOCSS_PROPS: [
+    '-torque-frame-count',
+    '-torque-time-attribute',
+    '-torque-aggregation-function',
+    '-torque-data-aggregation',
+    '-torque-resolution'
+  ],
 
   LAYER_NAME_IN_CARTO_CSS: 'Map',
 
@@ -37,7 +43,7 @@ var TorqueLayer = LayerModelBase.extend({
     var reloadMap = _.any(this.ATTRIBUTES_THAT_TRIGGER_MAP_RELOAD, function (attr) {
       if (this.hasChanged(attr)) {
         if (attr === 'cartocss') {
-          return this.previous('cartocss') && this._torqueTimeAttributeCartoCSSPropChanged();
+          return this.previous('cartocss') && this._torqueCartoCSSPropsChanged();
         }
         return true;
       }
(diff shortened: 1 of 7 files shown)


src/geo/map/torque-layer.js
```diff
@@ -18,12 +18,36 @@ var TorqueLayer = LayerModelBase.extend({
...
+        return true;
+      }
+    }, this);
+
+    if (reloadMap) {
+      this._reloadMap();
+    }
+  },
+
+  _torqueTimeAttributeCartoCSSPropChanged: function () {
```
javisantana: I don't think this is the right way to do it, check this method that allows to get properties from cartocss https://github.com/CartoDB/torque/blob/31a81e760b249e76b159c9966c219ae570f9acb3/lib/torque/common.js#L33

alonsogarciapablo: Changed this in [09e10e742b714f7dc4fc7436522bb159f3c3961d](https://github.com/CartoDB/cartodb.js/pull/1268/commits/09e10e742b714f7dc4fc7436522bb159f3c3961d).

src/geo/map/torque-layer.js
```diff
@@ -44,10 +49,18 @@ var TorqueLayer = LayerModelBase.extend({
   },
 
   _torqueTimeAttributeCartoCSSPropChanged: function () {
-    var regexp = /-torque-time-attribute:\s*["'](\w*)["']/;
-    var previousTorqueTimeAttributeProp = this.previous('cartocss').match(regexp) && this.previous('cartocss').match(regexp)[1];
-    var newTorqueTimeAttributeProp = this.get('cartocss').match(regexp) && this.get('cartocss').match(regexp)[1];
-    return newTorqueTimeAttributeProp !== previousTorqueTimeAttributeProp;
+    var currentCartoCSS = this.get('cartocss');
+    var previousCartoCSS = this.previous('cartocss');
+    var currentValue = this._getPropertyValueFromCartoCSS(currentCartoCSS, this.TORQUE_TIME_ATTRIBUTE_PROP);
```
javisantana: you are parsing several times the cartocss and it's an expensive operation

javisantana: you need to check not only time attribute, steps, resolution, column and aggregation are also needed

alonsogarciapablo: > if you check more than one attrs you will be parsing

I don't understand what you mean by that...

I'm adding the rest of the attributes and I think I I'll need to parse both the new and old cartocss (so parsing will only be done twice)...

alonsogarciapablo: What about `-torque-animation-duration` and `-torque-resolution:2` ?

javisantana: -torque-animation-duration is managed client side, -torque-resolution needs a map reload (it's the pixel resolution, it changes the aggregation)

javisantana:
tested on the editor, working fine :es:

alonsogarciapablo:
Happy to hear that! 👍

PR #1270, opened 2016-05-23
Add default infowindow template instead of trying to load "infowindow_light"

alonsogarciapablo:
Fixes CartoDB/cartodb#7421.

@xavijam 👍  or 👎 ?

### DIFF
--- a/src/geo/ui/infowindow-model.js
+++ b/src/geo/ui/infowindow-model.js
@@ -1,31 +1,6 @@
 var _ = require('underscore');
 var Backbone = require('backbone');
-
-var defaultInfowindowTemplate = [
-  '<div class="CDB-infowindow CDB-infowindow--light js-infowindow">',
-  '  <div class="CDB-infowindow-container">',
-  '    <div class="CDB-infowindow-bg">',
-  '      <div class="CDB-infowindow-inner">',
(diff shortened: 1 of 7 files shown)


src/geo/ui/infowindow-model.js
```diff
@@ -1,31 +1,58 @@
...
 
-/**
- * Usage:
- * var infowindowModel = new InfowindowModel({
- *   template_name: 'infowindow_light',
- *   latlng: [72, -45],
- *   offset: [100, 10]
- * });
- */
+var defaultInfowindowTemplate = [
```
xavijam: Could we move to a template file?

xavijam:
It sounds like a workaround, and I'm scared that there is no better solution and the same template is in several places :S.

src/geo/ui/infowindow-model.js
alonsogarciapablo: Moved in [4138bdca9d15ad3d3f9f6085db6a97f2c5f12d86](https://github.com/CartoDB/cartodb.js/pull/1270/commits/4138bdca9d15ad3d3f9f6085db6a97f2c5f12d86).

xavijam:
yay! 👍

alonsogarciapablo:
retest this please

PR #1272, opened 2016-05-23
Make sure indexes are correct in the styles key

alonsogarciapablo:
Fixes CartoDB/cartodb#7422.

@fdansv can you please review this? Thanks!

cc: @javitonino

### DIFF
--- a/src/windshaft/named-map.js
+++ b/src/windshaft/named-map.js
@@ -4,18 +4,29 @@ var MapBase = require('./map-base.js');
 var NamedMap = MapBase.extend({
   toJSON: function () {
     var json = {};
-    var layers = this._getLayers();
-    var styles = layers.reduce(function (p, c, i) {
+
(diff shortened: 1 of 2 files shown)


fdansv:
Looking good! I was only wondering if in the test, 'Tiled' covers also plain layers. I'm saying this because we often forget to consider plain layers when implementing new stuff and it'd be useful that the test fails whenever that happens.

javitonino:
Hey @alonsogarciapablo. I was going to take a look at this but it is closed, should I take a look anyway?

alonsogarciapablo:
Not sure what happened. I reopened it again.

javitonino:
Works great! Ship it!

alonsogarciapablo:
@fdansv it's not very relevant in this case. I just wanted to test the case where a layer with no cartocss was above the rest of CartoDB layers.

alonsogarciapablo:
Thanks guys!

PR #1276, opened 2016-05-24
Mark analysis node that changed as "failed" if request to Maps API fails

### LINKED ISSUE #1274: Failing analysis doesn't update status attr
Add an analysis which causes the analysis fails with a 400 bad request error.

Expected result:
- Concerned node to trigger change event due to changed attributes, its `status` attr to be `failed`

Actual result:
- No changes, no status update

alonsogarciapablo:
Fixes #1274.

@viddo 👍  ?


viddo:
Looks legit 👍

PR #1282, opened 2016-05-25
Attribution issues

xavijam:
- [x] Parsing map attributes in parse function instead of initialize function.
- [x] Fixed problem with attribution-view where it was unbinding map attribution changes 😱 .

CR: @alonsogarciapablo


src/geo/ui/attribution/attribution-view.js
```diff
@@ -23,14 +23,12 @@ var Attribution = View.extend({
       visible: false
     });
     this.map = this.options.map;
-    this.model.bind('change:visible', function (mdl, isVisible) {
-      this[ isVisible ? '_showAttributions' : '_hideAttributions' ]();
-    }, this);
-    _.bindAll(this, '_onKeyDown', '_toggleAttributions');
+
```
alonsogarciapablo: Out of curiosity... why are we moving from ☝️ to 👇 ? I kind of like `_.bindAll` better...

xavijam: I don't like to bind not necessary things.

src/geo/ui/attribution/attribution-view.js
```diff
@@ -43,15 +41,22 @@ var Attribution = View.extend({
   },
 
   _initBinds: function () {
```
alonsogarciapablo: Are we actually using this method?

src/geo/ui/attribution/attribution-view.js
alonsogarciapablo: From what I understand,` _.bindAll(this, '_onKeyDown', '_toggleAttributions');` will only bind the `this` context to `_onKeyDown` and `_toggleAttributions` functions, so it should be the same as 👇 , right?

src/geo/ui/attribution/attribution-view.js
xavijam: Yep, I forgot to keep using it 😅

xavijam:
Tests now passing...

src/geo/ui/attribution/attribution-view.js
alonsogarciapablo: I'm still curious about this...

xavijam: It is the same thing, in fact `_.bind` uses `….bind(this)` internally, so, no difference, just manias 😬

src/geo/map.js
```diff
@@ -26,39 +25,43 @@ var Map = Model.extend({
 
   RELOAD_DEBOUNCE_TIME: 10,
 
+  parse: function (r) {
```
alonsogarciapablo: According to [Backbone docs](http://backbonejs.org/#Model-parse):

> **parse** is called whenever a model's data is returned by the server, in fetch, and save. The function is passed the raw response object, and should return the attributes hash to be set on the model. The default implementation is a no-op, simply passing through the JSON response. Override this if you need to work with a preexisting API, or better namespace your responses.

I'd personally not use it in this case since there's no server response here, and It forces us to always do:

```
var map = new Map({ ... }, { parse: true })
```

This is not very convenient IMO... It's easy to forget. I originally added this code to the constructor but I was probably wrong. We could acutally move it [here](https://github.com/CartoDB/cartodb.js/blob/v4/src/api/vizjson.js#L10) (we'd still need to set original_center, original_view_bounds_sw and original_view_bounds_ne in the `initialize` method). What do you think?

src/geo/ui/attribution/attribution-view.js
```diff
@@ -23,14 +23,14 @@ var Attribution = View.extend({
...
-    _.bindAll(this, '_onKeyDown', '_toggleAttributions');
+
+    this._onDocumentClick = this._onDocumentClick.bind(this);
+    this._onDocumentKeyDown = this._onDocumentKeyDown.bind(this);
+
+    this._initBinds();
   },
 
   render: function () {
-    this._disableBinds();
```
alonsogarciapablo: 👍

alonsogarciapablo:
I left a comment/suggestion about usage of `parse`...

About the attributions BUG: 👍

xavijam:
Let me remove the parse map thing

xavijam:
👍

alonsogarciapablo:
👍

PR #1284, opened 2016-05-26
Makes dataviews consider temporary states

fdansv:
@javisantana

### DIFF
--- a/src/dataviews/category-dataview-model.js
+++ b/src/dataviews/category-dataview-model.js
@@ -29,7 +29,7 @@ module.exports = DataviewModelBase.extend({
     return params;
   },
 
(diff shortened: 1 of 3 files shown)


src/dataviews/category-dataview-model.js
```diff
@@ -29,7 +29,7 @@ module.exports = DataviewModelBase.extend({
     return params;
   },
 
-  initialize: function (attrs, opts) {
+  initialize: function (attrs, opts, state) {
```
javisantana: I think this model should not know anything about state, I'd use attributes for this, if they have a filter attribute just set it

Is there any reason to do that?

PR #1286, opened 2016-05-26
Hide infowindow when fields are cleared

### LINKED ISSUE #1267: If infowindow fields are set as empty, infowindow view should be closed
As an example, in the editor 3, when we clear the fields of an infowindow model, it doesn't trigger the close of the infowindow.

alonsogarciapablo:
Fixes #1267.

I also implemented some changes so that the filter of the tooltipView is cleared when the infowindow is closed. This was causing tooltips not to work on the feature that was clicked and infowindow was then closed.

@xavijam since you opened that issue... 👍 or 👎 ? Thanks!


xavijam:
:+1:

PR #1288, opened 2016-05-27
Always fetch attributes when clicking on a feature (fields might have changed)

alonsogarciapablo:
@matallo please take a look! Thanks!


matallo:
LGTM, needed for the connection with the infowindowDefinitionmodel, plus I suppose this call will be cached in the browser

👍

PR #1291, opened 2016-05-31
fixes layergroup view not being cleared

### LINKED ISSUE #1289: grouplayer instance is not cleared when last layer is removed
when you add a cartodb layer to a map and then remove it a new layer can't be added.

The problem is here https://github.com/CartoDB/cartodb.js/blob/v4/src/geo/map-view.js#L181

The layer is removed but `_cartoDBLayerGroupView` is not cleared so this check (https://github.com/CartoDB/cartodb.js/blob/v4/src/geo/map-view.js#L146) avoids to create it again

javisantana:
fixes #1289

### DIFF
--- a/src/geo/map-view.js
+++ b/src/geo/map-view.js
@@ -17,6 +17,7 @@ var MapView = View.extend({
       throw new Error('you should specify a layerViewFactory');
     }
 
+    this._cartoDBLayerGroupView = null;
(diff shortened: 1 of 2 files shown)


javisantana:
cc @alonsogarciapablo

alonsogarciapablo:
🇪🇸

PR #1292, opened 2016-05-31
Allow creating layers at a given position

alonsogarciapablo:
This PR changes things so that `map.createCartoDBLayer`, `map.createTorqueLayer`, etc. accept `at` option to specify where the position where the layer should be inserted. Orders are automatically re-assigned [here](https://github.com/CartoDB/cartodb.js/blob/v4/src/geo/map/layers.js#L16). Also refactored the method that assign indexes a bit.

@viddo would you please take a look? thx!


viddo:
Looks legit 👍

alonsogarciapablo:
Thanks!

PR #1294, opened 2016-06-02
Fixed analysis section in map config for anonymous maps:

alonsogarciapablo:
Fixes https://github.com/CartoDB/cartodb/issues/7605.

- Dataviews get sourceId from `{ source: { id: 'X' } }` attribute (that's what deep-insights.js is sending). Before, CartoDB.js was assuming attribute was called 'sourceId'.
- Assume dataviews will always have a source, instead of falling back to layer.
- Include an analysis node for dataviews in the map config for named maps, which was missing before.

@viddo 👍  👎  ? Thanks!

cc: @javisantana

### DIFF
--- a/src/dataviews/category-dataview-model.js
+++ b/src/dataviews/category-dataview-model.js
@@ -261,7 +261,7 @@ module.exports = DataviewModelBase.extend({
   toJSON: function () {
     return {
       type: 'aggregation',
-      source: { id: this._getSourceId() },
+      source: { id: this.getSourceId() },
       options: {
         column: this.get('column'),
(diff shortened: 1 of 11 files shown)


viddo:
This might do it here, but the editor is sending `source_id` and not `source` (or `sourceId` for that matter), so this fix might not work alone anyways. I'll investigate it and try adapt the editor if this alone doesn't work. 👍

viddo:
Verified working (with some minor adaptions on the editor side).
Travis fails due to npm issues, but works fine on our own CI, so merging.

javisantana:
:es::es::es::es::es:

PR #1295, opened 2016-06-02
Source attribute for dataviews is not required anymore

alonsogarciapablo:
- Instead a default sourceId is generated based on the layer id. This will fix the tests and examples in deep-insights.js
- Also, an analysis source is included in the map config for dataviews that have a layer as their source.

@fdansv please take a look. thanks!


fdansv:
🚢  🇮🇹

Issue #1283, opened 2016-05-26
Infinite loop on Map instantiation success => Analysis poller.start()


viddo:
Found this while investigating https://github.com/CartoDB/cartodb/issues/7523 but have not managed to pinpoint exactly _why_.

The calls from the editor looks legit and are done, but then cartodb.js enters an infinite loop of requests between checking status of a node and doing an map-instantiation:

- [windshaft/map-base `instanceCreated` event](https://github.com/CartoDB/cartodb.js/blob/v4/src/windshaft/map-base.js#L77) - [`visView_onMapInstanceCreated`](https://github.com/CartoDB/cartodb.js/blob/v4/src/vis/vis-view.js#L266)
- [`analysisPoller.start()`](https://github.com/CartoDB/cartodb.js/blob/v4/src/analysis/analysis-poller.js#L19)
- start over…

Seems like the point-in-polygon node is in `pending` state during this time, if that helps.

viddo:
Could this be related to recent changes of [#1276 Mark analysis node that changed as "failed" if request to Maps API fails](https://github.com/CartoDB/cartodb.js/pull/1276)?

alonsogarciapablo:
Yes, so the problem is that 

1. Map instance is created.
2. One of the analysis nodes (point in polygon) is "pending" so polling starts.
3. When node becomes "ready" a new map instance is created.
4. One of the analysis nodes (point in polygon) is "pending" **again** so polling starts **again**
5. rinse, repeat ...

In theory, the Maps API should somehow know that the analysis has been completed before and should return the "ready" status as long as we send the same payload.

Talked to @rochoa about it and it seems something that needs to be fixed at the Maps API level, so I'll open an issue there.

viddo:
Although, shouldn't cartodb.js be resilient to this kind of malfunctions? E.g. in this case, if it's already polling why start the poll again?

alonsogarciapablo:
CartoDB.js keeps polling as long as the node is "pending". When the node becomes "ready" polling stops and a new instance of the map is created. At that point, from the POV of CartoDB.js, we're "starting from scratch again", so it'd be hard to tell wether a change from "ready" to "pending" state in a node should trigger a map reload or not.

I guess one way to stop this particular infinite loop would be to check if the status of analysis node that was "ready" has changed, and the payload that was used to create the map is the same as the payload that will be used again...

viddo:
[...]

PR #1296, opened 2016-06-03
Limit the number of subsequent requests to the Maps API using the same payload, params and options

alonsogarciapablo:
Fixes: https://github.com/CartoDB/cartodb.js/issues/1283.

@viddo I'd like to know your thoughts about this approach... CartoDB.js won't try to instantiate the exact same map more than three times with this change.

cc: @rochoa

### DIFF
--- a/src/windshaft/map-base.js
+++ b/src/windshaft/map-base.js
@@ -74,13 +74,11 @@ var WindshaftMap = Backbone.Model.extend({
     return !this._requestTracker.maxNumberOfRequestsReached();
   },
 
-  _trackRequest: function (request) {
-    this._requestTracker.track(request);
+  _trackRequest: function (request, response) {
+    this._requestTracker.track(request, response);
   },
 
   _performRequest: function (request) {
-    this._trackRequest(request);
-
     var payload = request.payload;
     var params = request.params;
     var options = request.options;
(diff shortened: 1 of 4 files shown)


src/windshaft/map-base.js
```diff
@@ -52,28 +59,37 @@ var WindshaftMap = Backbone.Model.extend({
...
-      stat_tag: this.get('statTag')
-    };
-
-    if (this.get('apiKey')) {
-      params.api_key = this.get('apiKey');
+    var request = new Request(payload, params, options);
+    if (this._canPerformRequest(request)) {
+      this._trackRequest(request);
+      this._performRequest(request);
     }
```
alonsogarciapablo: We're not doing anything when the request cannot be performed... Should we raise an error?

viddo:
Aight, I wait with reviewing it then, ok?

alonsogarciapablo:
Yes please. I'll ping you here once I'm done.

alonsogarciapablo:
@viddo ready again! With the latest change, request to instantiate map will only be triggered if something has changed (payload, params, options) and one of responses from the 3 previous requests is different.

viddo:
looks legit, not much to say 👍

Issue #1298, opened 2016-06-06
Torque is broken under RequireJS or Dojo


redhog:
If an AMD loader is present when the Torque module is loaded by CartoDB (regardless of how CartoDB itself was loaded), the following error shows up in the console, and the torque layer is not loaded: "cartodb.mod.torque.uncompressed.js:5388 Uncaught ReferenceError: torque is not defined"

As far as I can tell the reason is that if cartodb.mod.torque.uncompressed.js detects an AMD capable define() function, it uses this for defining the torque module and sub modules instead of setting window.torque here: https://github.com/CartoDB/cartodb.js-bower/blob/master/cartodb.mod.torque.uncompressed.js#L11 

However, later in the code, on the top level, there is an assumption that window.torque has already been defined, even though this will be done asynchronously: https://github.com/CartoDB/cartodb.js-bower/blob/master/cartodb.mod.torque.uncompressed.js#L5388

ivanmalagon:
cartodb.js v3 is not under active development. Only major bugs are addressed. (Like recently where we had to add support for the new Google Maps renderer).

This is stated in https://github.com/CartoDB/carto.js/blob/develop/README.md

I'm sorry but feel free to use your own fork. There'll be no activity in v3 so it'd be easy to keep updated with this upstream in case we make some fixes.

PR #1300, opened 2016-06-07
3.15.8 amd torque

redhog:
Work around / possible solution for https://github.com/CartoDB/cartodb.js/issues/1298

### DIFF
--- a/scripts/mod.torque.footer.js
+++ b/scripts/mod.torque.footer.js
@@ -6,3 +6,5 @@ Profiler = cartodb.core.Profiler
 ['require', 'define', 'module'].map(function (name) {
   window[name] = window.cartotmp[name];
 });
+
+delete window.cartotmp;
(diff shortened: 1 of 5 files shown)


Cartofante:

scripts/mod.torque.footer.js
```diff
@@ -3,3 +3,6 @@ cartodb.moduleLoad('torque', torque);
 
 Profiler = cartodb.core.Profiler
 
+['require', 'define', 'module'].map(function (name) {
+  window[name] = window.cartotmp[name];
```
javisantana: could we delete cartotmp after this so the global namespace is not polluted?

redhog: Added a commit that does that :)

javisantana:
thanks @redhog 

@alonsogarciapablo your call

rubenmoya:
As stated in https://github.com/CartoDB/carto.js/issues/1298#issuecomment-404193738, cartodb.js v3 is not under active development. Closing this.

PR #1305, opened 2016-06-10
fix break word

piensaenpixel:
@xavijam with love for you :)


xavijam:
No better option then? :(

piensaenpixel:
no... but i think It can work

xavijam:
👍 💋 ❤️

Issue #1299, opened 2016-06-07
Filtering should always be done server side when filtering dataviews linked to an analysis 


alonsogarciapablo:
If a map is being rendered using vector tiles and a dataview is linked to an analysis node, filtering should NOT be done client side (d3.cartodb doesn't know anything about analyses). Instead, map should be re-instantiated and GeoJSON tiles should be fetched again.

Also, dataviews that have source that is different to the layer's source should not try to get data from the vector tiles.

cc: @javisantana @fdansv

alonsogarciapablo:
Closed via https://github.com/CartoDB/cartodb.js/pull/1306

PR #1306, opened 2016-06-14
Fix vector + analysis + dataviews 

alonsogarciapablo:
close https://github.com/CartoDB/cartodb.js/issues/1299

1. There's a new class (geojson/data-provider.js) that represents a dataProvider for a given layer. This class now triggers the 'dataChanged' event when new data is available.
2. Dataviews use the above mentioned class to:
  - Check if it can provide data for a given dataview.
  - Retrieve data for a given dataview.
  - Check if a filter can be applied.
  - Apply the given filter using the column from the given dataview.
3. When a dataview doesn't have the same source as it's layer (and vector is enabled ofc):
  - Data for that dataview is retrieved from Windshaft.
  - Filtering triggers a new request to the Maps API and GeoJSON tiles are loaded again.

@fdansv can you please take a look? Thanks!


fdansv:
🚢 🇮🇹

PR #1312, opened 2016-06-21
Relax camshaft-reference version used

viddo:
Same motivation as for #1199
A `npm update camshaft-reference` would be needed to update locally

cc @rochoa @alonsogarciapablo ok?

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "backbone": "1.2.3",
     "backbone-poller": "^1.1.3",
     "browserify-shim": "3.8.12",
-    "camshaft-reference": "0.4.0",
+    "camshaft-reference": "1.x || <1.x",
     "d3": "3.5.8",
     "d3.cartodb": "CartoDB/d3.cartodb#gh-pages",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -28,7 +28,7 @@
     "backbone": "1.2.3",
     "backbone-poller": "^1.1.3",
     "browserify-shim": "3.8.12",
-    "camshaft-reference": "0.4.0",
+    "camshaft-reference": "1.x || <1.x",
```
alonsogarciapablo: What does this mean? Do we want to enforce 1.x? Why do we need the `|| <1.x` part?

viddo: 1.x would mean `>=1.0.0 <2.0.0` which would exclude the current versions, a mayor version of 0 works differently than if >=1. I set it to this to have version range `>=0.0.0 <2.0.0`. See https://docs.npmjs.com/misc/semver#caret-ranges-123-025-004 for more details

alonsogarciapablo:
👍

PR #1313, opened 2016-06-23
cdb.createVis returns a vis model that triggers load event when vizjson has been loaded

alonsogarciapablo:
Prep work for https://github.com/CartoDB/cartodb/issues/7574.

Changes:
- `cdb.createVis` returns a vis model object that responds to `.done` and `.error`. I didn't change how this works and done callback will be called when the map is first instantiated.
- `vis` object triggers `.load` event when vizjson has been loaded so that apps that use `createVis` with `skipMapInstantiation` option can instantiate the map when thy need to when everything is loaded (eg: deep-insights.js -> see https://github.com/CartoDB/deep-insights.js/pull/380).
- To make ^^ happen, I moved model related stuff from `vis-view.js` to `vis.js` (everything related with loading the vizjson).

@viddo: could you please take a look? Thx!

### DIFF
--- a/src/vis/vis-view.js
+++ b/src/vis/vis-view.js
@@ -72,14 +72,9 @@ var Vis = View.extend({
 
     this.mapView = mapViewFactory.createMapView(this.model.map.get('provider'), this.model.map, div_hack, this.model.layerGroupModel);
 
-    // Bindings
-    if (options.legends || (options.legends === undefined && this.model.map.get('legends') !== false)) {
-      this.model.map.layers.bind('reset', this.addLegends, this);
-    }
-
     // Infowindows && Tooltips
     var infowindowManager = new InfowindowManager(this, {
-      showEmptyFields: options.show_empty_infowindow_fields
+      showEmptyFields: this.model.get('showEmptyInfowindowFields')
     });
     infowindowManager.manage(this.mapView, this.model.map);
 
@@ -87,6 +82,11 @@ var Vis = View.extend({
     tooltipManager.manage(this.mapView, this.model.map);
 
     this.mapView.bind('newLayerView', this._bindLayerViewToLoader, this);
+
+    // Bindings
+    if (this.model.get('showLegends')) {
+      this.addLegends();
+    }
 
     this._addOverlays(options);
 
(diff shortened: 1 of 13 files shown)


src/vis/vis-view.js
```diff
@@ -49,113 +30,17 @@ var Vis = View.extend({
         }
       }
     }, this);
-  },
 
-  done: function (fn) {
-    return this.bind('done', fn);
-  },
+    this.model.on('load', this.render, this);
```
viddo: Can this be triggered multiple times? Otherwise a `.once` might be preferable.

src/vis/vis-view.js
```diff
@@ -49,113 +30,17 @@ var Vis = View.extend({
...
-    // the center provided in the parameters and the
-    // correct size.
-    var map_h = this.$el.outerHeight();
-    if (map_h === 0) {
-      $(window).bind('resize', function () {
-        this._onResize();
-      }.bind(this));
-    }
+  render: function () {
+    // TODO: Get this from this.model
```
viddo: What's "this"? I'd recommend to create an issue to keep track of work your plan to do in the future instead of leaving a TODO here

viddo:
> Prep work for CartoDB/cartodb#7574.

The desc sounds reasonable, but I lack some info what this PR is preparing _for_? :)

The changes looks legit. 👍

src/vis/vis-view.js
alonsogarciapablo: 👍

src/vis/vis-view.js
alonsogarciapablo: I forgot it. I'll take care of it in this PR. Thanks!

alonsogarciapablo:
> I lack some info what this PR is preparing for? :)

Basically, before this PR, createVis was returning an instance of `vis-view.js` which had a lot of stuff that belongs to a model. Since we want to expose some error through the `vis` object I feel this was a great opportunity to tackle this and I think it'll make it easier now to work with that `vis` object internally :)

alonsogarciapablo:
@viddo I addressed your feedback in https://github.com/CartoDB/cartodb.js/pull/1313/commits/9bc7d6e28461297c9c8fe8f1a68b400d5cb68d14. Would love a second pair of 👀  on that commit to make sure it makes sense. Thanks!

viddo:
Thx for clarification, that makes it easier to understand the _why_ behind the changes. :)
Regarding the last changes I don't really have anything particular to say, looks legit and accurate (from the little knowledge I have of these parts).

alonsogarciapablo:
Retest this please

PR #1315, opened 2016-06-23
Abort previous requests to the Maps API before sending a new one

alonsogarciapablo:
@javisantana Didn't think too much about this, but this could be a start... what do you think? Am I missing something (apart from some tests 😄)? Thanks!

### DIFF
--- a/src/windshaft/client.js
+++ b/src/windshaft/client.js
@@ -74,7 +74,8 @@ WindshaftClient.prototype._usePOST = function (payload, params) {
 };
 
 WindshaftClient.prototype._post = function (payload, params, options) {
-  $.ajax({
+  this._abortPreviousRequest();
+  this._previousRequest = $.ajax({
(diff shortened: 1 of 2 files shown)


javisantana:
probably is enough. I don't know if cancelling a request is enough for the maps api to cancel the whole thing cc @rochoa

javisantana:
but from the cartodb.js that's good enough I'd say

rochoa:
This is kind of related to:
 - https://github.com/CartoDB/cartodb/issues/8052.
 - https://github.com/CartoDB/Windshaft/issues/176.

We have to start thinking about the best way to handle _cancellations_.

alonsogarciapablo:
Ok. I'll add some tests and merge this as a first step. It we implement a mechanism to cancel stuff that is in progress, we can build it on top of this. Thx

PR #1318, opened 2016-06-24
vis.map triggers `dragend` event

alonsogarciapablo:
Title says it all :)

@xavijam this is what we need in https://github.com/CartoDB/cartodb/pull/8182 😄

### DIFF
--- a/src/geo/leaflet/leaflet-map-view.js
+++ b/src/geo/leaflet/leaflet-map-view.js
@@ -90,7 +90,11 @@ var LeafletMapView = MapView.extend({
     this._leafletMap.on('dragend', function() {
       var c = self._leafletMap.getCenter();
       this.trigger('dragend', [c.lat, c.lng]);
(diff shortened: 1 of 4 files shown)


xavijam:
Could we trigger moveend too?

xavijam:
yes!

src/geo/leaflet/leaflet-map-view.js
```diff
@@ -92,6 +92,11 @@ var LeafletMapView = MapView.extend({
       this.trigger('dragend', [c.lat, c.lng]);
     }, this);
 
+    this._leafletMap.on('moveend', function() {
+      var c = self._leafletMap.getCenter();
+      self.map.trigger('moveend', [c.lat, c.lng]);
```
xavijam: I feel it should be this.trigger('moveend', [c.lat, c.lng])

xavijam: Forget it, working like a charm.

PR #1319, opened 2016-06-24
Fixing search problem

xavijam:
Vis model is not included within vis, now it is available under `data.map`.

cc @alonsogarciapablo


xavijam:
Passing tests, merging :)

PR #1321, opened 2016-06-25
Reloading map when widget with filter is removed

xavijam:
Filter and widget were removed correctly, but map was not reloaded. Now it checks if there was a filter applied and it reloads the map.

CR: @alonsogarciapablo 
Fixes https://github.com/CartoDB/cartodb/issues/7812.
cc @saleiva


alonsogarciapablo:
👍

PR #1324, opened 2016-06-27
Expose some errors from Maps API

alonsogarciapablo:
First iteration on https://github.com/CartoDB/cartodb/issues/7574.

So far:

- `vis` model triggers:
  - `ok` event when a new instance of the "map" is created successfully in the Maps API.
  - `error` event when an unknown error occured error occurred (other errors will be triggered by more specific models like layers). **Error event will only include an error message for now (we can iterate on this and send more info)**.
- `vis` model still supports `.done(callback)` and `.error(callback)`. One of these two callbacks will be invoked when `cdb.createVis` is used without the `skipMapInstantiation` option:

  ```
  var vis = createVis(...);
  vis.done(function () { ... });
  vis.error(function (error) { ... });

  // Also, we can listen to `ok` and `error` events
  vis.on('ok', function () { ... });
  vis.on('error'), function () { ... });
  ```

- When using `cdb.createVis` and setting `skipMapInstantiation` is set to true (only happening in `deep-insights.js` currently) `.done` or `.error` callback will be invoked once `vis.instantiateMap` has been executed:

  ```
  var vis = createVis(..., { skipMapInstantiation: true });
  vis.done(function () { ... });
  vis.error(function (error) { ... });
  vis.on('load', function (vis) { // Waits for the vizjson to be loaded
    vis.instantiateMap();
  )};

  // Also, we can listen to `ok` and `error` events
  vis.on('ok', function () { ... });
  vis.on('error'), function () { ... });
  ```

My plan is to follow a similar approach for layer, dataviews and analysis models:
  - models expose a `setOk` and `setError` methods, that basically trigger `ok` and `error` events.
  - `ModelUpdater#setErrors` knows how errors coming from the Maps API and "distribute" them to the corresponding models based on the error type.

@viddo I'd love to get some feedback on this, before following the same approach for other models. Thx!

UPDATE: I pushed https://github.com/CartoDB/cartodb.js/pull/1324/commits/1163bfc318b332a92228223b41e8bd4da3dccc51, which implements something similar for analysis models.

cc: @xavijam

### DIFF
--- a/src/vis/vis.js
+++ b/src/vis/vis.js
@@ -15,11 +15,9 @@ var LayersCollection = require('../geo/map/layers');
 var AnalysisPoller = require('../analysis/analysis-poller');
 var Layers = require('./vis/layers');
 
-var STATES = {
-  INIT: 'init', // vis hasn't been sent to Windshaft
-  OK: 'ok', // vis has been sent to Windshaft and everything is ok
-  ERROR: 'error' // vis has been sent to Windshaft and there were some issues
-};
+var STATE_INIT = 'init'; // vis hasn't been sent to Windshaft
+var STATE_OK = 'ok'; // vis has been sent to Windshaft and everything is ok
+var STATE_ERROR = 'error'; // vis has been sent to Windshaft and there were some issues
 
 var VisModel = Backbone.Model.extend({
   defaults: {
@@ -27,7 +25,7 @@ var VisModel = Backbone.Model.extend({
     https: false,
     showLegends: false,
     showEmptyInfowindowFields: false,
-    state: STATES.INIT
+    state: STATE_INIT
   },
 
   initialize: function () {
@@ -42,12 +40,12 @@ var VisModel = Backbone.Model.extend({
 
   setOk: function () {
     this.trigger('ok');
-    this.set('state', STATES.OK);
+    this.set('state', STATE_OK);
   },
 
   done: function (callback) {
     this.once('ok', function () {
-      if (this.get('state') === STATES.INIT) {
+      if (this.get('state') === STATE_INIT) {
         callback(this);
       }
     }, this);
@@ -56,12 +54,12 @@ var VisModel = Backbone.Model.extend({
 
   setError: function (error) {
     this.trigger('error', error);
-    this.set('state', STATES.ERROR);
+    this.set('state', STATE_ERROR);
   },
 
   error: function (callback) {
     this.once('error', function (error) {
-      if (this.get('state') === STATES.INIT) {
+      if (this.get('state') === STATE_INIT) {
         callback(error);
       }
     }, this);
--- a/src/vis/vis.js
+++ b/src/vis/vis.js
@@ -15,12 +15,19 @@ var LayersCollection = require('../geo/map/layers');
 var AnalysisPoller = require('../analysis/analysis-poller');
 var Layers = require('./vis/layers');
 
+var STATES = {
+  INIT: 'init', // vis hasn't been sent to Windshaft
+  OK: 'ok', // vis has been sent to Windshaft and everything is ok
+  ERROR: 'error' // vis has been sent to Windshaft and there were some issues
+};
+
 var VisModel = Backbone.Model.extend({
   defaults: {
     loading: false,
     https: false,
     showLegends: false,
-    showEmptyInfowindowFields: false
+    showEmptyInfowindowFields: false,
+    state: STATES.INIT
   },
 
   initialize: function () {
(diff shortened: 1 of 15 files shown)


alonsogarciapablo:
@viddo, I pushed https://github.com/CartoDB/cartodb.js/pull/1324/commits/1163bfc318b332a92228223b41e8bd4da3dccc51, which implements something similar for analysis models.

src/vis/vis.js
```diff
@@ -33,13 +40,31 @@ var VisModel = Backbone.Model.extend({
     this.overlaysCollection = new Backbone.Collection();
   },
 
+  setOk: function () {
+    this.trigger('ok');
```
viddo: (this is general feedback, I write here instead of the PR to have a threaded conversation about this particular issue):

It's not clear to me what _ok_ really means, it feels a bit too generic/unclear/ambiguous. What do you think about clarifying it, depending on the model?  

Also, why not remove the duplication of state data (event+state attr), in favor of just having one data point? This also have some benefits:
- better model the intent, that these "states" are really mutually exclusive
- get rid of manual triggers
- be able to get state after (last) event happened

```js
vis.on(`change:state`, …
vis.get('state'); // => 'error' 
```

src/vis/vis.js
```diff
@@ -15,12 +15,19 @@ var LayersCollection = require('../geo/map/layers');
 var AnalysisPoller = require('../analysis/analysis-poller');
 var Layers = require('./vis/layers');
 
+var STATES = {
+  INIT: 'init', // vis hasn't been sent to Windshaft
+  OK: 'ok', // vis has been sent to Windshaft and everything is ok
+  ERROR: 'error' // vis has been sent to Windshaft and there were some issues
```
viddo: Why not just use separate variables instead (e.g. `STATE_INIT`) which has the benefit that non-usages and typos are caught by the test suite (through the linter)? The only motivation I can see for using an object would be if the vars would be exposed outside the file's own scope.

viddo:
Except for the questions, 👍

src/vis/vis.js
alonsogarciapablo: :+1: Makes sense.

src/vis/vis.js
alonsogarciapablo: > It's not clear to me what ok really means, it feels a bit too generic/unclear/ambiguous. What do you think about clarifying it, depending on the model?

Agree that `ok` is a bit ambiguous... What I'm trying to say with `ok` and `error` is that the "resource" was or wasn't "created/updated" successfully on the Maps API. They would be the equivalent to Backbone's `sync` and `error`. I can rename `ok` to `sync`... Is that clearer?

> Also, why not remove the duplication of state data (event+state attr), in favor of just having one data point? This also have some benefits:

The problem I see with that is that we'd need to things like:

```
vis.on('change:state', function (model, state) {
  if (state === 'error') {
    // do something
  } else if (state === 'ok') {
    // do something else
  }
});
```

I feel like the following is simpler (that's why I went with custom events):

```
vis.on('sync', function () { ... });
vis.on('error', function () { ... });
```

alonsogarciapablo:
Did some changes after my last CR:

- Layer models can now trigger `error` and `ok` events too. So far we have:
  ```
  vis.on('ok', function () { ... });
  vis.on('error', function (error) { ... });
  analysis.on('ok', function () { ... });
  analysis.on('error', function (error) { ... });
  layer.on('ok', function () { ... });
  layer.on('error', function (error) { ... });
  ```

- Callbacks for `error` events will receive an instance of [src/windshaft/error.js](https://github.com/CartoDB/cartodb.js/blob/f253f074ae460d480b10161f050557111c82a961/src/windshaft/error.js) with more info than the error message that we had before:

  ```
  error.message
  error.type  // eg: 'turbo-carto'
  error.context // eg: { "selector": "#layer", "source": { ... } }
  error.layerId  // if error is linked to a layer
  error.analysisId // if error is linked to an analysis
  error.isLayerError()
  error.isAnalysisError()
  ```

@viddo mind CRing the last commits after your latest comment? Thx
@xavijam do you think this should be enough for the editor? are you missing something? Thx

xavijam:
`isWidgetError()`

xavijam:
In any case, for a first version it is great. We will need to iterate for sure, but looks promising :).

alonsogarciapablo:
Dataview errors are not categorized by the Maps API yet

alonsogarciapablo:
@gfiorav, If you want to test this...

1. Open the inspector an look for the instantiation request to the Maps API (url should be something like: http://cdb.localhost.lan:8181/api/v1/map/named/tpl_31422150_3797_11e6_a0d1_080027880ca6?stat_tag=31422150-3797-11e6-a0d1-080027880ca6).

2. The response should look something like:
  ```
  {
    "layergroupid": "cdb@da78e72c@51ce037dfcde5516c2a05f13ad36db78:1466604167999",
    "metadata": {
      "layers": [
        {
          "type": "http",
          "id": "http-layer0",
          "meta": {
            "stats": []
          }
        },
        {
          "type": "mapnik",
          "id": "THE_LAYER_ID",
          "meta": {
            "stats": [],
            "cartocss": "..."
          }
        }
      ],
      "dataviews": {},
      "analyses": []
   },
    "last_updated": "2016-06-22T14:02:47.999Z"
  }
  ```

3. Verify that the "id" of the "mapnik" layer (whose value is "LAYER_ID" in this case) matches the ID of the layer in the named map template.

Ping me if you have questions. Thanks

gfiorav:
I'm a bit lost @alonsogarciapablo , did you mean to post that in https://github.com/CartoDB/cartodb/pull/8340?

In that case, yeah, I have an automated test that checks that of course. My question was towards integration, i.e, making sure the tiler likes these templates.

alonsogarciapablo:
> did you mean to post that in CartoDB/cartodb#8340?
No, I was explaining you the steps to manually verify that  it works as expected (in case you want to verify in staging for example).

alonsogarciapablo:
> did you mean to post that in CartoDB/cartodb#8340?

Actually, yes! I should have posted this in CartoDB/cartodb#8340. Moving it.

xavijam:
Already tested in the builder and works like a charm :)

alonsogarciapablo:
Ole! 😄

PR #1326, opened 2016-06-27
Tooltip issue

xavijam:
- Fixing problem rendering false values like 0, false, ...

Fixes https://github.com/CartoDB/cartodb/issues/8251

also, map reloads after change fields in the tooltip


matallo:
🚀

Issue #1226, opened 2016-04-06
Include auth_token param when instantiating password protected maps


alonsogarciapablo:
viz.json includes  a section of `auth_tokens` for maps that are password protected. This token (the first one) should be used in the request to the Maps API as the `auth_token` parameter.

Right now request to the Maps API is returning `Unauthorized template instantiation` for password protected maps.

PR #1330, opened 2016-07-04
Requests to Windshaft include auth_token param

alonsogarciapablo:
Fixes https://github.com/CartoDB/cartodb.js/issues/1226.
Fixes https://github.com/CartoDB/cartodb/issues/7859.

When the vizjson has an auth_token (this will be true for maps that are password protected), the following requests to Windshaft will use it in the query string:

- Requests to instantiate the map
- Requests to get the tiles
- Requests to get the grids
- Requests for dataviews
- Requests for analyses

NOTE: If apiKey and authToken are present these (☝️ ) requests will only include the api_key param.

@xavijam PTAL! Thank you! 💋


xavijam:
🇪🇸

PR #1331, opened 2016-07-04
Fixes number of bins not being sent to tiler

fdansv:
We were just omitting the bins attribute because we were checking if it was a number, when the attribute is stored as a string.

@rochoa


rochoa:
Why don't we store it as a number?

PR #1333, opened 2016-07-04
Bind overlays collection to make the view responsive to changes.

nobuti:
This PR allows the view to show/hide overlays when builder says so.

### DIFF
--- a/src/vis/vis-view.js
+++ b/src/vis/vis-view.js
@@ -33,7 +33,7 @@ var Vis = View.extend({
 
     this.model.once('load', this.render, this);
     this.model.on('invalidateSize', this._invalidateSize, this);
-    this.model.overlaysCollection.on('add remove', this._addOverlays, this);
+    this.model.overlaysCollection.on('add remove', this._resetOverlays, this);
 
     this.overlays = [];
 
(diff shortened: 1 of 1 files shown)


xavijam:
:+1:

src/vis/vis-view.js
```diff
@@ -33,6 +33,8 @@ var Vis = View.extend({
 
     this.model.once('load', this.render, this);
     this.model.on('invalidateSize', this._invalidateSize, this);
+    this.model.overlaysCollection.on('add remove', this._addOverlays, this);
```
alonsogarciapablo: Will `_addOverlays` "un-render" already rendered overlays?

I was expecting something like:

```
this.model.overlaysCollection.on('remove', this._removeOverlay, this);
```

nobuti: I did it that way because _addOverlays already cleans the overlays: https://github.com/CartoDB/cartodb.js/blob/v4/src/vis/vis-view.js#L224-L227

alonsogarciapablo: I see. Then I think we could rename it `this._resetOverlays` or something a bit more meaningful.

nobuti:
Are we ok @alonsogarciapablo? do I need to bump anything? I love bumping...

alonsogarciapablo:
👍

PR #1334, opened 2016-07-05
vis.instantiateMap supports success and error

alonsogarciapablo:
When `createVis` is used with `skipMapOptions:true` option, vis will have to be instantiated as follows:

```
vis.on('load', function () {
  vis.instantiateMap({
    success: function () { ... },
    error: function () { ... }
  });
});
```

@xavijam since you're already in the loop... will you give me a 👍? thx!


xavijam:
:+1:

xavijam:
🇺🇸

xavijam:
🏄

PR #1335, opened 2016-07-05
Hello logo

xavijam:
screen shot 2016-07-05 at 14 19 38

Fixes https://github.com/CartoDB/cartodb/issues/8473.
CR: @alonsogarciapablo 
cc @saleiva @urbanphes @piensaenpixel

### DIFF
--- a/src/api/create-vis.js
+++ b/src/api/create-vis.js
@@ -27,16 +27,18 @@ var createVis = function (el, vizjson, options) {
   options = _.defaults(options || {}, DEFAULT_OPTIONS);
 
   var isProtocolHTTPs = window && window.location.protocol && window.location.protocol === 'https:';
-
+console.log(options);
   var visModel = new VisModel({
     title: options.title || vizjson.title,
     description: options.description || vizjson.description,
     apiKey: options.apiKey,
     authToken: vizjson.auth_tokens && vizjson.auth_tokens.length && vizjson.auth_tokens[0],
     showLegends: options.legends === true || vizjson.legends === true,
+    showLogo: options.logo === undefined || options.logo === true || vizjson.logo === true,
     showEmptyInfowindowFields: options.show_empty_infowindow_fields === true,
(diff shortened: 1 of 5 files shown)


xavijam:
@nobuti, using the overlays in order to render the logo, but the name has changed, let's get rid of the cartodb prefix. So just **logo**, will it fit with your map options behaviour?

nobuti:
Yes! 🚀

src/api/create-vis.js
```diff
@@ -34,6 +34,7 @@ var createVis = function (el, vizjson, options) {
     apiKey: options.apiKey,
     authToken: vizjson.auth_tokens && vizjson.auth_tokens.length && vizjson.auth_tokens[0],
     showLegends: options.legends === true || vizjson.legends === true,
+    showLogo: options.logo === undefined || options.logo === true || vizjson.logo === true,
```
alonsogarciapablo: We could also add a default value for showLogo [here](https://github.com/CartoDB/cartodb.js/blob/66740c833adfeb8729e0bfe6966f953a17e5e27c/src/api/create-vis.js#L7). and remove the `options.logo === undefined`...

src/vis/vis-view.js
```diff
@@ -223,6 +223,14 @@ var Vis = View.extend({
       return overlay.order === null ? Number.MAX_VALUE : overlay.order;
     });
 
+    // if logo is not included, let add it :)
+    var overlayNames = _.pluck(overlays, 'type');
+    if (!_.contains(overlayNames, 'logo')) {
```
alonsogarciapablo: Take a look at [this part of the code where we're already adding new overlays to the vizjson object](https://github.com/CartoDB/cartodb.js/blob/v4/src/api/create-vis.js#L86). Perhaps worth adding this overlay there too? (just to have it in the same place)

xavijam:
I'll take those comments into account in other PR, don't worry @nobuti :).

PR #1337, opened 2016-07-05
Replaced `ok` and `error` events by `error` attribute on vis, layer and analyses models

alonsogarciapablo:
The "API" for errors that we implemented initially (see https://github.com/CartoDB/cartodb.js/pull/1324#issuecomment-229647295) is not working properly because we might need to access the errors after the `error` event has been triggered (eg: right after the editor has been loaded).

This PR replaces `ok` and `error` events triggered by vis, layers and analyses by an `error` attribute that we can 'query' at any time. (this was suggested by @viddo at some point).

@viddo 👍 ?

### DIFF
--- a/src/analysis/analysis-model.js
+++ b/src/analysis/analysis-model.js
@@ -39,11 +39,11 @@ module.exports = Model.extend({
   },
 
   setOk: function () {
-    this.trigger('ok');
+    this.unset('error');
   },
 
   setError: function (error) {
-    this.trigger('error', error);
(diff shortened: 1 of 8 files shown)


viddo:
Looks legit 👍

alonsogarciapablo:
Added some bonus feature to this PR 😋 : now the `vis` object triggers a `reload` event every time CartoDB.js create a new instance of the map. @nobuti please take a look at the last too commits. Thx!

PR #1350, opened 2016-07-12
Fix infowindow issues 

alonsogarciapablo:
Fixes https://github.com/CartoDB/cartodb/issues/8558

- Infowindow was only being updated when it had been clicked before and it was visible.
- Internal collection of layers of the "LayerGroupModel" wasn't being updated correctly (the order wasn't right) when new layers were added/removed.

cc: @matallo


matallo:
neat!

PR #1351, opened 2016-07-13
Ensure new requests to Windshaft can be performed when request limit has been reached

alonsogarciapablo:
Ensure that new requests to Windshaft can be performed after the max number of subsequent request has been reached, when the request is different.

@xavijam 👍 ?

### DIFF
--- a/src/windshaft/request-tracker.js
+++ b/src/windshaft/request-tracker.js
@@ -27,6 +27,11 @@ RequestTracker.prototype.reset = function () {
   this._numberOfRequests = 0;
 };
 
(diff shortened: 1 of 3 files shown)


src/windshaft/request-tracker.js
```diff
@@ -27,6 +27,11 @@ RequestTracker.prototype.reset = function () {
   this._numberOfRequests = 0;
 };
 
+RequestTracker.prototype.canRequestBePerformed = function (request) {
```
xavijam: :+1:

xavijam:
:es:

PR #1355, opened 2016-07-14
Add jasmine CLI test runner

viddo:
To make spec bundle smaller and have quicker feedback on tests that doesn’t require a browser window/DOM.

Basically same changes that was done on cartodb repo, https://github.com/CartoDB/cartodb/pull/7997 

@alonsogarciapablo ok?


grunt/tasks/watch.js
```diff
@@ -3,8 +3,20 @@
  *
  */
 module.exports = {
-  task: function() {
+  task: function () {
     return {
+      jasmine: {
```
alonsogarciapablo: This is great to get some instant feedback when writing/changing tests, but I personally find it not so useful when doing something else (eg: changing tests that are not executed by the `npm-test` task). It makes things a bit slower (even if tests are very fast). I see there's also `npm run test-watch`, so perhaps we can remove this?

viddo: Sure, I'll 🔥 it then

alonsogarciapablo:
Thanks for opening this PR @viddo! 

The only thing I don't like much about this is that we end having `test/spec` and `spec/specs` which might be a bit confusing... Perhaps we can have `spec/unit` and `spec/views` or something along those lines?

viddo:
Agree, although the (remaining) tests in `test/spec` dir are not only for views per se, so I'd leave that dir name as it is.

alonsogarciapablo:
What directory structure are you suggesting then?

viddo:
`spec/unit` (as you proposed, already changed), and leave test/spec as it is.

alonsogarciapablo:
> spec/unit (as you proposed, already changed), and leave test/spec as it is.

I was actually suggesting having all tests under one main folder and then having a subfolder for "unit" vs "browser" tests, so that anyone coming to CartoDB.js will surely know where to add new tests...

viddo:
Ah, yes, I misunderstood then. Agree, updated again 😄

alonsogarciapablo:
👍 Thanks!

viddo:
Good to go?

alonsogarciapablo:
:shipit:

PR #1356, opened 2016-07-14
Fix dataviews pointing to a layer that has a source

alonsogarciapablo:
@javisantana this fixes the issue you encountered. Basically, if a dataview "points" to a layer, and that layer doesn't have `sql`, it means that it's pointing to an analysis node, and the dataview should point to the layers's source in the payload that is sen't to the Maps API. 👍 ?

### DIFF
--- a/src/dataviews/dataview-model-base.js
+++ b/src/dataviews/dataview-model-base.js
@@ -228,7 +228,7 @@ module.exports = Model.extend({
   getSourceId: function () {
     // Dataview is pointing to a layer that has a source, so its
     // source is actually the the layers's source
-    if (this.hasLayerAsSource() && this.layer.get('source')) {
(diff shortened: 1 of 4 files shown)


src/dataviews/dataview-model-base.js
```diff
@@ -226,9 +226,22 @@ module.exports = Model.extend({
   },
 
   getSourceId: function () {
+    // Dataview is pointing to a layer that has a source, so its
+    // source is actually the the layers's source
+    if (this.hasLayerAsSource() && this.layer.get('source')) {
```
viddo: `this.layer.has('source')` perhaps is a tad clearer?

alonsogarciapablo: :+1:

javisantana:
:es:

PR #1359, opened 2016-07-15
1357 fix dataviews analysis states

### LINKED ISSUE #1357: Dataviews should be in loading state while its related analysis is not ready
Related to https://github.com/CartoDB/cartodb/issues/8365, the dataviews should also be in `loading` when the related analysis doesn't have status `ready` or `failed`

viddo:
Fixes #1357 

_WIP_

### DIFF
--- a/src/dataviews/dataview-model-base.js
+++ b/src/dataviews/dataview-model-base.js
@@ -48,9 +48,8 @@ module.exports = Model.extend({
     attrs = attrs || {};
     opts = opts || {};
 
-    if (!opts.map) {
-      throw new Error('map is required');
-    }
+    if (!opts.map) throw new Error('map is required');
+    if (!attrs.source) throw new Error('source is a required attr');
 
     if (!attrs.id) {
       this.set('id', this.defaults.type + '-' + this.cid);
@@ -78,7 +77,7 @@ module.exports = Model.extend({
 
   _initBinds: function () {
     this.listenTo(this.layer, 'change:visible', this._onLayerVisibilityChanged);
-    this.listenTo(this.layer, 'change:source', this._onLayerVisibilityChanged);
+
     var layerDataProvider = this._getLayerDataProvider();
     if (layerDataProvider) {
       this.listenToOnce(layerDataProvider, 'dataChanged', this._onChangeBinds, this);
@@ -90,6 +89,7 @@ module.exports = Model.extend({
         });
       });
     }
+
     if (this.filter) {
       this.listenTo(this.filter, 'change', this._onFilterChanged);
     }
@@ -102,16 +102,18 @@ module.exports = Model.extend({
     this._removeExistingAnalysisBindings();
     this._analysis = this._analysisCollection.get(this.getSourceId());
     if (this._analysis) {
-      this._analysis.on('change:status', this._onAnalysisChange, this);
+      this._analysis.on('change:status', this._onAnalysisStatusChange, this);
     }
   },
 
   _removeExistingAnalysisBindings: function () {
     if (!this._analysis) return;
-    this._analysis.off('change:status', this._onAnalysisChange, this);
+
+    this.layer.off('change:status', this._onAnalysisStatusChange, this);
+    this._analysis.off('change:status', this._onAnalysisStatusChange, this);
   },
 
-  _onAnalysisChange: function (analysis, status) {
+  _onAnalysisStatusChange: function (analysis, status) {
     if (analysis.isLoading()) {
       this._triggerLoading();
     } else if (analysis.isFailed()) {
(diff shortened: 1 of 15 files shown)


viddo:
@alonsogarciapablo this is what I have in mind at the moment, would like some early feedback before I continue more, in case I'm overlooking something or if you think I should solve this in another way?

src/dataviews/dataview-model-base.js
```diff
@@ -76,6 +78,7 @@ module.exports = Model.extend({
 
   _initBinds: function () {
     this.listenTo(this.layer, 'change:visible', this._onLayerVisibilityChanged);
+    this.listenTo(this.layer, 'change:source', this._onLayerVisibilityChanged);
```
alonsogarciapablo: Do we need this?

alonsogarciapablo:
Current approach seems good to me 👍. Keep going!

src/dataviews/dataview-model-base.js
viddo: Yes, it's my understanding by looking at the code and take into consideration that the source will change from time to time (e.g. adding/remove nodes in the builder). 
So when the head node of the layer (i.e. source) changes we would need to make sure [the correct analysis](https://github.com/CartoDB/cartodb.js/blame/1357-fix-dataviews-analysis-states/src/dataviews/dataview-model-base.js#L103) is listened to.

alonsogarciapablo: `this._onLayerVisibilityChanged` ?

alonsogarciapablo: Aren't we taking care of that [here](https://github.com/CartoDB/cartodb.js/pull/1359/files/ab4d43799e3ba602de36efbe15339ecbf76f0f2f#diff-805364e485d740e26e33839755aab753R97)?

viddo: Hmm not sure I follow you… do you mean that the visibility of the layer should influence what `getSourceId` returns? E.g. if layer is not visible return the own source instead?

viddo: Argh, misread… yeah that's wrong, I'll fix that

viddo:
Done, @alonsogarciapablo please have a look again. I've been making some quick tests (using `npm link` on DI + cdb in the builder and network conditions set), and it seems to be working now.

src/dataviews/dataview-model-base.js
```diff
@@ -87,9 +89,45 @@ module.exports = Model.extend({
...
+    this._analysis = this._analysisCollection.get(this.getSourceId());
+    if (this._analysis) {
+      this._analysis.on('change:status', this._onAnalysisStatusChange, this);
+    }
+  },
+
+  _removeExistingAnalysisBindings: function () {
+    if (!this._analysis) return;
+
+    this.layer.off('change:status', this._onAnalysisStatusChange, this);
```
alonsogarciapablo: Do layers have a `status` attribute?

src/dataviews/dataview-model-base.js
```diff
@@ -225,6 +263,7 @@ module.exports = Model.extend({
     return this.getSourceId() === this.layer.get('source');
   },
 
+  // @return might return undefined if there's no source
```
alonsogarciapablo: I don't think this is true, given that we're always setting a source in the dataviewFactory.

test/spec/dataviews/dataview-model-base.spec.js
```diff
@@ -536,4 +568,51 @@ describe('dataviews/dataview-model-base', function () {
...
+        loading: this.loadingSpy = jasmine.createSpy('loading'),
+        error: this.errorSpy = jasmine.createSpy('failed')
+      });
+    });
+
+    sharedTestsForAnalysisEvents();
+
+    describe('when changed source', function () {
+      beforeEach(function () {
+        this.model.set('source', {id: 'a0'});
```
alonsogarciapablo: Wasn't source already `{id: 'a0'}`? (https://github.com/CartoDB/cartodb.js/pull/1359/files#diff-f3d05477ce22032bf6eedd3d279090f1R18)

alonsogarciapablo:
Left two minor comments! Other than those two... 👏 👏 👏

src/dataviews/dataview-model-base.js
viddo: no you're right…

src/dataviews/dataview-model-base.js
viddo: Right, this comment was written before I added the guard in the initialize too. I'll 🔥

test/spec/dataviews/dataview-model-base.spec.js
viddo: Considering the source is an _object_ and not a value we're technically creating a new object here, so the equality check would "fail" and it would set this as a "new' value and thus trigger an event.

Issue #1345, opened 2016-07-11
Widgets using output of an analysis doesn't get refreshed when filtering in another widget


viddo:
@saleiva commented on [Fri Jul 08 2016](https://github.com/CartoDB/cartodb/issues/8710)

But they get refreshed after panning the map.
Test in the map above in the formula widget
https://team.carto.com/u/ernestomb/builder/a89de72e-3fa4-11e6-80c4-0e674067d321/embed?state=%7B%22map%22%3A%7B%22center%22%3A%5B40.419769381446194%2C-3.6927795410156246%5D%2C%22zoom%22%3A12%7D%2C%22widgets%22%3A%7B%22c7ea4ffc-82b0-4cc7-8e4e-abc9fa9ab63f%22%3A%7B%22acceptedCategories%22%3A%5B%22L5%22%5D%7D%7D%7D

---

@viddo commented on [Mon Jul 11 2016](https://github.com/CartoDB/cartodb/issues/8710#issuecomment-231670624)

Moving and reassigning, since this should be handled internally in cartodb.js, by updating the dataviews appropriately.

alonsogarciapablo:
I'm looking into this... Some findings so far:

1. the formula widget (Total Population around) is linked to an analysis with id "b2":

```
{
  "id": "b2",
  "type": "data-observatory-measure",
  "params": {
    "source": {
      "id": "b1",
      "type": "buffer",
      "params": {
        "source": {
          "id": "a0",
          "type": "source",
          "params": {
            "query": "SELECT * FROM madrid_metro_stations"
          }
        },
        "radius": 1000
      }
    },
    "final_column": "pop",
    "segment_name": "es.ine.t1_1",
    "percent": false
  }
}
```

2. the other two widgets (Metro Line and Metro Station) are linked to "a0":

```
{
  "id": "a0",
  "type": "source",
  "params": {
    "query": "SELECT * FROM madrid_metro_stations"
  }
}
```

Until now, widgets/dataviews were only refreshed if [this was true](https://github.com/CartoDB/cartodb.js/blob/v4/src/dataviews/dataview-model-base.js#L169-L170). In this case, the formula widget is linked to a different layer than the other two widgets, and that's why widgets are not "connected".

We should change things to:

1. Work with source IDs instead of layer IDs.
2. Take into account that sources are now related... eg: "b2" depends on "a0", so filtering "a0" affect widgets that depend on "b2".

alonsogarciapablo:
[...]

PR #1360, opened 2016-07-15
Re-fetch dataviews when source is affected by some filtering

alonsogarciapablo:
Hopefully fixes https://github.com/CartoDB/cartodb.js/issues/1345.

Imagine the following situation, where there's a layer with 2 analysis and there're 4 widgets:

```
Layer A:     <- widget3
  - a2       <- widget2
  - a1       <- widget1
  - a0       <- widget0
```

If the filter of widget0 changes, widget1, widget2 and widget3 need to fetch data again (because a0 affects a1 and a2). But, if the filter of widget1 changes, only widget1, widget2 and widget3 will need to fetch data again...

In summary: dataviews whose source is affected by the source of the dataview that caused the map to be re-instantiated, should fetch again.

@xavijam :+1: ?

### DIFF
--- a/src/dataviews/dataview-model-base.js
+++ b/src/dataviews/dataview-model-base.js
@@ -48,9 +48,8 @@ module.exports = Model.extend({
     attrs = attrs || {};
     opts = opts || {};
 
-    if (!opts.map) {
-      throw new Error('map is required');
-    }
+    if (!opts.map) throw new Error('map is required');
+    if (!opts.analysisCollection) throw new Error('analysisCollection is required');
 
     if (!attrs.id) {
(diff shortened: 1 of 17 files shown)


xavijam:
:es:, just a question:

Would this change break compatibility with previous v4 version (the one we have published for partners)?

alonsogarciapablo:
> Would this change break compatibility with previous v4 version (the one we have published for partners)?

Not AFAIK. This PR only makes dataviews fetch when they have to. It doens't have any changes to the API, etc.

xavijam:
:+1:

PR #1361, opened 2016-07-19
Fix attributions

alonsogarciapablo:
Closes https://github.com/CartoDB/cartodb/issues/8959.

Events were being bound every time `map.instantiateMap` was called and this was causing:

1. `_updateAttributions` not being called right after the layers are reseted.
2. event callbacks probably being executed multiple times every time `map.instantiateMap` was called.

Also, `map.reload` does nothing until `map.instantiateMap`has been called at least once now.

@xavijam 👍 or 👎 ?

### DIFF
--- a/src/geo/map.js
+++ b/src/geo/map.js
@@ -60,13 +60,72 @@ var Map = Model.extend({
       });
     }
 
+    this._instantiateMapWasCalled = false;
+
     // This method is declared here so that we can spyOn _.debounce
     // in the tests that depend on this to work
     this.reload = _.debounce(function (options) {
-      options = options || {};
-      options = _.pick(options, 'sourceId', 'forceFetch', 'success', 'error');
-      this._windshaftMap.createInstance(options);
+      if (this._instantiateMapWasCalled) {
+        options = options || {};
+        options = _.pick(options, 'sourceId', 'forceFetch', 'success', 'error');
+        this._windshaftMap.createInstance(options);
+      }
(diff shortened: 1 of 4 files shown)


src/geo/map.js
```diff
@@ -60,13 +60,72 @@ var Map = Model.extend({
...
+
+    if (this._dataviewsCollection) {
+      // When new dataviews are defined, a new instance of the map needs to be created
+      this.listenTo(this._dataviewsCollection, 'add', _.debounce(this._onDataviewAdded.bind(this), 10));
+    }
+  },
+
+  _onLayersResetted: function () {
+    if (this.layers.size() >= 1) {
+      this._adjustZoomtoLayer(this.layers.models[0]);
```
xavijam: this.layers.at(0)?

src/geo/map.js
```diff
@@ -60,13 +60,72 @@ var Map = Model.extend({
...
+    this.layers.bind('change:attribution', this._updateAttributions, this);
+
+    if (this._dataviewsCollection) {
+      // When new dataviews are defined, a new instance of the map needs to be created
+      this.listenTo(this._dataviewsCollection, 'add', _.debounce(this._onDataviewAdded.bind(this), 10));
+    }
+  },
+
+  _onLayersResetted: function () {
+    if (this.layers.size() >= 1) {
```
xavijam: if (this.layers.size()) {
…
}

xavijam:
Not much to add :es: when tests have passed 👅

PR #1365, opened 2016-07-21
Wait until map has been instantiated explicitly before binding events

alonsogarciapablo:
@xavijam

### DIFF
--- a/src/geo/map.js
+++ b/src/geo/map.js
@@ -173,7 +173,6 @@ var Map = Model.extend({
     this.layers.bind('reset', this._onLayersResetted, this);
     this.layers.bind('add', this._onLayerAdded, this);
     this.layers.bind('remove', this._onLayerRemoved, this);
-    this.layers.bind('change:attribution', this._updateAttributions, this);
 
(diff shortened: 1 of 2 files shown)


xavijam:
Tests failing

src/geo/map.js
```diff
@@ -72,46 +72,10 @@ var Map = Model.extend({
...
     }.bind(this), this.RELOAD_DEBOUNCE_TIME);
 
-    this._initBinds();
-  },
-
-  _initBinds: function () {
-    this.layers.bind('reset', this._onLayersResetted, this);
-    this.layers.bind('add', this._onLayerAdded, this);
-    this.layers.bind('remove', this._onLayerRemoved, this);
+    this.layers.bind('reset', this._updateAttributions, this);
```
xavijam: I'd remove this binds if we set after map instantiation, no?

alonsogarciapablo: The thing is that layers are actually reset before the map is instantiated... and attributions need to be updated. I'll think about a better approach right after we fix this.

PR #1366, opened 2016-07-21
Better error handling when trying to instantiate the map

alonsogarciapablo:
@xavijam 👍 ?

### DIFF
--- a/src/windshaft/map-base.js
+++ b/src/windshaft/map-base.js
@@ -60,17 +60,26 @@ var WindshaftMap = Backbone.Model.extend({
   createInstance: function (options) {
     options = options || {};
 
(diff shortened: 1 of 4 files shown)


src/windshaft/map-base.js
```diff
@@ -60,17 +60,26 @@ var WindshaftMap = Backbone.Model.extend({
   createInstance: function (options) {
     options = options || {};
 
-    var payload = this.toJSON();
-    var params = this._getParams();
+    try {
```
xavijam: try and catch??

alonsogarciapablo: yes, try and catch.

xavijam:
I like the approach so :es:

PR #1367, opened 2016-07-22
Add layers to the map when .render is called to allow listening to 'newLayerView' event

alonsogarciapablo:
Hopefully fixes https://github.com/CartoDB/cartodb/issues/9064.

We're trying to listen to `newLayerView` event, after the events had been triggered, and that was preventing the Blue Bar from being displayed while tiles were being loaded... 

@nobuti could you please take a look? thx!


nobuti:
👍

PR #1369, opened 2016-07-26
Remove map.reload, in favor of vis.reload

alonsogarciapablo:
To remove some indirection and (hopefully) troubleshoot failed map instantiations easier.

We had this relationships before:

```
  Layers (CartoDBLayer and TorqueLayer) -> Map#reload
  Analysis -> Map#reload
  Dataviews -> Map#reload
  Map#reload -> Vis#reload
```

This PR removes Map#reload and makes objects reload the vis directly:

```
  Layers (CartoDBLayer and TorqueLayer) -> Vis#reload
  Analysis -> Vis#reload
  Dataviews -> Vis#reload
  Map -> Vis#reload
```

As a follow up, I think it'd make sense to centralice all calls to `vis.reload` in a single class, so that it's easier to control when the map gets instantiated... 

@viddo what do you think? Thx!

@viddo what do you think?

### DIFF
--- a/src/analysis/analysis-factory.js
+++ b/src/analysis/analysis-factory.js
@@ -7,15 +7,15 @@ var AnalysisFactory = function (opts) {
   if (!opts.analysisCollection) {
     throw new Error('analysisCollection option is required');
   }
-  if (!opts.map) {
-    throw new Error('map option is required');
+  if (!opts.vis) {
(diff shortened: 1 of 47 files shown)


viddo:
Sounds reasonable and code looks legit 👌

PR #1370, opened 2016-07-28
Fix overlays

alonsogarciapablo:
`InfowindowView` depends on the `mapView`, but [the InfowindowManager wasn't able to get it from the `vis` object](https://github.com/CartoDB/cartodb.js/blob/7d1e91c82d284f18d61d10b92f267ca75b2adeea/src/vis/overlays.js#L287) (we needed the visView, not the vis model).

Also, I took the time to:

- Unify `src/vis/overlays.js` and `src/vis/vis/overlays.js` into `src/vis/overlays-factory.js`.
- Write some (very basic) tests to ensure different types of overlays can be constructed.

@xavijam can you please take a look, thanks!


xavijam:
Can't say no, sounds great :es:.

PR #1371, opened 2016-07-28
Support viz.json with flat layers + cleanup

### LINKED ISSUE #829: Support viz.json files with flat layers
New viz.json format is implemented in https://github.com/CartoDB/cartodb/pull/9118.

cc: @rochoa @javisantana

alonsogarciapablo:
Fixes #829.

The main goal of this PR is mainly to support viz.json files with flat layers, which is implemented [here](https://github.com/CartoDB/cartodb.js/pull/1371/files#diff-f99ed960498d5ea8717b3e74d54bfb01R331). It turns out that this was already supported but I prepared things so that we can get rid of our mechanism to flatten layers. If, for any chance, cartodb.js gets a viz.json where layers are not flat, it will still work. I'll remove this "fallback" (in https://github.com/CartoDB/cartodb.js/issues/1372) when we know everything is working fine: https://github.com/CartoDB/cartodb/pull/9118 has been released and we're sure no "old" viz.json files are being served.

Also:

- Deleted `src/api/create-layer.js` which is not being used and not working anyway.
- Unified `src/vis/layers` and `src/vis/vis/layers` into `src/vis/layers-factory`.
- `src/vis/layers-factory` doesn't provide a way to register new layer types anymore (not needed right now).

@viddo: looks legit? :wink:. Thanks!

cc: @javitonino @juanignaciosl


viddo:
Sounds/Looks legit 👍

PR #1376, opened 2016-08-03
A different way to show/hide layers

### LINKED ISSUE #1374: Change mechanism to show/hide layers
Preparation for https://github.com/CartoDB/Windshaft-cartodb/issues/550

Changes:

- [x] CartoDB.js sends all "CartoDB" layers to Windshaft when instantiating Anonymous Maps (instead of [only sending the visible ones](https://github.com/CartoDB/cartodb.js/blob/51b0c5d5085639f007809a1eb5a5d0a97538caf0/src/windshaft/anonymous-map.js#L23)).
- [x] CartoDB.js no longer sends ["The HACK"](https://github.com/CartoDB/cartodb.js/blob/51b0c5d5085639f007809a1eb5a5d0a97538caf0/src/windshaft/named-map.js#L8-L10) to Windshaft when instantiating Named Maps.
- [x] Improve how index of CartoDB (interactive) layers is calculated when requesting tiles, grids and attributes. Match the response from Windshaft and the collection of layers to calculate the index of each layer.
- When a layer is shown/hided:
  - [x] re-fetch tiles instead of reloading the whole map.
  - [x] re-fetch grids to enable/disable interaction of layers, instead of reloading the whole map.

alonsogarciapablo:
Closes #1374.

Changes:
  - (Hopefully) improved/simplified the way that "remote" layer indexes are calculated when generating URLs to fetch tiles, grids and attributes. We now use the response from Windshaft to get the real indexex of layer models.
  - When the visibility of a CartoDB layer changes:
    - Map is no longer instantiated (\0/).
    - Instead:
      - Tiles are re-fetched using the indexes of visible layers (unless all layers are hidden, in which case an empty GIF is requested).
      - Grids are re-fetched again (interactivity is calculated again).
   - Removed some stuff that wasn't in use right now along the way.
    
cc: @rochoa @dgaubert


rochoa:
I updated this PR with `v4` branch because it had some conflicts due #1377.

I resolved the conflict with `--ours` because map-base.js supports arrays in params, as it's no longer manipulating the params as a string. However, I had to also apply the same solution to `cartodb-layer-group.js` as per https://github.com/CartoDB/cartodb.js/pull/1376/commits/8008075ec3afd740fa3688ba87f0f5a3e1d16a16.

cc @javitonino as you were involved in previous `auth_token` discussion, just in case I'm missing something.

xavijam:
Adding myself to the loop.

src/geo/map-view.js
```diff
@@ -179,7 +179,7 @@ var MapView = View.extend({
   _removeLayer: function (layerModel) {
     var layerView = this._layerViews[layerModel.cid];
     if (layerModel.get('type') === 'CartoDB') {
-      if (this._cartoDBLayerGroup.layers.size() === 0) {
+      if (this.map.layers.getCartoDBLayers().length === 0) {
```
xavijam: We could change the method, now that we don't have DB.

src/windshaft/anonymous-map.js
```diff
@@ -13,23 +18,22 @@ var AnonymousMap = MapBase.extend({
   },
 
   _calculateLayersSection: function () {
-    return _.chain(this._getLayers())
+    return _.chain(this._getCartoDBAndTorqueLayers())
```
xavijam: Same here

xavijam:
Tested with @rochoa, working like a charm.

alonsogarciapablo:
Thanks @rochoa and @xavijam for taking care of this! 💋

PR #1377, opened 2016-08-03
[WIP] createVis supports authToken option

alonsogarciapablo:
Fixes https://github.com/CartoDB/cartodb.js/issues/1381.

cc: @javitonino... do you think we should leave the actual code (i.e: getting auth tokens from viz.json) as a fallback?

### DIFF
--- a/src/analysis/analysis-model.js
+++ b/src/analysis/analysis-model.js
@@ -32,7 +32,15 @@ module.exports = Model.extend({
       if (this.get('apiKey')) {
         url += '?api_key=' + this.get('apiKey');
       } else if (this.get('authToken')) {
-        url += '?auth_token=' + this.get('authToken');
+        var authToken = this.get('authToken');
+        if (authToken instanceof Array) {
+          var tokens = _.map(authToken, function (token) {
+            return 'auth_token[]=' + token;
+          });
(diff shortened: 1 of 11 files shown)


javitonino:
I think that we can remove it, as we will start passing the token via the embed at the same time.

We also need to support multiple auth tokens (as in cartodb.js 3), although I think passing an array as the authToken might work with this change (I'll check and report back).

javitonino:
So I've just tested this, and this only allows passing a single auth token parameter. Passing an array of auth tokens makes it post an incorrect request to the maps API: `&auth_token=%5B%22Fv7Wi4BsA6tCNaN7r2qGtg%22%2C%22b%22%5D` (JSON encoded array).

Can we make it work like in v3 (https://github.com/CartoDB/cartodb.js/blob/2e26d00a1fd40e385918ce05ae45b236bc61ebb1/src/geo/layer_definition.js#L165)?

javitonino:
Added support for multiple tokens. Also added authToken to the widgets data models, as they were not working with password protected/privately shared maps (also in production).

xavijam:
I see the pattern/method for auth_tokens repeated several times in the dataview models, but in any case :+1:

PR #1378, opened 2016-08-03
added searchbox as a ui function

csobier:
@oriolbx, this was created as a result of [Docs issue#942](https://github.com/CartoDB/docs/issues/942). I'm added your example code as an enhancement in the Specific UI Functions of the CARTO.js doc.
- Can you do a quick copyedit for review?
- Ideally, I would rather add a complete code example in the Docs instead of linking to a bl.ocks page that might change. If time allows, can you create an updated example that I can add and reference from the cartodb.js/examples folder?

### DIFF
--- a/doc/ui_functions.md
+++ b/doc/ui_functions.md
@@ -80,3 +80,102 @@ layer.leafletMap.viz.addOverlay({
   fields: [{ name: 'name' }]
   });
 {% endhighlight %}
+
+## Add a Search Button
+
+If you want to add a search box to find data points on your map, you can add a custom searchbox with CARTO.js. Search results move the map to the specified location, based on your data.
+
+Any `div` element can be used to specify a default block setting as part of the HTML code for your map. This section specifically describes how to include a searchbox on a map layer, and link it to your dataset using a SQL query.
+
+**Note:** This is different from the [`cartodb.createVis`](https://carto.com/docs/carto-engine/carto-js/api-methods/#cartodbcreatevis) `search` function, which adds a search control to query location by CARTO’s default location data service providers.
+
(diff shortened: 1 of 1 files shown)


alonsogarciapablo:
@csobier I wouldn't (personally) categorise this an "Specific UI Function"... For me, this is just an example showing how cartodb.js can be used to implement some search functionality. I know there are some ongoing conversations about how to improve the IA of our docs... Do you think it would make sense to have a section with some relevant examples?

Here're some other libraries that use this "pattern" too:

- [Google Maps API Code Samples](https://developers.google.com/maps/documentation/javascript/examples/)
- [Leaflet Tutorials](http://leafletjs.com/examples.html)
- [MapBox GL JS Examples](https://www.mapbox.com/mapbox-gl-js/examples/)

cc: @michellechandra

oriolbx:
@csobier it looks good to me. Ok, I will create an example in the cartodb.js/example folder.
I agree with @alonsogarciapablo, the example is just a way to show how to create a searchbox with CARTO.js and the CARTO SQL API but is not a "UI Function" or the only way to create one. Maybe it could be better to  have a section with use cases/examples with CARTO.js as @alonsogarciapablo says.

csobier:
Yeah, I was just trying to make it fit with the structure we have.  I'll have to bump this to a future milestone after we have time to redesign the IA.

csobier:
@alonsogarciapablo , can you confirm if this will be a documented feature in 4.0? If so, let's create a new issue for tracking, if not- close this issue?

csobier:
Closing via https://github.com/CartoDB/docs/issues/979#issuecomment-352457739

PR #1384, opened 2016-08-04
adding https flag to avoid cross domain issues

ohasselblad:
fixing the cross domain issues here: https://carto.com/docs/tutorials/named_maps/

cc @alonsogarciapablo

### DIFF
--- a/examples/tutorials/named-maps-example.html
+++ b/examples/tutorials/named-maps-example.html
@@ -50,7 +50,7 @@
 	        interactivity: "cartodb_id, name, pop_max"
 	       }]
 	     }
-	    })
+     }, {https: true})
 	    .addTo(map)
 	    .done(function(layer) {
 	      layer.getSubLayer(0).setInteraction(true);


alonsogarciapablo:
hello @ohasselblad! I'm not sure this is the way to solve this issue... This will make the example work when served from https://cartodb.github.io/cartodb.js/examples/tutorials/named-maps-example.html, but example will fail when served when served over http :(

My opinion, instead of linking to examples in the CartoDB.js repo, I think we should create a separate repo with some relevant examples (like [this other one](https://github.com/CartoDB/cartodb.js/pull/1378)) that we know will be linked from the docs/academy... This way we'll know where they are being served from (eg: github pages) and where they are being used (which is not possible right now)...

cc: @michellechandra @csobier @oriolbx what do you guys think?

oriolbx:
If we create a new repo with examples, then should we move all the examples from the cartodb.js/examples folder to the new repo or only specific ones? Or how should we decide which examples are relevant?

ohasselblad:
We're actually going to be migrating the tutorials in docs to academy this month, but I wanted to fix this example still. We could just add a message like "this tutorial is no longer maintained. See the Map Academy lesson ... instead" and then leave this example broken from tutorials.

alonsogarciapablo:
Merged @ohasselblad!

ohasselblad:
Thanks!

alonsogarciapablo:
Thank you!

PR #1385, opened 2016-08-08
add custom searchbox example

oriolbx:
Add example requested in https://github.com/CartoDB/cartodb.js/pull/1378

@alonsogarciapablo @csobier

### DIFF
--- a/examples/add_custom_searchbox.html
+++ b/examples/add_custom_searchbox.html
@@ -51,21 +51,22 @@
         // set the map options
         map = vis.getNativeMap().setView([41.390205, 2.154007],4);
        })
+        // When the search button is clicked
+        $('#searchButton').click(function () {
+          // store in the input variable the value from the input box
+          input = $('#ad').val();
+
+          // execute the CARTO SQL API
+          var sql = new cartodb.SQL({ user: 'documentation' });
+          // get the bounds of the geometry which has the same string value (in the name column) than the string value written in the input box
+          sql.getBounds("SELECT * FROM world_borders_1 where name ILIKE '" + input + "'").done(function (bounds) {
+            // sets the map view that contains the bounds defined by the getBounds method with the maximum zoom possible
+            map.fitBounds(bounds);
+          });
+        });
   }
 
-  // When the search button is clicked
-  $('#searchButton').click(function () {
-    // store in the input variable the value from the input box
-    input = $('#ad').val();
-
-    // execute the CARTO SQL API
-    var sql = new cartodb.SQL({ user: 'documentation' });
-    // get the bounds of the geometry which has the same string value (in the name column) than the string value written in the input box
-    sql.getBounds("SELECT * FROM world_borders_1 where name ILIKE '" + input + "'").done(function (bounds) {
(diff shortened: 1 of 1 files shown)


csobier:
Looks good, I believe we've changed our attribution code though?  Looking forward to using this one we establish our new examples format.

examples/add_custom_searchbox.html
```diff
@@ -0,0 +1,76 @@
...
+    var map = L.map('map', {
+      zoomControl: false,
+      center: [41.390205, 2.154007],
+      zoom: 4
+    });
+    // add basemap to our map object
+    L.tileLayer('http://{s}.basemaps.cartocdn.com/light_nolabels/{z}/{x}/{y}.png', {attribution: '&copy; <a href="http://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors, &copy; <a href="http://cartodb.com/attributions">CartoDB</a>'}).addTo(map);
+
+    // add CartoDB layer
+    cartodb.createLayer(map, 'https://documentation.carto.com/api/v2/viz/34dfd2e4-d62b-11e5-b5e7-0ea31932ec1d/viz.json').addTo(map)
```
alonsogarciapablo: I guess we could use `cartodb.createVis` here @oriolbx to make the example a bit easier:

```

cartodb.createVis(...)
.done(function(vis, layers) {
  ...
  var map = vis.getNativeMap();
  ...
  map.fitBounds(bounds);
  ...
});
```

oriolbx:
changed to createVis() method @csobier @alonsogarciapablo

examples/add_custom_searchbox.html
```diff
@@ -42,33 +42,30 @@
...
-      });
+    cartodb.createVis(map, layerUrl)
+      .done(function (vis,layer) {
+        // set the map options
+        map = vis.getNativeMap().setView([41.390205, 2.154007],4);
+       })
   }
+
+  // When the search button is clicked
+  $('#searchButton').click(function () {
```
alonsogarciapablo: This needs would need to be moved inside the callback for the `.done`, otherwise it wouldn't work if users try to search before the vis has been "created". Something like:

```
  function main () {
    
    cartodb.createVis('map', 'https://documentation.carto.com/api/v2/viz/34dfd2e4-d62b-11e5-b5e7-0ea31932ec1d/viz.json')
      .done(function (vis,layer) {
        // set the map options
        var map = vis.getNativeMap();
        map.setView([41.390205, 2.154007],4);

        // When the search button is clicked
        $('#searchButton').click(function () {
          // store in the input variable the value from the input box
          var query = $('#ad').val();

          // execute the CARTO SQL API
          var sql = new cartodb.SQL({ user: 'documentation' });

          // get the bounds of the geometry which has the same string value (in the name column) than the string value written in the input box
          sql.getBounds("SELECT * FROM world_borders_1 where name ILIKE '" + query + "'").done(function (bounds) {
            // sets the map view that contains the bounds defined by the getBounds method with the maximum zoom possible
            map.fitBounds(bounds);
          });
        });
      })
  }
```

alonsogarciapablo:
🚀

oriolbx:
sorry @alonsogarciapablo , silly me :(  Updated again

PR #1386, opened 2016-08-12
Check for error_message when update analysis node.

nobuti:
This PR fixes https://github.com/CartoDB/cartodb/issues/9415

When an analysis fails, the error message come wrapped inside the maps api request as `error_message` parameter:

screen shot 2016-08-12 at 10 38 41

When the model is updated, we look for this parameter to set error properly. Otherwise the error message will be empty. 

I know there were some intents of solving notifications from the maps api in the past, and I guess some of the solutions implemented aren't valid now. Not sure if this change of behaviour obey to a change in Whidshaft but this is the easy way to fix this today.

analysis-notification-error


xavijam:
:+1: :es:

PR #1389, opened 2016-08-18
infowindows: json array

matallo:
close https://github.com/CartoDB/cartodb/issues/9171

replicating the implementation of current infowindows, which support json arrays, right now we can now a infowindow is custom if it has `template` but `template_name` is empty

screen shot 2016-08-18 at 18 18 11

CR @xavijam @javisantana 

/cc @ramiroaznar


matallo:
damn, fixing those tests :/

src/geo/ui/infowindow-view.js
```diff
@@ -107,8 +111,12 @@ var Infowindow = View.extend({
 
       var data = this.model.get('content') ? this.model.get('content').data : {};
 
-      // Sanitized fields
-      fields = _.map(fields, this._sanitizeField, this);
+      // If a custom template is not applied, let's sanitize
+      // fields for the template rendering
+      if (!this.isCustomTemplate()) {
```
alonsogarciapablo: Aren't all templates "custom" in the context of cartodb.js now that we don't rely on `template_name` anymore?

matallo:
indeed they are, but these still work with _old_ templates where only template name (and no template attribute) is given

matallo:
aaand tests passing again

xavijam:
I can't add much more, all agree?

alonsogarciapablo:
Talked to @matallo about this solution. While I think it solves the problem we're currently seeing with old editor infowindow templates (CartoDB/cartodb#9171), I think we can fix this in a more generic way. Users might encounter this problem when using the API.

We found out that we're [sanitising fields here](https://github.com/CartoDB/cartodb.js/blob/666d671e006bff8faa10ecedf1113b738d4f1f79/src/geo/ui/infowindow-view.js#L275) and this is not working properly when a field has an object as it's value... My suggestion would be to try to handle this case gracefully instead of skipping sanitisation when something that the builder sets (today) (`template_name`) in some cartodb.js models is present.

@matallo I can help you with this, ofc! Thanks and... avant!

matallo:
@alonsogarciapablo finally, sorry it took too much (holidays, priorities, ...)

- added corresponding tests

/cc @ernesmb

src/geo/ui/infowindow-view.js
```diff
@@ -246,6 +246,16 @@ var Infowindow = View.extend({
     }
   },
 
+  _sanitizeValue: function (key, obj) {
+    var t = typeof (obj);
+
+    if (t !== 'object' || obj === null) {
```
alonsogarciapablo: Suggestion: we could use [`_.isObject`](http://underscorejs.org/#isObject) and [`isArray`](http://underscorejs.org/#isArray) here... 😄

matallo: ooh, totally, changing and publishing! ^^

PR #1390, opened 2016-08-22
Look for options only if not null.

nobuti:
This PR adds logo option backward compatibility: https://github.com/CartoDB/cartodb/issues/8674

cc @xavijam


xavijam:
:+1:

PR #1392, opened 2016-08-23
Require torque with github repository instead of version

xavijam:
@rochoa @javisantana

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "leaflet": "0.7.3",
     "mustache": "1.1.0",
     "perfect-scrollbar": "0.6.7",
-    "torque.js": "^2.15",
+    "torque.js": "CartoDB/torque#master",
     "underscore": "1.8.3",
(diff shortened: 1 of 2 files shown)


xavijam:
The problem using versioning here is, if we change something in torque.js, and we don't bump the package version, running `npm install` will not update anything from that dependency because it thinks it is the same.

rochoa:
IMO we should not use a branch for a dependency.

BTW why did you have to change those tests?/

xavijam:
Related with last changes in torque, testing that `Map {}` is present in the `cartocss`.
https://github.com/CartoDB/torque/commit/0e9184e02629c07a6323c8fc955e2760f6710226

xavijam:
If we don't use a branch, we should take care of the version, I'm ok with that, but if it has happened once, it could happen again.

rochoa:
I'm lost: what was the problem?

xavijam:
We have to publish the version to NPM.

PR #1394, opened 2016-08-25
Add/append api_key and auth_token in torque layers

rochoa:
This is a temporary workaround, it reuses api_key and auth_token
from existing _cartoDBLayerGroup model.

cc @alonsogarciapablo


xavijam:
:+1:

PR #1398, opened 2016-09-01
Try to create/instantiate maps using JSONP before using POST

### LINKED ISSUE #1396: Named maps JSONP instantiation
@rochoa commented on [Tue Aug 23 2016](https://github.com/CartoDB/cartodb/issues/9523)

Previous public pages use JSONP endpoint for intantiation:
https://documentation.cartodb.com/viz/2a1b745a-e219-11e4-a4f2-0e018d66dc29/public_map

New embeds use a POST to instantiate:
https://documentation.cartodb.com/builder/2a1b745a-e219-11e4-a4f2-0e018d66dc29/embed

JSONP endpoints are more cache friendly and we should try to use them when the params fit into the URL, and in case the payload is too big, then fallback to POST/CORS.

cc @xavijam 

---

@xavijam commented on [Tue Aug 23 2016](https://github.com/CartoDB/cartodb/issues/9523#issuecomment-241753011)

Well, map instantiation happens in CartoDB.js, but I really like to see the track here too :)

alonsogarciapablo:
Fixes #1396 and fixes CartoDB/cartodb#9523.

We're always (and mistakenly) enforcing creation/instantiation of maps using POST. This PR implements the following algorithm to determine when to use GET (JSONP) vs POST requests:

- Length of URL with encoded params + `config` param < 2033 -> `GET /api/v1/map?config=…&stat_tag=...`
- Length of URL with encoded params + `lzma` param < 2033 -> `GET /api/v1/map?lzma=…&stat_tag=...`
- Length of URL with encoded params + `lzma` param >= 2033 -> `POST /api/v1/map?stat_tag=…`

@rochoa would you mind taking a look?

### DIFF
--- a/src/windshaft/client.js
+++ b/src/windshaft/client.js
@@ -12,6 +12,9 @@ var validatePresenceOfOptions = function (options, requiredOptions) {
   }
 };
 
+var MAX_URL_LENGTH = 2033;
+var COMPRESSION_LEVEL = 3;
+
 /**
  * Windshaft client. It provides a method to create instances of maps in Windshaft.
  * @param {object} options Options to set up the client
@@ -26,9 +29,6 @@ var WindshaftClient = function (options) {
 
   this.url = this.urlTemplate.replace('{user}', this.userName);
 };
-
-WindshaftClient.prototype.MAX_URL_LENGTH = 2033;
-WindshaftClient.prototype.COMPRESSION_LEVEL = 3;
 
 WindshaftClient.prototype.instantiateMap = function (options) {
   if (!options.mapDefinition) {
@@ -61,11 +61,11 @@ WindshaftClient.prototype.instantiateMap = function (options) {
   };
 
   var encodedURL = this._generateEncodedURL(mapDefinition, params);
-  if (!this._isURLTooLong(encodedURL)) {
+  if (this._isURLValid(encodedURL)) {
     this._get(encodedURL, ajaxOptions);
   } else {
     this._generateCompressedURL(mapDefinition, params, function (compressedURL) {
-      if (!this._isURLTooLong(compressedURL)) {
+      if (this._isURLValid(compressedURL)) {
         this._get(compressedURL, ajaxOptions);
       } else {
         var url = this._getURL(params);
@@ -88,7 +88,7 @@ WindshaftClient.prototype._generateCompressedURL = function (payload, params, ca
     config: JSON.stringify(payload)
   });
 
-  LZMA.compress(data, this.COMPRESSION_LEVEL, function (compressedPayload) {
+  LZMA.compress(data, COMPRESSION_LEVEL, function (compressedPayload) {
     params = _.extend({
       lzma: util.array2hex(compressedPayload)
     }, params);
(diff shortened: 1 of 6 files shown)


src/windshaft/client.js
```diff
@@ -23,13 +23,12 @@ var WindshaftClient = function (options) {
...
   this.endpoint = options.endpoint;
   this.statTag = options.statTag;
-  this.forceCors = options.forceCors || false;
 
   this.url = this.urlTemplate.replace('{user}', this.userName);
 };
 
-WindshaftClient.DEFAULT_COMPRESSION_LEVEL = 3;
-WindshaftClient.MAX_GET_SIZE = 2033;
+WindshaftClient.prototype.MAX_URL_LENGTH = 2033;
```
rochoa: Aren't these just constants that only make sense within the definition scope? Is anyone using them?

I would declare the as `var MAX_URL_LENGTH = 2033;`, avoiding them to be available in the WindshaftClient instance.

src/windshaft/client.js
```diff
@@ -62,49 +60,72 @@ WindshaftClient.prototype.instantiateMap = function (options) {
     }
   };
 
-  if (this._usePOST(payload, params)) {
-    this._post(payload, params, ajaxOptions);
+  var encodedURL = this._generateEncodedURL(mapDefinition, params);
+  if (!this._isURLTooLong(encodedURL)) {
```
rochoa: Isn't it better to be positive? I mean:

```
if (this._isURLTooLong(encodedURL)) {
...
```

test/spec/api/create-vis.spec.js
```diff
@@ -281,7 +281,7 @@ describe('src/api/create-vis', function () {
 
   var mapInstantiationRequestDone = function () {
     return _.any($.ajax.calls.allArgs(), function (args) {
-      var expectedURLRegexp = /(http|https):\/\/cdb.localhost.lan:8181\/api\/v1\/map\/named\/tpl_6a31d394_7c8e_11e5_8e42_080027880ca6\?stat_tag=6a31d394-7c8e-11e5-8e42-080027880ca6/;
+      var expectedURLRegexp = /(http|https):\/\/cdb.localhost.lan:8181\/api\/v1\/map\/named\/tpl_6a31d394_7c8e_11e5_8e42_080027880ca6\?/;
```
rochoa: Why do we change this expectation?

src/windshaft/client.js
alonsogarciapablo: Totally. I'll change it. Thx

src/windshaft/client.js
alonsogarciapablo: I always try to do that, but I think in this case it reads better as it's now given that it's a sequence of options.

```
if (something) {
  // first option
} else {
  if (somethingElse) {
    // second option
  } else {
    // third option
  }
}
```

I thought about different method names to remove the negation in the conditional, but I could only think of names that were not so specific (eg: `_isURLValid`).

test/spec/api/create-vis.spec.js
alonsogarciapablo: I changed the order of params in this PR and it started to failed so I decided to "relax" the expectation a bit. Whenever I use this helper in the tests, all I want to ensure is that a request to the Maps API was performed (there are other tests to ensure URLs are generated correctly [here](https://github.com/CartoDB/cartodb.js/blob/4eb7182dc406559215b64b1295dc6d6143d153f9/test/spec/windshaft/client.spec.js)).

PR #1401, opened 2016-09-05
New legends :)

### LINKED ISSUE #1397: Legends
Meta issue with all the work that we need to implement the new legends:

- [x] Models, Collections(2)
  - Base class for legends and subclasses for all types of legends:
    - Legend configuration (eg: title, position)
    - Metadata to render the legend (eg: category names and colors)
  - Collection of legends
- [x] Maps API Integration (3)
  - Every time a new instance of the map is created:
    - Metadata of existing legends models is completed with the new data coming from the maps api (eg: new size for bubbles)
    - If no metadata for an existing legend model is present in the response... legend model is removed?
- [x] Load existing legends from viz.json (1)
  - Read vizjson and initialise legends
- [x] UI: Markup + css (@piensaenpixel)
- [x] Render legends (4)
  - Stacked view to render all legends
  - View for 'ramp' legend
  - View for 'category' legend
  - View for 'bubble' legend
- [x] External API (3)
  - Get legend (from layer?)
  - Show/hide legend (to layer?)
  - Update existing legend
  - Reset legend

cc: @rochoa @nobuti @piensaenpixel

alonsogarciapablo:
fixes #1397. Implements the new legends.

## External API to work with legends

```
var layer = vis.getLayer(1);

// Layers supports 5 types of legends, and can be accessed via:

layer.legends.bubble;
layer.legends.choropleth;
layer.legends.category;
layer.legends.custom;
layer.legends.html;

// Legends support the following methods (using bubble here as an example):

layer.legends.bubble.show();
layer.legends.bubble.hide();
layer.legends.bubble.set({ title: 'My legend' });
layer.legends.bubble.hasData(); // returns true if Maps API returned some data for the legend or if the legend is static
layer.legends.bubble.reset(); // resets customisable attrs to their defaults values

// Legends trigger a 'change' event every time they are updated with some new data from the Maps API.

// Here's an example:
layer.legends.bubble.on('change', function (legend) {
  if (legend.hasData()) {
    // do something (eg: disable the button to select bubble legends)
    legend.show();
  } else {
    legend.hide();
  }
});
```

@nobuti ☝️. This is what the BUILDER should be using. What do you think? Is there anything else you think we'll be needing? 

## Changes to viz.json

Changes were implemented in https://github.com/CartoDB/cartodb/pull/9838.


juanignaciosl:
I think that you actually wanted to ping @javitonino ;-)

LGTM. A couple of questions that came up to my mind while I was reading this:

1.- Order: currently, it's implicit, isn't it? Are legends displayed depending on its type?
2.- Type: one layer can only have one legend of each type, isn't it? Could this be a limitation, specially considering custom type?

alonsogarciapablo:
> 1.- Order: currently, it's implicit, isn't it? Are legends displayed depending on its type?
I haven't thought about this. So far, CartoDB.js decides, for each layer, in which order they should be displayed.

> 2.- Type: one layer can only have one legend of each type, isn't it? Could this be a limitation, specially considering custom type?

With the new legends, a layer might have more than one legend (eg: if a layer is showing features as colorised bubbles, user could choose to add a bubble legend and a category legend)...

nobuti:
LGTM only one concern, in the UI the title can be checked or unchecked, showing or hiding the title even when it is not empty, is this taking it into account? Also, for categories I guess the builder only has to replace ìtems` as the items collection in the order the user decides, right?

gfiorav:
Looks sharp to me, well done!

alonsogarciapablo:
> in the UI the title can be checked or unchecked, showing or hiding the title even when it is not empty, is this taking it into account?

Good catch @buti! I didn't take this into account. I can think of two options here:

1. The UI always sends a title when checkbox is checked (eg: "Color" or "My Legend").
2. We add support for a flag (`showTitle`) so that CartoDB.js knows when to display the given title or a default one.

I like the first one better, but we can go with the second one if you feel strong about it.

> Also, for categories I guess the builder only has to replace ìtems` as the items collection in the order the user decides, right?

Only for custom legends, the editor will have to set the "categories" explicitly. You'll be able to do something like:

```
layer.legends.custom.set('items', [
  { type: 'color', name: 'Item1', color: '#CACACA' },
  { type: 'color', name: 'Item2', color: '#FABADA'}
]);
```

Would this work? For other types of legends, data will come from the Maps API.

nobuti:
@alonsogarciapablo it sounds perfect, if checked, we set the title to user's choice or layer's name by default. If unchecked, we set title to empty. Perfect as well the categories part.
well done!

alonsogarciapablo:
@nobuti from what I understand, the name of the layer will always be displayed and is not customisable. Users can only choose a title for each of the legends:

screen shot 2016-09-06 at 14 10 20

alonsogarciapablo:
@rochoa and @xavijam, would love your feedback here too!

nobuti:
@gfiorav @javitonino @juanignaciosl one question, is the backend (rails) ready for managing more than one legend per layer? I mean, right now, in the editor, the legend is linked to the layer definition model, so it's using that endpoint. Is this still valid? What should we consider in order to save the data with the new implementation?

xavijam:
About API, not sure, but here it goes my 2 cents:

```js
var legends = layer.getLegends(); // -> legends collection
var size = legends.size();
var bubbleLegends = legends.getBubbleLegends(); // -> bubbles collection (a filter by type from legends?)
var firstBubble = bubbleLegends.first();
var secondBubble = bubbleLegends.at(1);
firstBubble.show();
firstBubble.hide();
firstBubble.append('perico');
firstBubble.prepend('de los palotes');
firstBubble.update({ title: 'whatever' });
firstBubble.isAvailable();

firstBubble.on('change', function () {
  console.log('hello world');
});
```

juanignaciosl:
@nobuti legend is currently stored inside `options` "free schema" json, so storing it is not a problem. In order to display it backend presenters should be aware of the new `legends` (plural) attribute, but storing it should* work right out of the box.

_*: there is some filtering for some risky options, but I'm almost sure that `legends` would be stored without doing anything._

alonsogarciapablo:
Thanks for your feedback @xavijam.

Guys, I was thinking about the need of exposing multiple legends of a given type for the given layers (eg: ` legends.getBubbleLegends();`). Here're some of my thoughts:

- If we went with exposing multiple legends... how would the client of this API differentiate between the different legends and know which is the one he/she should work with?. eg:

  ```
  var layer = vis.getLayer(1);
  var bubbleLegends = layer.getBubbleLegends(); // returns 3 different layers

  // does that mean CartoDB.js is rendering 3 legends?
  // how do I know what legend is begin rendered?
  // if I want to customise the title of the bubble legend for a layer... do I need to change it in all three models?
  ```
- Based on ☝️ , and regarding the use case that @rochoa mentioned offline (eg: different color ramps/category legends based on the zoom level), I think that this could be handled by cartodb.js internally. A single legend of each type could be exposed in the API for each layer so that users can customise it (and also rendered) and cartodb.js could take care of everything else internally... eg: as the map is zoomed in/out, it can select the right categories from the response of the Maps API...

To summarise, while we could/might need this in the future, AFAIK we don't need it right now, so I suggest everything working with our **current needs** and worrying about this when we get to this particular thing. We'll need to change things anyway (eg: taking the zoom into account). Agree @nobuti @xavijam @rochoa? thanks!

nobuti:
good for me.

rochoa:
I wouldn't worry about the per zoom legends use case yet. But if we go with an API like the one @xavijam is proposing I think we are covering those future use cases. Answering your questions:

> var bubbleLegends = layer.getBubbleLegends(); // returns 3 different layers

> // does that mean CartoDB.js is rendering 3 legends?
No, you just get a reference to the collection.

// how do I know what legend is begin rendered?

I see several options, I don't know much about how we usually expose collections, but it could be:
 - layer.getBubbleLegends().visible();
 - layer.getBubbleLegends({ visible: true });
 - layer.getBubbleLegends().filter({ visible: true });

// if I want to customise the title of the bubble legend for a layer... do I need to change it in all three models?

I don't understand this question. Can you provide us with an example?

rochoa:
I'm missing here the `raw`/`html` legend type. We will need it for migrating old legends.

xavijam:
Yeah, quite important ^^^.

xavijam:
About the `legends.getBubbleLegends()`, if we consider that there is no possibility to have more than one bubble (or other type) by layer, it is likely it doesn't make sense to "pluralize" that method.

alonsogarciapablo:
Talked to @xavijam and @rochoa. API we'll finally expose one legend of each type per layer for now. Internally, CartoDB.js might need to handle cases were the Maps API returns multiple metadata for the same legend type, but that we'll be taken care of internally and it'll be transparent in terms of the API...

xavijam:
Fine by me!

alonsogarciapablo:
@nobuti @gfiorav I have updated the "Changes to viz.json" section in the PR description with all the different legend types that I'm supporting in this PR.

src/windshaft-integration/model-updater.js
```diff
@@ -130,6 +137,25 @@ ModelUpdater.prototype._generateTorqueTileURLTemplate = function (windshaftMap,
   return windshaftMap.getBaseURL() + '/' + layerIndexes.join(',') + '/{z}/{x}/{y}.json.torque';
 };
 
+ModelUpdater.prototype._updateLegendModels = function (layerModel, remoteLayerIndex, windshaftMap) {
+  var layerMetadata = windshaftMap.getLayerMetadata(remoteLayerIndex);
```
alonsogarciapablo: Let's review this. This is where rules are converted to legend models, using [these adapters](https://github.com/CartoDB/cartodb.js/tree/d27434a26c552ba40ccd3138e2a7a3f00378f83b/src/windshaft-integration/legends).

src/geo/map/legends/legends.js
```diff
@@ -0,0 +1,105 @@
+var _ = require('underscore');
```
alonsogarciapablo: Let's review this.

alonsogarciapablo: This is how legend models are organised:
```
LegendModelBase
|-- CustomLegendModel
|-- HTMLLegendModel
|-- DynamicLegendModelBase
    |-- BubbleLegendModel
    |-- CategoryLegendModel
    |-- ChoroplethLegendModel
```

src/windshaft-integration/model-updater.js
```diff
@@ -197,4 +224,25 @@ ModelUpdater.prototype._setError = function (error) {
   }
 };
 
+ModelUpdater.prototype._setLegendErrors = function () {
+  // TODO: Improve this
+  var legendModels = this._layersCollection.chain()
```
alonsogarciapablo: Let's review this. This class knows a lot about layers, legends, etc..

src/geo/ui/legends/layer-legends-view.js
```diff
@@ -0,0 +1,91 @@
...
+  },
+
+  initialize: function () {
+    this._legendViews = [];
+
+    this.model.on('change:visible', this._onLayerVisibilityChanged, this);
+    this.model.on('change:layer_name', this.render, this);
+  },
+
+  render: function () {
```
alonsogarciapablo: Let's review this. This is how views are organised:

```
LegendsView
|-- LayerLegendsView
|   |-- LegendViewBase
|   |   |-- CustomLegendView
|   |   |-- HTMLLegendView
|   |   |-- DynamicLegendViewBase
|   |       |-- BubbleLegendView
|   |       |-- ChoroplethLegendView
|   |       |-- CategoryLegendView
|   |
|   |-- LegendViewBase
|   |-- ...
|
|-- LayerLegendsView
|   |-- ...
```

src/geo/map/cartodb-layer.js
```diff
@@ -29,6 +30,11 @@ var CartoDBLayer = LayerModelBase.extend({
     this.unset('infowindow');
     this.unset('tooltip');
 
+    this.legends = new Legends(attrs.legends, {
```
alonsogarciapablo: Let's review this. This is where legends for CartoDB layers are defined.

alonsogarciapablo:
Merged! Great job @piensaenpixel! And sorry for such a big PR.

I will open a follow-up PR with some feedback from @rochoa and some things that are changing in the Maps API.

cc: @xavijam @nobuti @rochoa

nobuti:
👏 👏 👏

PR #1409, opened 2016-09-21
Check if changed CartoCSS is using turbo carto in order to instantiate map

xavijam:
Basically if the `CartoCSS` is using TurboCarto, we have to instantiate in order to receive the "readable" `CartoCSS` from the instantiation, no matter what property has changed.

Things needed for this purpose:

- Added `sql_wrap` to those attributes that trigger vis reload.
- "Updated" *torque* dependency within package.json, now requesting it from repository instead of npm.
- "Updated" *carto* dependency as we do in Builder dependencies, requested from the repository.

CR: @alonsogarciapablo 
cc @rochoa and @javisantana 

This is part from https://github.com/CartoDB/cartodb/issues/9389.

### DIFF
--- a/src/geo/map/torque-layer.js
+++ b/src/geo/map/torque-layer.js
@@ -2,6 +2,7 @@ var _ = require('underscore');
 var LayerModelBase = require('./layer-model-base');
 var carto = require('carto');
 var ATTRIBUTES_THAT_TRIGGER_VIS_RELOAD = ['sql', 'sql_wrap', 'source', 'cartocss'];
+var PROPERTIES_THAT_CAN_USE_TURBO_CARTO = ['marker-fill', 'marker-line-color'];
 var TORQUE_LAYER_CARTOCSS_PROPS = [
   '-torque-frame-count',
   '-torque-time-attribute',
@@ -42,9 +43,7 @@ var TorqueLayer = LayerModelBase.extend({
     var reloadVis = _.any(ATTRIBUTES_THAT_TRIGGER_VIS_RELOAD, function (attr) {
       if (this.hasChanged(attr)) {
         if (attr === 'cartocss') {
-          var hasTorquePropsChanged = this._torqueCartoCSSPropsChanged();
-          var usesTurboCarto = this._hasTurboCarto();
-          return this.previous('cartocss') && (hasTorquePropsChanged || usesTurboCarto);
+          return this.previous('cartocss') && (this._torqueCartoCSSPropsChanged() || this._hasTurboCarto());
         }
         return true;
       }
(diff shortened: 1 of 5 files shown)


src/geo/map/torque-layer.js
```diff
@@ -56,6 +55,23 @@ var TorqueLayer = LayerModelBase.extend({
...
 
+  _hasTurboCarto: function () {
+    var currentCartoCSS = this.get('cartocss');
+    var shader = new carto.RendererJS().render(currentCartoCSS);
+    var layer = shader.findLayer({ name: '#layer' });
+
+    if (layer) {
+      var layerShader = layer.getStyle({ property: 1 }, { zoom: 10 });
+
+      var markerFill = layerShader['marker-fill'];
```
alonsogarciapablo: Other properties could have ramps too: `polygon-fill`, `marker-width`... Are these relevant here too?

src/geo/map/torque-layer.js
```diff
@@ -110,36 +126,35 @@ var TorqueLayer = LayerModelBase.extend({
...
-    return this.get('type') === other.get('type') && _.every(properties, function(p) {
-      return other.get(p) === self.get(p);
-    });
+    return this.get('type') === other.get('type') && _.every(properties, function (p) {
+        return other.get(p) === self.get(p);
+      });
   },
 
-  isVisible: function() {
+  isVisible: function () {
```
alonsogarciapablo: If we're fixing all linter errors from this file... let's remove it from .eslintignore!

src/geo/map/torque-layer.js
```diff
@@ -1,6 +1,15 @@
...
 var carto = require('carto');
+var ATTRIBUTES_THAT_TRIGGER_VIS_RELOAD = ['sql', 'sql_wrap', 'source', 'cartocss'];
+var TORQUE_LAYER_CARTOCSS_PROPS = [
+  '-torque-frame-count',
+  '-torque-time-attribute',
+  '-torque-aggregation-function',
+  '-torque-data-aggregation',
+  '-torque-resolution'
+];
+var LAYER_NAME_IN_CARTO_CSS = 'Map';
```
alonsogarciapablo: 👍

src/geo/map/torque-layer.js
```diff
@@ -42,10 +39,12 @@ var TorqueLayer = LayerModelBase.extend({
...
 
   _onAttributeChanged: function () {
-    var reloadVis = _.any(this.ATTRIBUTES_THAT_TRIGGER_VIS_RELOAD, function (attr) {
+    var reloadVis = _.any(ATTRIBUTES_THAT_TRIGGER_VIS_RELOAD, function (attr) {
       if (this.hasChanged(attr)) {
         if (attr === 'cartocss') {
-          return this.previous('cartocss') && this._torqueCartoCSSPropsChanged();
+          var hasTorquePropsChanged = this._torqueCartoCSSPropsChanged();
+          var usesTurboCarto = this._hasTurboCarto();
+          return this.previous('cartocss') && (hasTorquePropsChanged || usesTurboCarto);
```
alonsogarciapablo: What if layer uses/has TurboCarto but it hasn't changed?

src/geo/map/torque-layer.js
```diff
@@ -42,10 +39,12 @@ var TorqueLayer = LayerModelBase.extend({
   },
 
   _onAttributeChanged: function () {
-    var reloadVis = _.any(this.ATTRIBUTES_THAT_TRIGGER_VIS_RELOAD, function (attr) {
+    var reloadVis = _.any(ATTRIBUTES_THAT_TRIGGER_VIS_RELOAD, function (attr) {
       if (this.hasChanged(attr)) {
         if (attr === 'cartocss') {
-          return this.previous('cartocss') && this._torqueCartoCSSPropsChanged();
+          var hasTorquePropsChanged = this._torqueCartoCSSPropsChanged();
+          var usesTurboCarto = this._hasTurboCarto();
```
alonsogarciapablo: Personal detail... I would remove these intermediate variables since they are only used once and are very similar to invoking the method directly...

xavijam: I tried to make the conditional shorter and clearer, but I'm fine with that.

src/geo/map/torque-layer.js
xavijam: it doesn't matter, it should re-instantiate the map because maps api has to convert that CartoCSS (with turbo-carto) to a valid and readable CartoCSS that torque understands.

src/geo/map/torque-layer.js
alonsogarciapablo: Perhaps we could iterate over the properties of `layerShader` and see if any of them is a `ramp`?

src/geo/map/torque-layer.js
xavijam: True that, sorry.

src/geo/map/torque-layer.js
xavijam: We could do that yes, not sure if that makes sense in any case, because polygon-fill with torque is not possible (yet), or apply categories (ramp) to line-stroke.

xavijam:
Yay or nay @alonsogarciapablo ?

PR #1416, opened 2016-09-28
fix legends ui

piensaenpixel:
image


src/geo/ui/legends/base/legend-title.tpl
```diff
@@ -1 +1 @@
-<h3 class="CDB-Text CDB-Size-small u-upperCase u-bSpace--m u-altTextColor"><%= title %></h3>
+<h3 class="CDB-Text CDB-Size-small u-upperCase u-bSpace u-altTextColor"><%= title %></h3>
```
alonsogarciapablo: Now that I see it... we should use `<%- title %>` here. Would you please change it for me?

PR #1418, opened 2016-09-28
Adapt to new format of carto_css meta coming from Maps API

alonsogarciapablo:
@rochoa this is it 😄


rochoa:
Removing if/elseifs 👍 .

PR #1419, opened 2016-09-29
Catch errors thrown by legend adapters

alonsogarciapablo:
And reflect that in the UI by changing legend's state to "error" when this happens. It was stopping the map re-instantiation before.

Also, ensure that only one rule is used to update each legend model.

@rochoa 👍 ?

### DIFF
--- a/src/windshaft-integration/model-updater.js
+++ b/src/windshaft-integration/model-updater.js
@@ -150,19 +150,15 @@ ModelUpdater.prototype._updateLegendModel = function (legendModel, layerMetadata
     };
     if (cartoCSSRules) {
       var adapter = RuleToLegendModelAdapters.getAdapterForLegend(legendModel);
-      var cartoCSSRule = _.find(cartoCSSRules, function (rule) {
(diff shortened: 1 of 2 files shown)


src/windshaft-integration/model-updater.js
```diff
@@ -136,21 +137,29 @@ ModelUpdater.prototype._generateTorqueTileURLTemplate = function (windshaftMap,
...
+      var adapter = RuleToLegendModelAdapters.getAdapterForLegend(legendModel);
+      var ruleForLegend = _.find(cartoCSSRules, adapter.canAdapt);
+      if (ruleForLegend) {
+        newLegendAttrs = _.extend(newLegendAttrs, adapter.adapt(ruleForLegend));
+      }
     }
     legendModel.set(newLegendAttrs);
-  });
+  } catch (error) {
+    legendModel.set({ state: 'error' });
```
rochoa: 👍

rochoa:
:shipit:

Issue #1395, opened 2016-08-25
Torque API key and auth_token


rochoa:
We were not including those params in torque layers. We did a quick and dirty workaround to fix the problem but we need to find a proper solution for this situation.

We also need to fix/homogenize the naming for those params because there are several names for the same thing:
 - authToken, auth_token.
 - apiKey, api_key, map_key.
And we also store them in model.attributes.extra_params sometimes, but in other situations we store them directly in model.attributes.

Reference https://github.com/CartoDB/cartodb.js/pull/1394

cc @xavijam

PR #1421, opened 2016-10-03
Initialize torque models correctly

alonsogarciapablo:
Fixes CartoDB/cartodb#9548 and #1395.

@xavijam what do you think?


src/api/create-vis.js
```diff
@@ -9,7 +9,6 @@ var DEFAULT_OPTIONS = {
   loaderControl: true,
   infowindow: true,
   tooltip: true,
-  time_slider: true,
```
xavijam: That's a good one! :+1:

src/api/create-vis.js
```diff
@@ -27,10 +26,29 @@ var createVis = function (el, vizjson, options) {
...
+  return visModel;
+};
+
+var loadVizJSON = function (el, visModel, vizjsonData, options) {
+  var vizjson = new VizJSON(vizjsonData);
+  applyOptionsToVizJSON(vizjson, options);
 
-  // TODO: We can check if all required options are present here! eg: if viz.json has some analyses
-  // apiKey or authToken will be required (otherwise we know requests to Windshaft will fail)...
+  var isProtocolHTTPs = window && window.location.protocol && window.location.protocol === 'https:';
```
xavijam: `window && window.location && window.location.protocol && window.location.protocol === 'https:';` ?

xavijam: 💃

src/vis/vis.js
```diff
@@ -89,23 +87,25 @@ var VisModel = Backbone.Model.extend({
...
+    this._windshaftSettings = {
+      urlTemplate: vizjson.datasource.maps_api_template,
+      userName: vizjson.datasource.user_name,
+      statTag: vizjson.datasource.stat_tag,
+      apiKey: this.get('apiKey'),
+      authToken: this.get('authToken')
+    };
+
+    if (vizjson.isNamedMap()) {
+      this._windshaftSettings.templateName = vizjson.datasource.template_name;
```
xavijam: `vizjson.getDatasourceTemplateName();`

src/vis/layers-factory.js
```diff
@@ -74,33 +76,52 @@ var LAYER_CONSTRUCTORS = {
...
-  create: function (type, data, options) {
-    var LayerClass = LAYER_CONSTRUCTORS[type.toLowerCase()];
-    if (!LayerClass) {
-      log.error("error creating layer of type '" + type + "'");
-      return null;
-    }
-    // Flatten "options"
-    var layerAttributes = _.extend({}, _.omit(data, 'options'), data.options);
-    return new LayerClass(layerAttributes, options);
+var LayersFactory = function (deps) {
```
xavijam: I like `deps` name, I was using `params` for the same but sounds better, note taken.

src/vis/vis.js
```diff
@@ -21,7 +20,6 @@ var STATE_ERROR = 'error'; // vis has been sent to Windshaft and there were some
 var VisModel = Backbone.Model.extend({
   defaults: {
     loading: false,
-    https: false,
```
xavijam: https option not provided anymore?

alonsogarciapablo: In theory, `https` could be undefined. According to our actual docs:

- `http === true`. Will try to transform **http** urls of basemaps to **https**
- `https === false`. Will try to tranform **https** urls of basemaps to **http**
- `https === undefined`. Will leave basemap urls as they are.

The truth is that, right now the vizjson files don't include `https` and we're not using the `https` option either. So, `https` is being set `true` if the page is served over `https` and `false` otherwise.

PR #1425, opened 2016-10-04
Decouple layer selector from legends.

nobuti:
In order to be able to show one without the other because of map options.

cc @alonsogarciapablo @piensaenpixel


src/vis/vis-view.js
```diff
@@ -20,6 +20,7 @@ var Vis = View.extend({
     this.overlays = [];
 
     this.model.on('change:showLegends', this._onShowLegendsChanged, this);
+    this.model.on('change:showLayerSelector', this._onShowLayerSelectorChanged, this);
```
alonsogarciapablo: Do we need this? From what I understand, `legendsView` is going to be rendered no mater what... I see `legendsView` already hides itself when these two flags are false...

src/geo/ui/legends/legends-view.js
```diff
@@ -9,16 +9,24 @@ var LegendsView = Backbone.View.extend({
...
-  initialize: function (deps) {
-    if (!deps.layersCollection) throw new Error('layersCollection is required');
-    this._layersCollection = deps.layersCollection;
+  initialize: function (options) {
+    if (!options.layersCollection) throw new Error('layersCollection is required');
+    this._layersCollection = options.layersCollection;
 
     this._isRendered = false;
+
+    this.renderModel = new Backbone.Model({
```
nobuti: 👍

nobuti:
Ready for a second pass.

src/api/create-vis.js
```diff
@@ -53,23 +54,37 @@ var loadVizJSON = function (el, visModel, vizjsonData, options) {
...
+  if (_.isBoolean(options.layer_selector)) {
+    showLayerSelector = options.layer_selector;
+  } else if (vizjson.embed_options && _.isBoolean(vizjson.embed_options.layer_selector)) {
+    showLayerSelector = vizjson.embed_options.layer_selector;
+  }
+
+  var settingsModel = new SettingsModel({
+    showLegends: showLegends,
+    showLayerSelector: showLayerSelector
+  });
```
alonsogarciapablo: 👍

src/vis/ui-settings.js
```diff
@@ -0,0 +1,10 @@
+var Backbone = require('backbone');
```
alonsogarciapablo: I would just call this file `settings.js`

src/api/create-vis.js
```diff
@@ -53,23 +54,37 @@ var loadVizJSON = function (el, visModel, vizjsonData, options) {
...
     title: options.title || vizjson.title,
     description: options.description || vizjson.description,
     apiKey: options.apiKey,
     authToken: options.authToken,
-    showLegends: showLegends,
     showEmptyInfowindowFields: options.show_empty_infowindow_fields === true,
     https: isProtocolHTTPs || options.https === true || vizjson.https === true
   });
 
+  visModel.settings = settingsModel;
```
alonsogarciapablo: This kind of strange... the visModel won't have a `settings` prop until we get to this point. Imagine the following scenario:

```
var vis = createVis(...);
vis.settings.set('showLegends', true);
```
Here, we'll get an error because `vis.settings` is still `undefined`... right?

Idea:

```
// visModel
...
initialize: () {
  ...
  // A vis model gets some settings (with defaults)
  this.settings = new Settings();
},

...
// we provide a method to set some new settings
setSettings: function (settings) {
  this.settings.set(settings);
}
...
```
in `create-vis.js`, once the vizjson has been loaded (instead of this) we can do:

```
visModel.setSettings({...});
```

src/api/create-vis.js
```diff
@@ -79,12 +73,15 @@ var loadVizJSON = function (el, visModel, vizjsonData, options) {
     https: isProtocolHTTPs || options.https === true || vizjson.https === true
   });
 
-  visModel.settings = settingsModel;
+  visModel.settings.set({
```
alonsogarciapablo: Let's use setSettings here! 👍

PR #1428, opened 2016-10-06
State url

nobuti:
This PR implements the needed changes for https://github.com/CartoDB/deep-insights.js/issues/445

It's WIP so every comment will be appreciated.


xavijam:
Could you give us an overview about what you have implemented so far?

nobuti:
This is ready for CR.

The approach is to do a double instantiation if filters are present. To do it, `instantiateMap` include a param to indicate if the request should be happen with or without filters. Also, in order to begin to listen all the dataviews, we wait until all of them have been fetched. To achieve this, a collection handles it. Basically, all dataviews are patched augmenting the fetch method, to manage the fetch status:

```
dataviewModel.fetch = function (options) {
    dataviewModel.set('fetchStatus', 'fetching');

    var newSuccess = function () {
      dataviewModel.set('fetchStatus', 'fetched');
      options && options.success && options.success();
    };

    var newError = function () {
      dataviewModel.set('fetchStatus', 'unavailable');
      options && options.error && options.error();
    };

    _oldFetch({
      success: newSuccess,
      error: newError
    });
};
```

When all are `fetched` an event is emitted, so we have green light to listen them to track the url state.

xavijam:
Could these changes support [this ticket](https://github.com/CartoDB/deep-insights.js/issues/431)?

src/vis/dataviews-tracker.js
```diff
@@ -0,0 +1,64 @@
...
+  _initBinds: function () {
+    this.on('add', this._patch, this);
+  },
+
+  track: function () {
+    this._tracking = true;
+    this._check();
+  },
+
+  _patch: function (dataviewModel) {
```
alonsogarciapablo: Why not let each dataview keep track of its status [in its own fetch method](https://github.com/CartoDB/cartodb.js/blob/v4/src/dataviews/dataview-model-base.js#L253)?

```

var UNFETCHED_STATUS = 'unfeteched';
var FETCHING_STATUS = 'fetching';
var FETCHED_STATUS = 'fetched;
var FETCH_ERROR_STATUS = 'fetched;

var DataviewModelBase = Model.extend({
  default: {
    ...,
    status: UNFETCHED_STATUS,
  },

  ...,

  fetch: function (options) {
    this.set('status', FETCHING_STATUS);

    ...

    return Model.prototype.fetch.call(this, _.extend(opts, {
      success: function () {
        
        ...
        this.set('status', FETCHED_STATUS);
      }.bind(this),
      error: function (mdl, err) {
        ...
        this.set('status', FETCH_ERROR_STATUS)
      }.bind(this)
    }));
  },

  isFetched: function () {
    return this.get('status', FETCHED_STATUS);
  },

  ...
});
```

src/windshaft-integration/model-updater.js
```diff
@@ -164,6 +167,8 @@ ModelUpdater.prototype._updateLegendModel = function (legendModel, layerMetadata
 
 ModelUpdater.prototype._updateDataviewModels = function (windshaftMap, sourceId, forceFetch) {
   this._dataviewsCollection.each(function (dataviewModel) {
+    this._dataviewsTracker.add(dataviewModel);
```
alonsogarciapablo: Instead of having to add this manually here... couldn't we pass the collection of dataviews to the `dataviewsTracker` so that it takes care of tracking dataviews?

src/vis/vis.js
```diff
@@ -262,29 +268,46 @@ var VisModel = Backbone.Model.extend({
...
+      });
+    }
+  },
+
   /**
    * Force a map instantiation.
    * Only expected to be called once if {skipMapInstantiation} flag is set to true when vis is created.
    */
-  instantiateMap: function (options) {
+  instantiateMap: function (options, includeFilters) {
```
alonsogarciapablo: Isn't includeFilter another `option`? Why a different parameter?

src/vis/vis.js
```diff
@@ -262,29 +268,46 @@ var VisModel = Backbone.Model.extend({
     }
   },
 
+  init: function (options) {
```
alonsogarciapablo: How are we planning to use this? `vis.instantiateMap` [can only be called once](https://github.com/CartoDB/cartodb.js/blob/a82b9b3e227099e41e454e97d677bf2a1a478cd9/src/vis/vis.js#L291). Do you think we have this logic there instead of adding a new public method?

src/windshaft/map-base.js
```diff
@@ -47,13 +47,21 @@ var WindshaftMap = Backbone.Model.extend({
   },
 
   // TODO: Move this somewhere else and keep this class as a wrapper for windshaft responses
-  createInstance: function (options) {
+  createInstance: function (options, includeFilters) {
```
alonsogarciapablo: I'd treat `includeFilters` as an option here too... IMO it's more clear to have something like `wadus({ includeFilter: true });` than `wadus(true)`.

src/vis/vis.js
```diff
@@ -360,6 +383,17 @@ var VisModel = Backbone.Model.extend({
       })
       .flatten()
       .value();
+  },
+
+  _getFilters: function () {
+    return this._dataviewsCollection.reduce(function (filters, dataview) {
```
alonsogarciapablo: All we care about here is if any of the dataviews have a filter, so I'd try to do something simpler like:

```
_isAnyDataviewFiltered: function () {
  return this._dataviewsCollection.any(function (dataview) {
    return dataview.isActive(); // !dataview.isEmpty()
  });
}
```

If you feel 💪 , perhaps we could create a `DataviewsCollection` so we can put this logic there...

src/vis/vis.js
nobuti: It's only called once. https://github.com/CartoDB/cartodb.js/blob/a82b9b3e227099e41e454e97d677bf2a1a478cd9/src/vis/vis.js#L276-L285

nobuti:
@xavijam the last commit fixes it

xavijam:
:heart:

nobuti:
Ready for another pass!

src/vis/vis.js
```diff
@@ -288,26 +285,26 @@ var VisModel = Backbone.Model.extend({
...
    */
-  instantiateMap: function (options, includeFilters) {
+  instantiateMap: function (options) {
     options = options || {};
     if (!this._instantiateMapWasCalled) {
       this._instantiateMapWasCalled = true;
       var successCallback = options.success;
       options.success = function () {
-        this._dataviewsTracker.track();
+        this._dataviewsCollection.track();
```
alonsogarciapablo: I kind of liked the dataview tracker that we had before: a class whose only responsibility is to trigger an event once all the given dataviews are fetched (I wasn't suggesting moving all the logic to the dataviewsCollection, but only `isAnyDataviewFiltered`...). Also, I'd rather trigger the `dataviewsFetched` from this class, so that it's not hidden somewhere...

```
var promise = DataviewStatus.waitUntilAllDataviewsFetched(this._dataviewsCollection.models);
promise.then(function () {
  this.trigger('dataviewsFetched');
}.bind(this));
```

Just an idea... We could do something similar with events too... what do you think?

xavijam:
Ready then?

nobuti:
Ready for CR again, yes

src/vis/vis.js
alonsogarciapablo: Would something like this work?

```
  instantiateMap: function (options) {
    options = options || {};
    if (this._instantiateMapWasCalled) {
      return;
    }
    this._instantiateMapWasCalled = true;

    this.reload({
      success: function () {
        options.success && options.success();
        this._onMapInstantiatedForTheFirstTime()
      }.bind(this)
      includeFilters: this._isAnyDataviewFiltered()
    });
  },

  _onMapInstantiatedForTheFirstTime: function () {
    whenAllDataviewsFetched(this._dataviewsCollection, this._onDataviewFetched.bind(this));
    this._initBindsAfterFirstMapInstantiation();
    
    if (this._isAnyDataviewFiltered()) {
      this.reload({
        includeFilters: false
      })
    }
  },

  _isAnyDataviewFiltered: function () {
    return this._dataviewsCollection.isAnyDataviewFiltered();
  }
```

alonsogarciapablo: (I'm not sure if the `options.success` callback that we're passing to `instantiateMap` would need to be called after the first or the second map reload when there're active filters.).

test/spec/vis/vis.spec.js
```diff
@@ -303,12 +303,44 @@ describe('vis/vis', function () {
...
           forceFetch: 'forceFetch',
-          success: 'success'
+          success: 'success',
+          includeFilters: true
         });
       });
     });
   });
 
+  describe('.init', function () {
```
alonsogarciapablo: There's no `.init` method anymore, so these tests should be part of `.createInstance`

nobuti:
thanks to you! 💋

xavijam:
cc @gfiorav, it will make him a bit happier.

PR #1429, opened 2016-10-06
popups: alternative names in custom templates

matallo:
fixed an issue where values in custom templates weren't appearing, related to alternative names overriding titles

also, removed functionality where titles had underscores and they were _sanitized_, but again used in custom templates, thus not appearing

- added test to check alternative_name, and no title in both default and custom template
- changes in default template, surfaced when adding the tests

find the discussion below:

---

managed to reproduce it, this was already happening in the current editor, where we sanitize the title in the infowindow and replace "_" with spaces (see attached)

https://github.com/CartoDB/cartodb.js/blob/develop/src/geo/ui/infowindow.js#L455

infowindows

I'm removing this sanitization as it doesn't contribute a lot, and the user could always make this change manually with the alternative name

another issue which is still happening in new builder (but doesn't in eld editor) is when alternative names are set, we lose the field in the infowindow no matter if sanitization is done or not, because the field title is overriden with the alternative name, thus the index doesn't exist

this is happening because we didn't do that previously for custom templates, but new infowindows don't take that into account

so, I'll be working in these two fixes (and update the tests to cover the cases), and then I'll move to CartoDB/cartodb#9993 which I also reproduced and seem serious

### DIFF
--- a/src/geo/ui/default-infowindow-template.tpl
+++ b/src/geo/ui/default-infowindow-template.tpl
@@ -12,7 +12,6 @@
               <li class="CDB-infowindow-listItem">
                 <% if (field.title) { %><h5 class="CDB-infowindow-subtitle"><%- field.title %></h5><% } %>
                 <% if (field.value) { %><h4 class="CDB-infowindow-title"><%- field.value %></h4><% } %>
-                <% if (!field.title) { %><h4 class="CDB-infowindow-title">null</h4><% } %>
               </li>
               <% }) %>
           <% } %>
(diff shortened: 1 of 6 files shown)


matallo:
CR @alonsogarciapablo

matallo:
this is such a good question I had to look it, indeed https://github.com/CartoDB/cartodb.js/blob/v4/src/geo/ui/infowindow-model.js#L90-L121

at the time I did the changes I grabbed it from current used attributes

merging then

PR #1430, opened 2016-10-06
Hide legends if no legend is visible.

nobuti:
This PR hides the legends container if layer selector is not activated and no legend is visible.

map-legends-options

@alonsogarciapablo 🙏

### DIFF
--- a/src/geo/ui/legends/layer-legends-view.js
+++ b/src/geo/ui/legends/layer-legends-view.js
@@ -2,9 +2,10 @@ var _ = require('underscore');
 var Backbone = require('backbone');
 var template = require('./layer-legends-template.tpl');
 var LegendViewFactory = require('./legend-view-factory');
+
 var LayerLegendsView = Backbone.View.extend({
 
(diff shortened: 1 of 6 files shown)


nobuti:
cc @xavijam

xavijam:
Reviewing myself.

src/geo/ui/legends/layer-legends-view.js
```diff
@@ -14,26 +15,40 @@ var LayerLegendsView = Backbone.View.extend({
     this._legendViews = [];
 
     this.settingsModel = options.settingsModel;
+    this.tryContainerVisibility = options.tryContainerVisibility;
```
xavijam: What is it?

nobuti: There is a general container for all layers. Then every layer has its own container for its legends. This general container is managed by a higher order view, so from the inner views we need to indicate a way to hide the general container.

xavijam: Ajam!

Issue #1423, opened 2016-10-04
Geometry Management


alonsogarciapablo:
- [x] Enter "Geometry drawing mode":
  - User can draw a geometry by clicking on the map (point, polygon, polyline). 
  - App can get the GeoJSON of the geometry.
  - App knows when geometry is "ready".
  - App can disable map interaction.
- [x] Exit "Geometry drawing model":
  - App can enable map interaction again.
- [x] Add a geometry to the map (based on GeoJSON and some styles).
- [x] Edit/Drag a geometry. Get the GeoJSON back.
- [x] Ensure search results work

alonsogarciapablo:
Closed via #1431

PR #1431, opened 2016-10-07
Geometry drawing and edition

alonsogarciapablo:
Part of https://github.com/CartoDB/cartodb.js/issues/1423.

Take a look at [this example](https://github.com/CartoDB/cartodb.js/blob/geometry-management/examples/v4/geometry-drawing.html) to learn more about the new API.

## API

**Drawing (disables map interaction - feature clicks and popups):**
```
vis.map.drawPoint(); // returns a Point
vis.map.drawPolygon(); // returns a Polygon
vis.map.drawPolyline(); // returns a Polyline
vis.map.stopDrawingGeometry() // removes geometry from the map
```

**Working with points:**
```
var point = vis.map.drawPoint();
point.setLatLng(...);
point.on('change', function (point) {
  if (point.isComplete()) {
    console.log(point.toGeoJSON());
  }
});
```

**Working with polygons:**
```
var polygon = vis.map.drawPolygon();
polygon.setLatLngs(...);
polygon.on('change', function (polygon) {
  if (polygon.isComplete()) {
    console.log(polygon.toGeoJSON());
  }
});
```

**Working with polylines:**
```
var polyline = vis.map.drawPolyline();
polyline.setLatLngs();
polyline.on('change', function (polyline) {
  if (polyline.isComplete()) {
    console.log(polyline.toGeoJSON());
  }
});
```

**Editing GeoJSON geometries (disables map interaction - feature clicks and popups):**

Supports Point, Polygon, LineString, MultiPoint, MultiPolygon, MultiLineString geometries:

```
vis.map.editGeometry({
  "type": "Feature",
  "properties": {},
  "geometry": {
    "type": "Point",
    "coordinates": [
      -3.779296875,
      40.245991504199026
    ]
  }
}
```

**Adding a custom overlay**
As part of this PR I also added the ability to add "custom" overlays to the map:

```
var overlay = new Backbone.View({ ... });
vis.map.addCustomOverlay(overlay);
```

**Enabling feature interactivity**
By default, feature interactivity is disabled. It can be enabled by passing the `interactiveFeatures` flag to `createVis`:

```
cdb.createVis('map', vizjson, { interactiveFeatures: true })
```
When feature interactivity is enabled, `vis.map` will trigger `featureClick` events.

cc: @rochoa @javisantana @xavijam


javisantana:
I think it's good, I miss editing feauture api.

Other things could be important:

- One feature everybody requested was to be able to add more than a point at once
- For you to give more vision one of the things we'd need is to draw a circle/rectangle for filtering. I think the api you propose supports that anyway

alonsogarciapablo:
> I miss editing feauture api.

I guess you're referring to the second part of https://github.com/CartoDB/cartodb.js/issues/1423, which is what I'm about to start working on. Basically: being able to listen to clicks on features and being able to add features (editable) based on some GeoJSON... right?

javisantana:
I just miss the example api, like:

```
var polygon = map.startEditFeature(geojson)
map.stopEditing(); // or maybe stopDrawing as well
```

alonsogarciapablo:
👍  That's exactly what I'm about to work on now! 😄  Thanks!

matallo:
just two comments I noticed while working with geometries, as @javisantana is taking the CR

the precision we're using for lnglats in the server is shorter than the one we sent (i.e. `-61.938950426660604` vs `-61.93895043`) so maybe it's just not worth sending the whole thing but a rounded one

I noticed this because after the fetch the geometry I sent was different than the one that was being saved

the other issue I'm not able to reproduce was exactly this https://github.com/CartoDB/cartodb.js/commit/1315b7abe00b617a8d85fae89ae41d051c28d73a

so ✨ for me man

src/geo/geometry-models/geometry-factory.js
```diff
@@ -0,0 +1,125 @@
+var _ = require('underscore');
+var Point = require('./point');
+var Polyline = require('./polyline');
+var Polygon = require('./polygon');
+var MultiPoint = require('./multi-point');
+var MultiPolygon = require('./multi-polygon');
+var MultiPolyline = require('./multi-polyline');
+var GeoJSONHelper = require('./geojson-helper');
+
+var GEOJSON_TYPE_TO_CREATE_METHOD_NAME = {
```
javisantana: wondering why not doing a real factory using map string -> function :troll:

javisantana:
The problem with truncate is the zoom level but I don't think it's a major problem right now

src/vis/map-cursor-manager.js
```diff
@@ -0,0 +1,29 @@
...
+  if (!deps.mapView) throw new Error('mapView is required');
+  if (!deps.mapModel) throw new Error('mapModel is required');
+  if (!deps.featureEvents) throw new Error('featureEvents is required');
+
+  this._mapView = deps.mapView;
+  this._mapModel = deps.mapModel;
+  this._featureEvents = deps.featureEvents;
+
+  this._featureEvents.on('featureOver', this._onFeatureOver, this);
+  this._featureEvents.on('featureOut', this._onFeatureOut, this);
```
alonsogarciapablo: This class is not going to "die" during the session... where would I unbind these events?

src/geo/geometry-models/geometry-factory.js
```diff
@@ -7,119 +7,125 @@ var MultiPolygon = require('./multi-polygon');
...
+  return createMultiPolyline({
     geojson: geoJSON,
     editable: true
   }, {
     latlngs: latlngs
   });
 };
 
-module.exports = GeometryFactory;
+var GEOJSON_TYPE_TO_CREATE_METHOD = {
```
alonsogarciapablo: cc: @javisantana 👍

javisantana:
:es:

PR #1436, opened 2016-10-07
hide legends in mobile view

piensaenpixel:
until we have a solution for mobile view, it's better hide them in mobile view.
could you take a look @nobuti ?

cc/ @xavijam

### DIFF
--- a/themes/scss/map/overlays.scss
+++ b/themes/scss/map/overlays.scss
@@ -504,3 +504,10 @@ $index-items: 7 6 5 4 3 2 1;
 .Legend-placeholderInner {
   padding: 16px 24px 0;
 }
+
+
+@media (max-width: 759px) {
+  .CDB-Legends-canvas {
+    display: none !important;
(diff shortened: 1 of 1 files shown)


themes/scss/map/overlays.scss
```diff
@@ -504,3 +504,10 @@ $index-items: 7 6 5 4 3 2 1;
 .Legend-placeholderInner {
   padding: 16px 24px 0;
 }
+
+
+@media (max-width: 759px) {
+  .CDB-Legends-canvas {
+    display: none !important;
```
nobuti: why don't you reverse the condition? something like this:
```
.CDB-Legends-canvas {
    display: none;
}

@media (min-width: 760px) {
    display: whatever;
}
```
it should wok without !important right?

piensaenpixel: not because we are adding display:block inline in the html
image

PR #1437, opened 2016-10-08
Legends fixes

nobuti:
This PR fixes some bugs at cartodb.js side:

- We need to validate ranges before to update model's values. For instance, if we try to style size using a column full of null values, the metadata returned by cartocss shows `null` for the start filter's property.
- In bubble legend, we need to check sizes before to reverse, because we were assuming first value is always bigger than the last one.
- We need to apply some additional restrictions in order to reset the legend model's attributes, besides `state` and `visible`. This provoked some weird side effects. From the builder we are calling `reset` when a legend is created, so for example, for a choropleth we were basically setting `color` as an empty array, so the model turned into a non valid one. So the no data view was rendered.
- Force parse for custom legend model in order to translate vizjson values correctly.

### DIFF
--- a/src/geo/map/legends/bubble-legend-model.js
+++ b/src/geo/map/legends/bubble-legend-model.js
@@ -12,8 +12,11 @@ var BubbleLegendModel = LegendModelBase.extend({
     });
   },
 
-  _avoidResetTheseAttributes: function () {
-    return ['values'];
+  getNonResettableAttrs: function () {
+    return _.union(
+      LegendModelBase.prototype.getNonResettableAttrs.apply(this),
+      ['values']
+    );
   },
 
   isAvailable: function () {
--- a/src/geo/map/legends/bubble-legend-model.js
+++ b/src/geo/map/legends/bubble-legend-model.js
@@ -12,6 +12,10 @@ var BubbleLegendModel = LegendModelBase.extend({
     });
   },
 
+  _avoidResetTheseAttributes: function () {
+    return ['values'];
+  },
+
   isAvailable: function () {
     return this.get('values') && this.get('values').length > 0;
   }
--- a/src/geo/map/legends/custom-legend-model.js
+++ b/src/geo/map/legends/custom-legend-model.js
@@ -7,24 +7,6 @@ var CustomLegendModel = StaticLegendModelBase.extend({
       type: 'custom',
       items: []
     });
-  },
-
-  parse: function (r, opts) {
-    var attrs = _.extend({},
-      _.omit(r, 'items')
-    );
-    attrs.items = [];
-
-    if (r.items) {
-      attrs.items = r.items.map(function (item) {
-        return {
-          name: item.title,
-          color: item.color
-        };
-      });
-    }
-
-    return attrs;
   }
 });
 
--- a/src/geo/map/legends/custom-legend-model.js
+++ b/src/geo/map/legends/custom-legend-model.js
@@ -7,6 +7,24 @@ var CustomLegendModel = StaticLegendModelBase.extend({
       type: 'custom',
       items: []
     });
+  },
+
+  parse: function (r, opts) {
+    var attrs = _.extend({},
+      _.omit(r, 'items')
+    );
+    attrs.items = [];
+
+    if (r.items) {
+      attrs.items = r.items.map(function (item) {
+        return {
+          name: item.title,
+          color: item.color
+        };
+      });
+    }
+
+    return attrs;
   }
 });
 
--- a/src/geo/map/legends/legend-model-base.js
+++ b/src/geo/map/legends/legend-model-base.js
@@ -54,9 +54,14 @@ var LegendModelBase = Backbone.Model.extend({
 
   reset: function () {
     var defaults = _.omit(this.defaults(),
-      NON_RESETEABLE_DEFAULT_ATTRS
+      NON_RESETEABLE_DEFAULT_ATTRS,
+      this._avoidResetTheseAttributes()
     );
     this.set(defaults);
+  },
+
+  _avoidResetTheseAttributes: function () {
+    return [];
   },
 
   isAvailable: function () {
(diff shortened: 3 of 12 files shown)


xavijam:
@alonsogarciapablo, could you take a look please? Because you were involved in the implementation.

src/geo/map/legends/custom-legend-model.js
```diff
@@ -7,6 +7,24 @@ var CustomLegendModel = StaticLegendModelBase.extend({
       type: 'custom',
       items: []
     });
+  },
+
+  parse: function (r, opts) {
```
alonsogarciapablo: Why are we using `parse` and not doing this in the constructor?

From Backbone's docs:
```
parse is called whenever a model's data is returned by the server,
```

The problem I see with this approach is that then we have to use `parse:true` everywhere, which is something easy to forget, don't you think?

src/geo/map/legends/legend-model-base.js
```diff
@@ -54,11 +54,16 @@ var LegendModelBase = Backbone.Model.extend({
 
   reset: function () {
     var defaults = _.omit(this.defaults(),
-      NON_RESETEABLE_DEFAULT_ATTRS
+      NON_RESETEABLE_DEFAULT_ATTRS,
+      this._avoidResetTheseAttributes()
```
alonsogarciapablo: There's already a mechanism to specify non resettable default attrs (`NON_RESETEABLE_DEFAULT_ATTRS`)... Instead of adding a new mechanism, could we unify them into one? 

```
// in legend-model-base.js
getNonResettableAttrs: function () {
   return ['state', 'visible'];
}
``` 

```
// in a subclass (eg: category-legend-model.js)

getNonResettableAttrs: function () {
  return _.union(
    LegendModelBase.prototype.getNonResettableAttrs.apply(this),
    ['categories']
  );
}

```

Anyway, not resetting these attributes (categories, etc) means that legend models will still be "available" after they are reset... is this what we want?

src/geo/ui/legends/bubble/legend-view.js
```diff
@@ -39,8 +39,18 @@ var BubbleLegendView = LegendViewBase.extend({
...
+    if (first < last) {
+      sizes = sizes.reverse();
+    }
+    return sizes;
+  },
+
   _calculateBubbleSizes: function () {
-    var sizes = this.model.get('sizes').slice(0).reverse();
+    var sizes = this.model.get('sizes').slice(0);
+    sizes = this._reverseSizes(sizes);
```
nobuti: I'll check it out

nobuti: You are right, fixed.

src/geo/map/legends/custom-legend-model.js
alonsogarciapablo: > there is no data returned by the server

That's exactly what I mean. IMO `parse` is meant to "adapt" responses coming form the server when the model is fetched... But here, models are use differently, since they are not connected to an endpoint and hence there's no need to parse IMO...

src/geo/map/legends/bubble-legend-model.js
```diff
@@ -12,6 +12,13 @@ var BubbleLegendModel = LegendModelBase.extend({
     });
   },
 
+  getNonResettableAttrs: function () {
+    return _.union(
+      LegendModelBase.prototype.getNonResettableAttrs.apply(this),
+      ['values']
+    );
+  },
+
```
alonsogarciapablo: 👍

src/geo/ui/legends/bubble/legend-view.js
```diff
@@ -39,8 +39,33 @@ var BubbleLegendView = LegendViewBase.extend({
...
+      values = values.reverse();
+    }
+
+    return {
+      sizes: sizes,
+      values: values
+    };
+  },
+
+  _reverseSizes: function (sizes) {
```
alonsogarciapablo: Can we get rid of this now?

nobuti: Oops

src/geo/ui/legends/bubble/legend-view.js
```diff
@@ -23,7 +23,7 @@ var BubbleLegendView = LegendViewBase.extend({
 
   _getCompiledTemplate: function () {
     return template({
-      labels: this.model.get('values').slice(0).reverse(),
+      labels: this._reverseIfNeeded(this.model).values,
```
alonsogarciapablo: To make it a bit more consistent, I'd call this `this._calculateLabels`, and do something like (`_reverseIfNeeded` is not expressing a clear intention IMO):

```
calculateLabels: function () {
  var labels = this.model.get('values').slice(0);
  if (this._areSizesInAscendingOrder()) {
    labels = labels.reverse();
  }
  return labels;
},

_calculateBubbleSizes: function () {
  var sizes = this.model.get('sizes').slice(0).reverse()
  if (this._areSizesInAscendingOrder()) {
    sizes = sizes.revers();
  }
},

_areSizesInAscendingOrder: function () {
  var sizes = this.model.get('sizes').slice(0);
  return _.first(sizes) < _.last(sizes);
},
...
```

Just an 💡 .

src/geo/map/legends/custom-legend-model.js
nobuti: Not sure if I understand completely. Do you mean something like this?

```
initialize: function (attrs, opts) {
    var attributes = _.extend({},
      _.omit(attrs, 'items')
    );
    attributes.items = [];

    if (attrs.items) {
      attributes.items = attrs.items.map(function (item) {
        return {
          name: item.title,
          color: item.color
        };
      });
    }

    this.set(attributes);
  }
```

If so, I don't like it because we are setting the attributes twice due to Backbone already set attributes before to call `initialize`. And the source code itself defines:
```
if (options.parse) attrs = this.parse(attrs, options) || {};
```
in the Model constructor in order to deal with this scenario. `parse` is defined as a no-op, so if a model doesn't implement it, there is no harm.

src/geo/ui/legends/bubble/legend-view.js
nobuti: 🎆

PR #1445, opened 2016-10-13
Clickable features and "disableable" popups

alonsogarciapablo:
Fixes https://github.com/CartoDB/cartodb.js/issues/1438.

This PR implements a mechanism to:

1. Enable/disable popups (enabled by default):
  ```
  vis.map.enablePopups
  vis.map.disablePopups
  vis.map.arePopupsEnabled
  vis.map.arePopupsDisabled
  ```

2. Enabled/disable "clickable features" (disabled by default).
  ```
  vis.map.enableFeatureInteractivity
  vis.map.disableFeatureInteractivity
  vis.map.isFeatureInteractivityEnabled
  ```

  When "clickable features" are enabled, `vis.map` triggers the following events:
  ```
  featureOver
  featureClick
  featureOut
  ```

Mouse cursor is now changed to `pointer` when "clickable features" are enabled or popups are enabled and any visible layer has popups (infowindow/tooltips).

In order for "clickable features" to work, all visible layers have interactivity enabled by default, and `cartodb_id` is always set as the interactivity field. This will cause the map instantiation to fail if the `sql` of a layer doesn't include the `cartodb_id`. I opened https://github.com/CartoDB/cartodb.js/issues/1446 to discuss/fix this.

@matallo would you please take a look? thx!

### DIFF
--- a/src/geo/cartodb-layer-group-view-base.js
+++ b/src/geo/cartodb-layer-group-view-base.js
@@ -1,7 +1,4 @@
 function CartoDBLayerGroupViewBase (layerGroupModel) {
-  this.visible = true;
-  this.interactionEnabled = [];
-
   layerGroupModel.on('change:urls', this._reload, this);
   layerGroupModel.onLayerVisibilityChanged(this._reload.bind(this));
 }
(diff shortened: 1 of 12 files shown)


matallo:
👏 nothing to add

knownasilya:
How would one use this with `createLayer`?

PR #1447, opened 2016-10-14
Fine tune layer selector

nobuti:
This PR fixes cartodb.js side of https://github.com/CartoDB/cartodb/issues/10093

In addition to the issue's description, the layer selector should only be visible in the settings view in the Builder. That's why we need to segment when the vis is rendered from an embed view (by default) or by Builder.

### DIFF
--- a/src/api/create-vis.js
+++ b/src/api/create-vis.js
@@ -64,9 +64,9 @@ var loadVizJSON = function (el, visModel, vizjsonData, options) {
     showLayerSelector = vizjson.options.layer_selector;
   }
 
-  var isEmbed = true;
-  if (_.isBoolean(options.embedMode)) {
-    isEmbed = options.embedMode;
+  var layerSelectorEnabled = true;
+  if (_.isBoolean(options.layerSelectorEnabled)) {
+    layerSelectorEnabled = options.layerSelectorEnabled;
   }
 
   visModel.set({
@@ -75,13 +75,13 @@ var loadVizJSON = function (el, visModel, vizjsonData, options) {
     apiKey: options.apiKey,
     authToken: options.authToken,
     showEmptyInfowindowFields: options.show_empty_infowindow_fields === true,
-    https: isProtocolHTTPs || options.https === true || vizjson.https === true,
-    isEmbed: isEmbed
+    https: isProtocolHTTPs || options.https === true || vizjson.https === true
   });
 
   visModel.setSettings({
     showLegends: showLegends,
-    showLayerSelector: showLayerSelector
+    showLayerSelector: showLayerSelector,
+    layerSelectorEnabled: layerSelectorEnabled
   });
 
   new VisView({ // eslint-disable-line
--- a/src/api/create-vis.js
+++ b/src/api/create-vis.js
@@ -64,7 +64,7 @@ var loadVizJSON = function (el, visModel, vizjsonData, options) {
     showLayerSelector = vizjson.options.layer_selector;
   }
 
-  var isEmbed = false;
+  var isEmbed = true;
   if (_.isBoolean(options.embedMode)) {
     isEmbed = options.embedMode;
   }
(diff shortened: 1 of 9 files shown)


src/api/create-vis.js
```diff
@@ -64,13 +64,19 @@ var loadVizJSON = function (el, visModel, vizjsonData, options) {
     showLayerSelector = vizjson.options.layer_selector;
   }
 
+  var isEmbed = true;
```
alonsogarciapablo: IMO, we shouldn't use the term `embed` in the context of CartoDB.js.... I'd try to find a more generic name...

src/geo/ui/legends/layer-legends-view.js
```diff
@@ -29,29 +31,67 @@ var LayerLegendsView = Backbone.View.extend({
...
+  _shouldLegendsBeVisible: function () {
+    var showLegends = this.settingsModel.get('showLegends');
+    return showLegends && this._isLayerVisible();
+  },
+
+  _shouldLayerLegendsBeVisible: function () {
+    var isEmbed = this._isEmbed;
+    var showLayerSelector = this.settingsModel.get('showLayerSelector');
+    var showLegends = this.settingsModel.get('showLegends');
+    var isSettingsView = !!this.settingsModel.get('settingsView');
```
alonsogarciapablo: Same here... CartoDB.js shouldn't know anything about "settings"... Are we initialising this anyway?

src/geo/ui/legends/layer-legends-view.js
```diff
@@ -29,29 +31,67 @@ var LayerLegendsView = Backbone.View.extend({
...
+    var hasLegends = this.model.legends.hasAnyLegend();
+    var shouldVisible;
+
+    if (!isEmbed) {
```
alonsogarciapablo: We can simplify this a bit... eg:

```
    var areLayerLegendsVisible = this._isLayerVisible() && showLegends && hasLegends;
    if (isEmbed || isSettingsView) {
      areLayerLegendsVisible = areLayerLegendsVisible || showLayerSelector;
    }
    return areLayerLegendsVisible;
```

src/geo/ui/legends/layer-legends-view.js
nobuti: kind of, `!!undefined === false`, right?

nobuti:
`showLayerSelector` comes from the user's settings, through `layer_selector` option, IMO is a bad idea to mess with these values, they should remain inmutable. I've changed a bit the approach but avoiding to touch the options from settings.

src/geo/ui/legends/layer-legends-view.js
```diff
@@ -29,29 +29,57 @@ var LayerLegendsView = Backbone.View.extend({
...
 
   render: function () {
-    var showLegends = this.settingsModel.get('showLegends');
-    var showLayerSelector = this.settingsModel.get('showLayerSelector');
-    var shouldVisible = showLayerSelector || this.model.legends.hasAnyLegend() && showLegends;
+    var showLegends = this._shouldLegendsBeVisible();
+    var showLayerSelector = this._shouldLayerSelectorBeVisible();
+    var shouldVisible = this._shouldLayerLegendsBeVisible();
+
+    this.$el.html(
```
alonsogarciapablo: I guess, we could skip this if `shouldVisible` is `false`, right?

```
if (this._isVisible()) {
  this.$el.html('');
} else {
  this.$el.html(template({ ... }));
}
```

src/api/create-vis.js
```diff
@@ -75,7 +80,8 @@ var loadVizJSON = function (el, visModel, vizjsonData, options) {
 
   visModel.setSettings({
     showLegends: showLegends,
-    showLayerSelector: showLayerSelector
+    showLayerSelector: showLayerSelector,
+    layerSelectorEnabled: layerSelectorEnabled
```
alonsogarciapablo: How will the Builder change this while on the settings page?

PR #1449, opened 2016-10-17
Set 2 as max digits after .

### LINKED ISSUE #1444: Minimize precission on numbers on legends.
@saleiva commented on [Tue Oct 11 2016](https://github.com/CartoDB/cartodb/issues/10145)

We should use the same approach than in the widgets. Numbers with decimal look very bad,

---

@xavijam commented on [Wed Oct 12 2016](https://github.com/CartoDB/cartodb/issues/10145#issuecomment-253274045)

Moving to cartodb.js.

nobuti:
This PR fixes #1444.

CR @alonsogarciapablo

### DIFF
--- a/src/util/formatter.js
+++ b/src/util/formatter.js
@@ -19,12 +19,10 @@ format.formatNumber = function (value, unit) {
 
   if (abs_v > 100) {
     p = 0;
-  } else if (abs_v > 10) {
-    p = 1;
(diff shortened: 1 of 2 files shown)


nobuti:
It should be nice, but we need to think about the implementation. Right now we have different precision wether the number is bigger than 100, 1 or 0. How it would fit with it?

PR #1450, opened 2016-10-18
Add a get global distribution type function to fix @CartoDB/deep-insi…

donflopez:
…ghts.js#468

With this method we can get global dist type for deep-insights.js auto-style in histogram, to fix @cartodb/deep-insights.js#468


src/dataviews/histogram-dataview-model.js
```diff
@@ -129,6 +129,10 @@ module.exports = DataviewModelBase.extend({
     };
   },
 
+  getGlobalDistributionType: function () {
+    return this.getDistributionType.call(this.getUnfilteredDataModel());
```
javisantana: shouldn't that method be in histogram model (that call changing the this sounds like a hack?)

donflopez:
Changed to allow data as a parameter

javisantana:
:es:

PR #1451, opened 2016-10-18
[WIP] Update Leaflet to v1.

IvanSanchez:
Aparte de que Grunt es un lío, que de no generar sourcemaps es un pifostio, de que hay un par caracteres UTF-8 invisibles (espacios) que joden los parsers de v8, de que Leaflet 1.0 usa unas cabeceras UMD distintas y de que alguien no sabe escribir "switch" correctamente...

...parece que el mayor obstáculo para actualizar Leaflet de la 0.7.x a la 1.0.x va a ser la gestión de eventos. Por algún motivo, los eventos de Wax no se están propagando correctamente a eventos de Leaflet (no se están lanzando el `_manageOnEvents` de las "capas interactivas"), así que de momento toda la interactividad está jodida.


Cartofante:

javisantana:
Hey Ivan, 

Thanks for this (really) but we have several issues here:

- cartodb.js v3 will be replaced by carto.js (v4 branch) so I don't think we want to invest more time on v3
- We have more than 1M maps running with this library so we need to be really careful when updating it, it's so easy to screw up a map.
- There are thousands of apps running with v3 so updating a library means we need to take care of backwards compatibility
- There are also some race condition bugs we were able to workaround (I think fixed now in leaflet 1.0)  and changing the library will make editor (and builder) fail
- We also need to update torque.js and other libraries we use

So unfortunately this is harder than actually updating the dependency

Cartofante:

IvanSanchez:
@javisantana Let me have a quick look at the `v4` branch, then.

PR #1454, opened 2016-10-19
Custom choropleth for pixel aggregation.

nobuti:
This PR fixes https://github.com/CartoDB/cartodb/issues/10224 from cartodb.js side.

custom-choropleth

CR @alonsogarciapablo

### DIFF
--- a/src/geo/ui/legends/custom-choropleth/legend-template.tpl
+++ b/src/geo/ui/legends/custom-choropleth/legend-template.tpl
@@ -1,0 +1,6 @@
+<div class="u-flex u-justifySpace u-bSpace--m">
+  <p class="CDB-Text CDB-Size-small"><%- prefix %> <%- suffix %></p>
+  <p class="CDB-Text CDB-Size-small"><%- prefix %> <%- suffix %></p>
+</div>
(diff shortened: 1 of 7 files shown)


src/geo/ui/legends/custom-choropleth/legend-template.tpl
```diff
@@ -0,0 +1,6 @@
+<div class="u-flex u-justifySpace u-bSpace--m">
+  <p class="CDB-Text CDB-Size-small"><%- prefix %> <%- suffix %></p>
```
alonsogarciapablo: Aren't we missing something here between the prefix and the suffix? Take a look at [this](https://github.com/CartoDB/cartodb.js/blob/0d1d64e9666ce236dcab7b32e8e9c753090b6607/src/geo/ui/legends/choropleth/legend-template.tpl#L2)...

alonsogarciapablo: Otherwise, what's the purpose of this?

nobuti: For heatmaps, we don't know the values, at least it's not trivial to get them, so we are generating only the ramp of colors.

PR #1457, opened 2016-10-20
Fix avg position.

### LINKED ISSUE #1452: Bubble legends not correctly styled in Firefox
screenshot_20161019_140324

STR:
 1. Create a map
 2. Style the fill size by value
 3. Create legend by size

In Firefox, the bubble legend is not correctly aligned at the center.

Please cc me in the PR, as we also may need to do some changes to legend migration.

nobuti:
This PR fixes #1452. Also, it fixes the avg label position for bubbles. In the same manner we did for choropleths, we are including the full range to calculate the position, not assuming that the min value is always 0.

Before:

screen shot 2016-10-20 at 11 59 43

After:

screen shot 2016-10-20 at 12 00 03

CR @alonsogarciapablo @piensaenpixel


piensaenpixel:
\o/

PR #1459, opened 2016-10-20
could you take a lofix 1458

piensaenpixel:
@xavijam ready. Also I consolidate mediaquery :)
zindex

### DIFF
--- a/themes/scss/map/overlays.scss
+++ b/themes/scss/map/overlays.scss
@@ -10,7 +10,7 @@ $buttonColor: #636D72;
   bottom: 16px;
   left: 16px;
   align-items: center;
-  z-index: 20;
+  z-index: 11;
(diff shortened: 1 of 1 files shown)


xavijam:
I love the title! 😂 . :es:

themes/scss/map/overlays.scss
```diff
@@ -10,6 +10,7 @@ $buttonColor: #636D72;
   bottom: 16px;
   left: 16px;
   align-items: center;
+  z-index: 20;
```
xavijam: I'd add a z-index, for example 1.

piensaenpixel: @xavijam it doesn't work with 1, carto logo has 10, but i change to 11 :)

PR #1460, opened 2016-10-20
Include extra param in the sql request when it is false or 0

xavijam:
Just avoiding `null` and `undefined` values.

CR: @alonsogarciapablo 
_Not a big deal_


xavijam:
Related with https://github.com/CartoDB/cartodb-management/issues/4670#event-830501189.

PR #1467, opened 2016-10-25
Check infowindow before getting fields

### LINKED ISSUE #1464: Check infowindow on cartodb.core before calling it
the `CartoDBSubLayer`.`getAttributes` method calls for layer's infowindow, which is not available when using only core library. The error occurs when trying to get interactivity on a core generated layer:

```
cartodb.core.uncompressed.js:3687 Uncaught TypeError: Cannot read property 'get' of undefined
```

The `infowindow` should have its existance checked before calling.

miguelpeixe:
Fix #1464


Cartofante:

PR #1470, opened 2016-10-27
fix custom legend

piensaenpixel:
@xavijam :roll_eyes:

### DIFF
--- a/themes/scss/map/overlays.scss
+++ b/themes/scss/map/overlays.scss
@@ -272,6 +272,15 @@ $maxLegendContainerHeight: 300px;
   margin-bottom: 12px;
 }
 
+
+.CDB-Legend-item {
+  div.cartodb-legend {
+    position: static;
+    box-shadow: 0 0 0;
+    border: 0;
+  }
+}
+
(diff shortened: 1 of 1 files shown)


themes/scss/map/overlays.scss
```diff
@@ -272,6 +272,15 @@ $maxLegendContainerHeight: 300px;
   margin-bottom: 12px;
 }
 
+
+.CDB-Legend-item {
+  div.cartodb-legend {
+    position: static;
```
xavijam: Should we add padding:0?

piensaenpixel: done

themes/scss/map/overlays.scss
```diff
@@ -272,6 +272,14 @@ $maxLegendContainerHeight: 300px;
   margin-bottom: 12px;
 }
 
+
+.CDB-Legend-item div.cartodb-legend {
```
xavijam: Will it override old cartodb-legend properties?

piensaenpixel: yes we only need override those properties

image

xavijam:
Dale neno!

PR #1478, opened 2016-10-28
We have to bind errors in the instantiation of the map

xavijam:
Basically trying to fix [this problem](https://github.com/CartoDB/cartodb/issues/10432), I've realized that if a map is broken, for example, styling by a column, and then applying a query without that column, if we reload the map, the map is not initialized, basically due to the problem that we weren't binded the error in the map instantiation.

cc @alonsogarciapablo @nobuti


xavijam:
Merging when default tests are passed.

PR #1479, opened 2016-10-28
remove hover

piensaenpixel:

hover

cc/ @saleiva


saleiva:
Can we avoid changing the mouse pointer as well?

piensaenpixel:
@alonsogarciapablo also added these minor ui fixes - https://github.com/CartoDB/cartodb/issues/10439

PR #1482, opened 2016-10-31
Take layer visibility into account to determine the mouse cursor

alonsogarciapablo:
If features of an interactive layer were being overed and the layer is hidden, the cursor wasn't being set to `auto` again. This PR solves that issue by setting the cursor back to `auto` if a layer that was being "feature overed" is hidden and no other clickable layer is being "feature overed".

@matallo can you please take a look?


matallo:
👍

PR #1491, opened 2016-11-04
Clean old layer selector code.

nobuti:
This PR remove old code related to the editor's layer_selector.


src/vis/overlays-factory.js
```diff
@@ -154,44 +153,6 @@ OverlaysFactory.register('loader', function (data) {
   return tilesLoader.render();
 });
 
-// layer_selector
-OverlaysFactory.register('layer_selector', function (data, visView, map) {
```
alonsogarciapablo: Warning: If viz.json includes a `layer_selector` overlay, CartoDB.js will [raise an error](https://github.com/CartoDB/cartodb.js/pull/1491/files#diff-31b157993079eb5a1949ceb9a19d29c1L41). I would just empty this function so that CartoDB.js does nothing if it encounters a `layer_selector` overlay, until we're sure we won't be getting one.

xavijam:
There is nothing like remove old code.
🔥

PR #1492, opened 2016-11-09
Only set coordinates when given a GeoJSON with different coordinates

alonsogarciapablo:
To prevent triggering of unnecessary 'change' events.

@matallo please take a look. Thanks!


matallo:
👍

PR #1495, opened 2016-11-09
unset feature events after layer remove

matallo:
tracked down this issue https://github.com/CartoDB/cartodb/issues/10542 until the feature events

despite I could easily have checked the `layerModel` in https://github.com/CartoDB/cartodb.js/blob/v4/src/vis/map-cursor-manager.js#L70 to prevent the `undefined` I thought something could still be not ok, I realized we were firing `featureOut` events on the layer we had removed

I don't really know if this change will bring unintended consequences but looks like unbinding the events on layer remove fixes this, let's discuss the tests tomorrow @alonsogarciapablo 

/cc @xavijam @javisantana @nobuti

### DIFF
--- a/src/vis/feature-events.js
+++ b/src/vis/feature-events.js
@@ -16,6 +16,7 @@ var CartoDBFeatureEvents = function (deps) {
   this._mapModel = deps.mapModel;
 
   this._mapModel.layers.on('add', this._setLayerView, this);
+  this._mapModel.layers.on('remove', this._unsetLayerView, this);
   this._mapModel.layers.on('reset', this._setLayerView, this);
   this._setLayerView();
 };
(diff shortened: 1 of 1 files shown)


nobuti:
LGTM, much better 👍

matallo:
still not ok :/

src/vis/feature-events.js
```diff
@@ -36,6 +37,13 @@ CartoDBFeatureEvents.prototype._setLayerView = function () {
   }
 };
 
+CartoDBFeatureEvents.prototype._unsetLayerView = function () {
+  if (this._layerView) {
+    this._unbindCartoDBFeatureEvents();
```
alonsogarciapablo: All CartoDB layers share the same "layerView", since they are grouped together in a single layer. `_unbindCartoDBFeatureEvents` will actually unbinding events from all CartoDB layers, so I'm afraid this will disable events for every other layer, until a new layer is added or layers are reseted :(

PR #1496, opened 2016-11-10
Do not trigger feature events if layer no longer exists

alonsogarciapablo:
Fixes https://github.com/CartoDB/cartodb/issues/10542.

After a layer is removed, there's a brief period of time in which the the new interaction grids are being requested/loaded and feature events for an inexistent layer model were being triggered. This fixes that issue by checking if the layer still exists before triggering the events.

@matallo please take a look! Thanks!


matallo:
LGTM

:___)

Issue #1499, opened 2016-11-11
Vis/map is being reloaded when infowindow/tooltip fields are just being reordered


alonsogarciapablo:
Part of https://github.com/CartoDB/cartodb.js/issues/1364

## Problem

When infowindow/tooltip fields are re-ordered in the Builder, we reset them in CartoDB.js. Before setting the "new" fields (they are not new, they just have a different order), CartoDB.js checks if we're resetting the same fields (both for [tooltips](https://github.com/CartoDB/cartodb.js/blob/7ddb3928aa5997b41fda9476861acb6749b836d6/src/geo/map/tooltip-template.js#L22) and [infowindows](https://github.com/CartoDB/cartodb.js/blob/7ddb3928aa5997b41fda9476861acb6749b836d6/src/geo/map/infowindow-template.js#L20)). Because, fields have different `positions`, CartoDB.js resets the fields, which trigger a vis/map reload. When same fields are re-ordered multiple times, subsequent equal requests are made and this triggers the error described in https://github.com/CartoDB/cartodb.js/issues/1364.

## Solution

- [x] Ignore the `position` attr of fields when checking if fields are the same.

alonsogarciapablo:
Closed via #1500.

PR #1500, opened 2016-11-11
Use only field names to compare infowindow/tooltip fields when they are updated

alonsogarciapablo:
Fixes https://github.com/CartoDB/cartodb.js/issues/1499. Please read the issue for some context.

@matallo 👍  ? Thanks!


matallo:
👏

PR #1508, opened 2016-11-23
Fixed bug & tests

hacheka:
Fixes a corner case in the SQL.getBounds function: when the API returns no bounds, neither the promise was resolved nor the callback was called.

CR @alonsogarciapablo 
CC @xavijam

### DIFF
--- a/src/api/sql.js
+++ b/src/api/sql.js
@@ -3,6 +3,10 @@ var $ = require('jquery');
 var Mustache = require('mustache');
 var Promise = require('./promise');
 
+var errorMessages = {
+  NO_BOUNDS: 'No bounds'
+};
+
 function SQL(options) {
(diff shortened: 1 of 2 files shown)


src/api/sql.js
```diff
@@ -3,6 +3,10 @@ var $ = require('jquery');
 var Mustache = require('mustache');
 var Promise = require('./promise');
 
+var errorMessages = {
```
alonsogarciapablo: Something I learnt recently... if we use something like `NO_BOUNDS_ERROR_MESSAGE` here (instead of an object), it's easier for us (well, the linter) to catch unused variables.

hacheka: I did it this way to have all error messages (so far we only have one) in an object instead of having a bunch of error messages each in one variable.

Changing it.

PR #1510, opened 2016-11-25
Remove header overlay

xavijam:
Fixes https://github.com/CartoDB/cartodb-management/issues/4708#issuecomment-262475222.
CR: @alonsogarciapablo


alonsogarciapablo:
:fire::fire::fire: should we also remove this from the viz.json/DB?

xavijam:
What's that @alonsogarciapablo ?

xavijam:
Ahh, the DB extension from new CARTO you mean?

xavijam:
that in another PR I feel, 😛

alonsogarciapablo:
Yes, this PR is perfect. I was just wondering if we should make sure this type of overlays is not included in the viz.json anymore...

xavijam:
We could do that, but it will need a viz.json cache flush in order to "clean" that. I'm not sure if it is worth it.

xavijam:
Merging!

PR #1512, opened 2016-11-28
attribution: disabled button

matallo:
CR @piensaenpixel 

easy one, needed for the feature overlay button

screen shot 2016-11-28 at 12 52 34

### DIFF
--- a/themes/scss/map/attributions.scss
+++ b/themes/scss/map/attributions.scss
@@ -40,6 +40,9 @@ $size: 20px;
   vertical-align: middle;
   z-index: 1;
 }
+.CDB-Attribution-button.disabled {
+  opacity: 0.24;
+}
 .CDB-Attribution-text {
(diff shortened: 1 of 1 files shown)


themes/scss/map/attributions.scss
```diff
@@ -40,6 +40,9 @@ $size: 20px;
   vertical-align: middle;
   z-index: 1;
 }
+.CDB-Attribution-button.disabled {
```
piensaenpixel: @matallo I think in cartoassets we have a state for this one. I don't know if it's enough for this one.
image

matallo: not really, those cases are not covered in the .CDB-Button (the icon is a .CDB-IconFont, not a svg, no white button, ...)

screen shot 2016-11-28 at 15 07 23

matallo:
🚀 after CR

PR #1514, opened 2016-11-28
Fix map bounds so it takes considers the biggest dimension

### LINKED ISSUE #1513: setBounds does not correctly select zoom level
This is more visible in maps with datasets that span along a single axis (horizontal or vertical).

The `Map.getBoundsZoom` function calculates the zoom level iterating it while one of the dimensions still fits in the map bounds, that is, it zooms in until both dimension are out of bounds and then zooms out a single step. This chooses a zoom level that covers the smallest dimension.

It should stop zooming as soon as one of the dimensions goes out of bounds, so the whole map is displayed.

javitonino:
Fixes #1513


javitonino:
Please CR @alonsogarciapablo

PR #1515, opened 2016-11-28
Revamp legends html

nobuti:
This PR revamps custom legends to accept html free edition from the Builder. As a consequence, html type is not needed anymore: https://github.com/CartoDB/cartodb/issues/10436

CR @alonsogarciapablo


src/geo/ui/legends/custom/legend-view.js
```diff
@@ -3,9 +3,21 @@ var template = require('./legend-template.tpl');
...
+      htmlTemplate = template({
+        items: this.model.get('items')
+      });
+    } else {
+      htmlTemplate = this._patchStylesForIE(htmlTemplate);
+    }
+    return htmlTemplate;
+  },
+
+  _patchStylesForIE: function (str) {
```
alonsogarciapablo: I'm sure this is here for a reason but... can't we do this in the CSS?

nobuti: We need to insert a css rule in the inline style declaration to make the legend visible for IE11, non sense, I know. IE11 doesn't show the bullet's color if only that style is declared 😞 .

alonsogarciapablo:
I need some more context to get a better understanding of the problem we're trying to solve with this...

nobuti:
Custom legends need to be fully editable, even the HTML, because it's needed for migration of some legends. You can view the ticket here https://github.com/CartoDB/cartodb/issues/10436. So taking this in mind, we are going to delete the readonly `html` type, and merge it with this `custom` type. The `custom` legend has two behaviours:

- if the user uses the form, she/he will be able to make a sort of category color legend, adding or removing colors and labels.
- if the user change to html mode, he will be able to edit the html code without restrictions. 

The rest of the legends will have the same behaviour as they have now: the users can edit the html code adding code before or after the core html code, but they can't edit the core html.

PR #1516, opened 2016-11-28
fix max-height

piensaenpixel:
I know is not the best solution... but for a first approach it works.
Fix https://github.com/CartoDB/cartodb.js/issues/1511


alonsogarciapablo:
@piensaenpixel ping! should we merge this?

PR #1518, opened 2016-11-29
log.error -> log.log

alonsogarciapablo:
Fixes https://github.com/CartoDB/cartodb/issues/10882.

@xavijam 👍  ?

### DIFF
--- a/src/vis/overlays-factory.js
+++ b/src/vis/overlays-factory.js
@@ -36,7 +36,7 @@ var OverlaysFactory = {
     var t = OverlaysFactory._types[type];
 
     if (!t) {
-      log.error("OverlaysFactory: '" + type + "' does not exist");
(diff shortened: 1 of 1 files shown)


src/vis/overlays-factory.js
```diff
@@ -36,7 +36,7 @@ var OverlaysFactory = {
     var t = OverlaysFactory._types[type];
 
     if (!t) {
-      log.error("OverlaysFactory: '" + type + "' does not exist");
+      log.log("Overlays of type '" + type + "' are not supported anymore");
```
xavijam: :+1:, what about using single quotes? 😛

alonsogarciapablo: There's a single quote inside the double quotes, that's why :P

xavijam: :P :P, defeated.

PR #1521, opened 2016-12-01
Clean filter before remove.

nobuti:
This PR clean the dataview's filter before removing it.

remove-widget-filter

CR @alonsogarciapablo 
cc @xavijam


nobuti:
Some tests pending.

src/windshaft/filters/category.js
```diff
@@ -140,5 +140,10 @@ module.exports = WindshaftFilterBase.extend({
 
   getRejectedCategoryNames: function () {
     return this.rejectedCategories.map(function (category) { return category.get('name'); });
+  },
+
+  remove: function () {
+    this.cleanFilter();
```
alonsogarciapablo: From what I see we can decide here if `cleanFilter` should or should not trigger a `change` event (it admits a boolean param)... Do we want it to trigger this event? Does this have any side effect? (I assume not but just in case...)

src/windshaft/filters/range.js
```diff
@@ -27,5 +27,10 @@ module.exports = WindshaftFilterBase.extend({
     };
 
     return json;
+  },
+
+  remove: function () {
+    this.unsetRange();
```
alonsogarciapablo: Same thing here... this will probably trigger a `change` event...

nobuti:
You are right @alonsogarciapablo, sorry for missing a clearer context. Right now, if you have a widget and apply a filter on it, the map obviously reacts showing the data properly. If the user removes that widget, it's removed properly but the filter is still applied to the data. So this PR tries to fix that, clear the filter before deleting the dataview.

alonsogarciapablo:
Thanks for the explanation @nobuti 😄 !  

> If the user removes that widget, it's removed properly but the filter is still applied to the data.

How is that possible? If the widget is removed, its dataview should be removed and the filter shouldn't be sent to the Maps API in the next request ([this is the code where cartodb.js gets the filters that will be sent to the Maps API](https://github.com/CartoDB/cartodb.js/blob/v4/src/dataviews/dataviews-collection.js#L19-L28))...

nobuti:
I think the problem is in how we remove the `dataview-base-model`: https://github.com/CartoDB/cartodb.js/blob/v4/src/dataviews/dataview-model-base.js#L328-L341
If filter is not empty, we reload the vis before to trigger the `remove` event. AFAIK, the event triggered from `this.filter.remove` do nothing because it's not inside a collection and I can't see anything listening that event. So reloading the vis check for filters again, returning the same vis we already have.

alonsogarciapablo:
💡 💡 💡  I think we could simplify this by just doing something [here](https://github.com/CartoDB/cartodb.js/blob/v4/src/vis/vis.js#L323).

```
// src/vis/vis.js

...

_initBindsAfterFirstMapInstantiation: function () {
  ...
  this.listenTo(this._dataviewsCollection, 'remove', this._onDataviewRemoved);
},

...

_onDataviewRemoved: function (dataviewModel) {
  if (dataviewModel.isFiltered()) {
    this.reload({
      sourceId: dataviewModel.getSourceId()
    });
  }
}
```

test/spec/vis/vis.spec.js
```diff
@@ -974,6 +975,36 @@ describe('vis/vis', function () {
       this.vis = new Vis();
     });
 
+    describe('remove dataview', function () {
```
alonsogarciapablo: 'remove dataview' -> 'when a dataview is removed'. As I said in my previous comment, I'd also add a test here: `when a dataview is added`.

test/spec/vis/vis.spec.js
```diff
@@ -974,6 +975,36 @@ describe('vis/vis', function () {
...
+      beforeEach(function () {
+        spyOn(this.vis, 'reload');
+        this.vis.load(new VizJSON(fakeVizJSON()));
+        this.vis.instantiateMap();
+      });
+
+      it('should reload vis', function () {
```
alonsogarciapablo: Also, we could test that the vis is not reloaded when the dataview is not filtered.

alonsogarciapablo:
Much simpler now @nobuti! Thanks!

nobuti:
Revamped test but now it's red because an unrelated tests, and I have no clue why, could you take a look @alonsogarciapablo?

alonsogarciapablo:
Feel free to ignore that test @nobuti. I'll fix it in the future if we encounter it again. Thanks! 👏

alonsogarciapablo:
Or if you want to fix it, we should change [this](https://github.com/CartoDB/cartodb.js/blob/v4/test/spec/geo/geocoder/yahoo-geocoder.spec.js#L5) to something Yahoo will not be able to geocode.

alonsogarciapablo:
retest this please

nobuti:
retest this please

nobuti:
retest this please

nobuti:
🚀

Issue #1539, opened 2016-12-27
[Google Map Support] Prep work for #1527 (II)


alonsogarciapablo:
Extracted from https://github.com/CartoDB/cartodb.js/issues/1537 to keep PRs smaller.

- [x] Remove [div_hack](https://github.com/CartoDB/cartodb.js/blob/90d9ee8aab95b6dcab4fe5eac1edf251486b828c/src/vis/vis-view.js#L42-L52).
- [x] Make sure native maps are correctly destroyed when map views are "cleaned".
- [x] LeafletCartoDBLayerGroupView and GMapsCartoDBLayerGroupView fetch tiles and interactivity automatically (when initialised) if model has URLs.

cc: @rochoa

PR #1541, opened 2016-12-29
More prep work for Google Maps support

alonsogarciapablo:
Part of https://github.com/CartoDB/cartodb.js/issues/1539.

- Map views manage/generate [it's own `div` element](https://github.com/CartoDB/cartodb.js/pull/1541/files#diff-546b824606bbde96318198869eb2de06R7) (`el` was being injected by the `vis-view` before). 
- Instances of CartoDBLayerGroupViewBase (LeafletCartoDBLayerGroupView and GMapsLayerGroupView) automatically [fetch tiles and setup interactivity when they are initialised](https://github.com/CartoDB/cartodb.js/pull/1541/files#diff-394491c3c8e791282436eb57ad5181c6R7) and the layerGroupModel has info (urls) about tiles/grids (eg: when a view is removed and re-rendered).
- Added [a bunch of tests](https://github.com/CartoDB/cartodb.js/pull/1541/files#diff-c1c98c4bc6a55dfe86034033cba4d625) that were missing.

**Acceptance:**

1. Checkout the code and run `grunt dev`.
2. Visit http://localhost:9001/examples/v4/leaflet-layergroup.html.
3. Run the following commands in the console:

```
// 1. remove the map view
mapView.clean();
// 2. you should not see the map any more
// 3. re-add the mapView element to the DOM and render the mapView again:
jQuery('#map').prepend(mapView.el)
mapView.render();
// 4. You should see map again (infowindows and tooltips are not working -> there's an issue for this)
```

### DIFF
--- a/themes/scss/map/map.scss
+++ b/themes/scss/map/map.scss
@@ -1,0 +1,5 @@
+.CDB-Map-wrapper {
+  position: relative;
+  height: 100%;
+  width: 100%;
+}
--- a/examples/v4/leaflet-layergroup.html
+++ b/examples/v4/leaflet-layergroup.html
@@ -32,18 +32,8 @@
           "legends":true,
           "url":null,
           "map_provider":"leaflet",
-          "bounds":[
-            [
-              -18.979025953255253,
-              -155.7421875
-            ],
-            [
-              80.58972691308571,
-              261.2109375
-            ]
-          ],
           "center":"[52.5897007687178, 52.734375]",
-          "zoom":2,
+          "zoom":4,
           "updated_at":"2015-03-13T11:24:37+00:00",
           "datasource": {
             "user_name": "documentation",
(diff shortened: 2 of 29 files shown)


themes/scss/map/map.scss
```diff
@@ -0,0 +1,5 @@
+.CDB-Map-wrapper {
+  position: relative;
+  height: 100%;
```
alonsogarciapablo: cc: @piensaenpixel ☝️

alonsogarciapablo: @piensaenpixel I would ❤️  to have some feedback from you on this change☝️

alonsogarciapablo: This is how the DOM looks like after the latest changes:

screen shot 2017-01-12 at 12 01 13

Issue #1542, opened 2016-12-29
[Google Map Support] Prep work for #1527 (III)


alonsogarciapablo:
Extracted from https://github.com/CartoDB/cartodb.js/issues/1539 to keep PRs smaller.

- [x] Vis is only reloaded when a CartoDB or Torque layer is [added/removed](https://github.com/CartoDB/cartodb.js/blob/90d9ee8aab95b6dcab4fe5eac1edf251486b828c/src/vis/vis.js#L340-L350).

cc: @rochoa

alonsogarciapablo:
## Acceptance:

1. Checkout the code and run `grunt dev`.
2. Open one of the examples: http://localhost:9001/examples/v4/leaflet-layergroup.html.
3. Open the inspector. You will be checking the network tab and the console.
4. Clear the requests in the network tab.
5. Remove and re-add the base layer using the console:
```
var baseLayer = vis.map.getLayerAt(0);
vis.map.layers.remove(baseLayer);
vis.map.layers.add(baseLayer, {at: 0});
```
6. Ensure that no request to the Maps API was made (it'd look something like: `http://documentation.cartodb.com/api/v1/map?config=....`)
7. Remove and re-add the first CartoDB Layer using the console:
```
var cartoDBLayer = vis.map.getLayerAt(1);
vis.map.layers.remove(cartoDBLayer);
vis.map.layers.add(cartoDBLayer, {at: 1});
```
8. Ensure that there's a new request to the Maps API now that a new CartoDB layer has been added.

ethervoid:
Checked and it's working as intended :)

alonsogarciapablo:
Thanks!

PR #1543, opened 2017-01-10
Only reload vis when CartoDB or torque layers are added/removed

alonsogarciapablo:
Part of https://github.com/CartoDB/cartodb.js/issues/1542.

### DIFF
--- a/src/vis/vis.js
+++ b/src/vis/vis.js
@@ -20,6 +20,14 @@ var STATE_INIT = 'init'; // vis hasn't been sent to Windshaft
 var STATE_OK = 'ok'; // vis has been sent to Windshaft and everything is ok
 var STATE_ERROR = 'error'; // vis has been sent to Windshaft and there were some issues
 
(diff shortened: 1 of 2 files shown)


alonsogarciapablo:
Retest this please

src/vis/vis.js
```diff
@@ -20,6 +20,14 @@ var STATE_INIT = 'init'; // vis hasn't been sent to Windshaft
 var STATE_OK = 'ok'; // vis has been sent to Windshaft and everything is ok
 var STATE_ERROR = 'error'; // vis has been sent to Windshaft and there were some issues
 
+var isCartoDBLayer = function (layerModel) {
+  return layerModel.get('type') === 'CartoDB';
```
ethervoid: This should not be cartodb in lower case?

alonsogarciapablo:
@ethervoid no, that's how this type is stored in the DB.

PR #1545, opened 2017-01-10
add support for colored svg images in legends

matallo:
basic support for images was already done (via img tag, however it had a bug), but now legends can have _colored_ images

the approach is pretty similar than icons in fill component of CARTO

screen shot 2017-01-10 at 18 56 25

also, updated legends example


src/geo/ui/legends/base/img-loader-view.js
```diff
@@ -0,0 +1,78 @@
...
+      $img.attr('src', this._icon + '?req=markup');
+      this.$el.empty().append($img);
+    }
+  },
+
+  _requestImage: function (url, callback) {
+    var self = this;
+    var completeUrl = url + '?req=ajax';
+
+    if (this._lastImage.url === completeUrl) {
```
nobuti: are we only caching the last request? why not memoizating all of them?

matallo: because we load the content which is *also* from the last request https://github.com/CartoDB/cartodb.js/pull/1545/files/d75d53509f790ee85330d159b20430e0394961f8#diff-57119bee5bb9bd1ded030b77a0368d16R53

we could make this a hash, storing url and content, and match against the url, but as this is mostly when changing the color of the same icon not to do the request again, it is acceptable in this case to do the request if it is different, even if we already did before

PR #1546, opened 2017-01-11
Filter null values if aggregation is not COUNT for categories dataview.

nobuti:
This PR fixes https://github.com/CartoDB/cartodb/issues/11198

We need to filter the null values if the aggregation is not `COUNT`. We need a reference to the aggregation (through the dataview model). When the aggregation changes, new data is fetched, so when the new data is reset, we filter the models if aggregation meets the condition.

Could you take a look @alonsogarciapablo? Do you see a better approach?

### DIFF
--- a/src/dataviews/category-dataview/categories-collection.js
+++ b/src/dataviews/category-dataview/categories-collection.js
@@ -10,6 +10,27 @@ var CategoryItemModel = require('./category-item-model');
 module.exports = Backbone.Collection.extend({
   model: CategoryItemModel,
 
+  initialize: function (models, options) {
+    this.aggregationModel = options.aggregationModel;
+    this.aggregation = options.aggregationModel.get('aggregation');
+    this.aggregationModel.on('change:aggregation', function (model, aggregation) {
+      this.aggregation = aggregation;
+      this.filterNull();
+    }, this);
+
+    this.on('reset', this.filterNull, this);
+  },
+
+  filterNull: function () {
+    if (this.aggregation === 'count') return;
+
+    var models = this.filter(function (category) {
+      return category.get('value') != null;
+    });
+
+    this.reset(models, {silent: true});
+  },
+
   comparator: function (a, b) {
     if (a.get('name') === 'Other') {
       return 1;
--- a/src/dataviews/category-dataview/categories-collection.js
+++ b/src/dataviews/category-dataview/categories-collection.js
@@ -1,5 +1,7 @@
 var Backbone = require('backbone');
+var _ = require('underscore');
 var CategoryItemModel = require('./category-item-model');
+var COUNT_AGGREGATION_TYPE = 'count';
 
 /**
  *  Data categories collection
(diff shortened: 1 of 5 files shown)


src/dataviews/category-dataview/categories-collection.js
```diff
@@ -10,6 +10,27 @@ var CategoryItemModel = require('./category-item-model');
 module.exports = Backbone.Collection.extend({
   model: CategoryItemModel,
 
+  initialize: function (models, options) {
+    this.aggregationModel = options.aggregationModel;
+    this.aggregation = options.aggregationModel.get('aggregation');
+    this.aggregationModel.on('change:aggregation', function (model, aggregation) {
```
alonsogarciapablo: Question... does this collection get reset when the type of aggregation changes? If that's the case, we might not need this binding.

src/dataviews/category-dataview/categories-collection.js
```diff
@@ -10,6 +10,27 @@ var CategoryItemModel = require('./category-item-model');
...
+      this.aggregation = aggregation;
+      this.filterNull();
+    }, this);
+
+    this.on('reset', this.filterNull, this);
+  },
+
+  filterNull: function () {
```
alonsogarciapablo: Another option could be to override the `reset` method and filter out categories with `null` values. eg:

```
  reset: function(models, options) {
    if (this.aggregation !== COUNT_AGGREGATION_TYPE) {
      models = _.filter(models, function (category) {
        return category.get('value') != null;
      });
      Backbone.Collection.prototype.reset.call(models, options);
    }
  }
```

```

File: test/spec/dataviews/category-dataview/categories-collection.spec.js [hunk 3]
```diff
@@ -0,0 +1,43 @@
...
+    }, {
+      name: 'bar',
+      value: 10
+    }, {
+      name: 'wadus',
+      value: null
+    }]);
+
+    collection.reset.calls.reset();
+    expect(CategoriesCollection.prototype.filterNull).toHaveBeenCalled();
```
alonsogarciapablo: Instead of using spies for this test... why not just test that the collection contains the elements that we are expecting so that the test doesn't know about the internals of the code? eg:

```
expect(collection.pluck('name').toEqual([ 'foo', 'bar', 'wadus' ]);
expect(collection.pluck('value').toEqual([ 1, 10, null ]);
```

or

```
expect(collection.toJSON()).toEqual([{
  name: 'foo',
  value: 1
}, {
  name: 'bar',
  value: 10
}, {
  name: 'wadus',
  value: null
}]);
```

test/spec/dataviews/category-dataview/categories-collection.spec.js
```diff
@@ -0,0 +1,50 @@
...
+    aggregationModel = new Backbone.Model({
+      aggregation: 'count'
+    });
+
+    collection = new CategoriesCollection(null, {
+      aggregationModel: aggregationModel
+    });
+  });
+
+  it('reset should filter null when reset', function () {
```
alonsogarciapablo: To make this test a bit more clear/explicit I'd split it into two tests:

```
describe('.reset', function () {
  it('should NOT filter null values when categories are aggregated by "count"') { ... }
  it('should filter null values when categories are NOT aggregated by "count"') { ... }
});
```

Consider the PR approved after this.

Issue #1544, opened 2017-01-10
[Google Map Support] Prep work for #1527 (IV)


alonsogarciapablo:
- [x] Interactivity (featureClick, featureOver, featureOut events) work properly when mapViews are cleaned and re-rendered (eg: a mapView is removed and re-rendered and infowindows, tooltips, geometry management still work)

**Acceptance:**

1. Checkout `prep-work-for-gmaps-basemaps-4` and run `grunt dev`.
2. Visit http://localhost:9001/examples/v4/leaflet-layergroup.html
3. Verify that:
  - Infowindows and tooltips are working as expected.
  - Mouse cursor changes as features are overed/"un-overed".
  - You can draw on the map:
    ```
    // type this in the console and click on the map
    vis.map.drawPolygon() 
    ```
4. Replace the existing MapView (this is what will happen when basemap changes from Leaflet to Google Maps and viceversa):
```
mapView.clean();
var mapView = new cdb.geo.LeafletMapView({
  mapModel: vis.map,
  visModel: vis,
  layerGroupModel: vis.layerGroupModel
})
jQuery('#map').append(mapView.el);
mapView.render();
```
5. Check point 3 again (infowindows, tooltips, mouse cursor, drawing...).

Thanks!

ethervoid:
Doing the acceptance tests I've found some weird behavior after rendering the new map view:

- Sometimes the Info window shows blank:

![captura de pantalla 2017-01-18 12 49 22 2](https://cloud.githubusercontent.com/assets/741240/22063142/8ccfeb38-dd7d-11e6-8553-11eba1d8cc7e.png)

- When I'm drawing the polygon:
  - First time all went fine, without any issue
  - After the rendering sometimes I have to click twice to draw a point of the polygon

alonsogarciapablo:
Thanks for you comments @ethervoid! My thoughts....

> Sometimes the Info window shows blank:
The infowindow is probably blank because there's a pending request to the attributes service. If you wait a bit the infowindow should probably be filled with the data.

> When I'm drawing the polygon, after the rendering almost all the time, I have to click twice to draw a point of the polygon
I've noticed that too, but not only when the map is re-rendered. I think that this happens when you click and pan the map a bit. If you make sure you just click (making sure the mouse is not moving), it should work.

@ethervoid could you please try ☝️ ? Thanks!

ethervoid:
Checked:

- Blank infowindow are waiting for response
[...]

PR #1549, opened 2017-01-16
MapView takes care of all things interactive

alonsogarciapablo:
Fixes https://github.com/CartoDB/cartodb.js/issues/1544.

Changes:

- VisView is no longer responsible of instantiating managers that were more related to the MapView. Instead a MapView instantiates and starts/stops those managers when initialised/rendered/cleaned. With this change, when a mapView is "replaced" by a new one (eg: when switching from Leaflet to Google Maps), everything (infowindows, tooltips, geometry management works properly again :crossed_fingers:).

- Events triggered by CartoDBLayerGroupViews now group all the event info into a single object. These events are only used internally, but this new format is more clear and also includes the model of the layer that was clicked/overed/"outered".

- In general, most of these managers (specially Infowindow and TooltipManager) are now much simpler since they now work directly with instances of CartoDBLayerGroupViews.

cc: @rochoa


src/geo/map-view.js
```diff
@@ -152,10 +212,23 @@ var MapView = View.extend({
...
+    this._layerViews[layerModel.cid] = this._cartoDBLayerGroupView;
+
+    // Render the infowindow and tooltip "overlays"
+    this._infowindowView.render();
+    this.$el.append(this._infowindowView.el);
+    this._tooltipView.render();
+    this.$el.append(this._tooltipView.el);
+
+    // Start managers that should be bound to the CartoDB layer group view
+    this._infowindowManager.start(this._cartoDBLayerGroupView);
```
rochoa: We need to create the `_*Manager`s with all required dependencies, this `.start(this._cartoDBLayerGroupView);` is showing that `_cartoDBLayerGroupView` is an actual dependency, so the manager doesn't make any sense without it.

src/geo/leaflet/leaflet-cartodb-layer-group-view.js
```diff
@@ -104,17 +104,30 @@ var LeafletCartoDBLayerGroupView = L.TileLayer.extend({
   },
 
   featureOver: function (e, latlon, pixelPos, data, layer) {
-    // dont pass leaflet lat/lon
-    this.trigger('featureOver', e, [latlon.lat, latlon.lng], pixelPos, data, layer);
+    this.trigger('featureOver', {
+      layer: this.model.getLayerAt(layer),
+      layerIndex: layer,
+      latlng: [latlon.lat, latlon.lng],
```
rochoa: What about using an object with properties as we do for the `position` property below.

src/geo/map-view.js
alonsogarciapablo: As discussed with @rochoa, this would be a nice improvement, but it's not trivial so we'll tackle it in the future.

src/geo/leaflet/leaflet-cartodb-layer-group-view.js
alonsogarciapablo: This would also be nice, but we can tackle it in a separate PR as it'd require many additional changes.

alonsogarciapablo:
Retest this please

alonsogarciapablo:
Retest this please!

rochoa:
👍

alonsogarciapablo:
I created a feature branch called `google-maps` and I'll merging all the google maps related PRs to that branch so we can test everything before releasing.

PR #1563, opened 2017-01-30
Adding moveCartoDBLayer function

xavijam:
Related to: https://github.com/CartoDB/cartodb/issues/11137.

In BUILDER, when a layer was moved, we used to remove the layer and add it again in the proper position instead of doing that operation in CARTO.js. Why? I don't know, but let's do it properly!

Also, when a CartoDB layer is moved, we have to refresh the map, of course, thanks reload.

### DIFF
--- a/src/geo/map.js
+++ b/src/geo/map.js
@@ -91,8 +91,10 @@ var Map = Model.extend({
   // PUBLIC API METHODS
 
   moveCartoDBLayer: function (from, to) {
-    this.layers.moveCartoDBLayer(from, to);
-    this.trigger('cartodbLayerMoved', {}, this);
+    var layerMoved = this.layers.moveCartoDBLayer(from, to);
+    if (layerMoved) {
+      this.trigger('cartodbLayerMoved', {}, this);
+    }
   },
 
   createCartoDBLayer: function (attrs, options) {
--- a/src/geo/map.js
+++ b/src/geo/map.js
@@ -90,6 +90,11 @@ var Map = Model.extend({
 
   // PUBLIC API METHODS
 
+  moveCartoDBLayer: function (from, to) {
+    this.layers.moveCartoDBLayer(from, to);
+    this.trigger('cartodbLayerMoved', {}, this);
+  },
+
   createCartoDBLayer: function (attrs, options) {
     this._checkProperties(attrs, ['sql|source', 'cartocss']);
     return this._addNewLayerModel('cartodb', attrs, options);
--- a/src/geo/map/layers.js
+++ b/src/geo/map/layers.js
@@ -78,7 +78,7 @@ var Layers = Backbone.Collection.extend({
     if (from === to) {
       return this;
     }
-    
+
     var tempLayerModel = this.at(from);
 
     if (tempLayerModel.get('type') !== CARTODB_LAYER_TYPE) {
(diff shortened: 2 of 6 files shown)


src/geo/map.js
```diff
@@ -90,6 +90,13 @@ var Map = Model.extend({
 
   // PUBLIC API METHODS
 
+  moveCartoDBLayer: function (from, to) {
+    var layerMoved = this.layers.moveCartoDBLayer(from, to);
+    if (layerMoved) {
+      this.trigger('cartodbLayerMoved', {}, this);
```
alonsogarciapablo: What about `this.trigger('cartodbLayerMoved', layerMoved);`?

src/geo/map/layers.js
```diff
@@ -72,6 +72,26 @@ var Layers = Backbone.Collection.extend({
...
       return !!layerModel.legends;
     });
+  },
+
+  moveCartoDBLayer: function (from, to) {
+    if (from === to) {
+      return false;
+    }
+
+    var tempLayerModel = this.at(from);
```
alonsogarciapablo: What about `movingLayer`?

test/spec/geo/map.spec.js
```diff
@@ -191,6 +191,21 @@ describe('core/geo/map', function () {
...
   });
 
+  describe('Layer action methods', function () {
+    describe('.moveCartoDBLayer', function () {
+      it('should provide a way to move CartoDB layers', function () {
+        expect(this.map.moveCartoDBLayer).toBeDefined();
+      });
+
+      it('should trigger an event in order to notify other elements', function () {
+        spyOn(this.map, 'trigger');
```
alonsogarciapablo: Instead of spying on `trigger`, we can bind an event and ensure the callback is invoked:

```
var moveCartoDBLayer = jasmine.createSpy('moveCartoDBLayer');
this.map.on('moveCartoDBLayer', callback);

this.map.moveCartoDBLayer(1, 2);

expect(moveCartoDBLayer).toHaveBeenCalledWith('wadus');
```

src/vis/vis.js
```diff
@@ -171,6 +171,8 @@ var VisModel = Backbone.Model.extend({
       layersFactory: layersFactory
     });
 
+    this.listenTo(this.map, 'cartodbLayerMoved', this.reload);
```
alonsogarciapablo: Can we add a test for this 🙏 ?

src/geo/map.js
xavijam: Because with listenTo, the bind is removed when Class is eliminated (http://stackoverflow.com/questions/16823746/backbone-js-listento-vs-on).

src/geo/map/layers.js
xavijam: :+1:

src/vis/vis.js
xavijam: Sure!

test/spec/geo/map.spec.js
xavijam: 👍

xavijam:
Feedback applied!

alonsogarciapablo:
:+1: when you add [this test](https://github.com/CartoDB/cartodb.js/pull/1563#discussion_r98483438)

xavijam:
man, you are right, I totally forgot it.

xavijam:
Added!

xavijam:
deal @alonsogarciapablo?

alonsogarciapablo:
ship it! 👍

xavijam:
💋

PR #1566, opened 2017-01-30
Remove d3 vector implementation and reuse the vector flag

donflopez:


src/geo/leaflet/leaflet-cartodb-vector-layer-group-view.js
```diff
@@ -1,73 +0,0 @@
-require('d3.cartodb');// TODO: The 'd3.cartodb' module doens't currently export L.CartoDBd3Layer
-// and it's currently relying on window.L so weed to do the following trick.
-// Check out: https://github.com/CartoDB/d3.cartodb/issues/93 for more info
-var CartoDBd3Layer = window.L.CartoDBd3Layer;
-var LeafletLayerView = require('./leaflet-layer-view');
-var GeoJSONDataProvider = require('../data-providers/geojson/data-provider');
```
alonsogarciapablo: We can remove this class/file too :point_up:

donflopez: Yeah, thank you Pablo!

donflopez: Done! 👍

alonsogarciapablo:
Just a comment... [Setting the data provider on a layer](https://github.com/CartoDB/cartodb.js/blob/v4/src/geo/map/cartodb-layer.js#L109) was only being used by our previous implementation of vector layers (which we're removing in this PR). We could remove this mechanism completely (`CartoDBLayer#setDataProvider`, `TorqueLayer#setDataProvider`, `CartoDBLayer#getDataProvider`, `TorqueLayer#getDataProvider`, and references to them) now that is not being used anymore and rethink it again (I have the feeling that we're missing some abstractions. eg: [in this case](https://github.com/CartoDB/cartodb.js/blob/v4/src/dataviews/dataview-model-base.js#L106-L116)). cc: @javisantana @rochoa

javisantana:
I'd not remove it, we'll need it in the next phase. The abstractions were right from my point of view, they were working. Please, do not confuse wrong abstractions with wrong implementations.

Anyway, really open to hear proposals about this and how do you think that should work

donflopez:
Can we move this topic to another issue so that we can start to think and discuss it?

I'm gonna merge it.

Thanks, Pablo!! :-)

alonsogarciapablo:
Sure thing @donflopez. I just wanted to raise my concerns. We can leave it and revisit it when we start using the code again. Just to clarify, the fact that we have to do things like (mostly inside our dataview base class):

 ```
    if (layerDataProvider && layerDataProvider.something(...)) {
      // do something
    } else {
      // do something else
    },
```

IMHO, this is telling us that we're missing a "default" data provider or perhaps a pool of data providers, so we can get rid of this kind of `if` statements.

javisantana:
Good point @alonsogarciapablo, agree on having a default data provider. Let me add something: in some cases the data provider should be really fine grained. Let me put an example: if we have a histogram widget we might want to fetch data from server (default provider) to calculate buckets and then use local data to fill the viewport histogram.

alonsogarciapablo:
> if we have a histogram widget we might want to fetch data from server (default provider) to calculate buckets and then use local data to fill the viewport histogram.

Yes! Perhaps the [Chain-of-responsibility pattern](https://en.wikipedia.org/wiki/Chain-of-responsibility_pattern) would work here. We currently have some commands (`applyFilter`, `getData`) and some providers ("default" and vector...), and clients (dataviews mainly) could send those commands to the chain of providers without having to know who will satisfy the request.

PR #1569, opened 2017-02-06
Get tile url by type (png or mvt)

donflopez:
This PR fixes https://github.com/CartoDB/tangram-carto/issues/28

It allows tangram to get protected MVT.

### DIFF
--- a/src/geo/cartodb-layer-group.js
+++ b/src/geo/cartodb-layer-group.js
@@ -68,7 +68,6 @@ var CartoDBLayerGroup = Backbone.Model.extend({
   },
 
   _replaceLayerIndexesOfVisibleLayers: function (type, url) {
-
     if (type === 'png') {
       var remoteIndexesOfVisibleLayers = _.reduce(this.getLayers(), function (indexes, layerModel, layerIndex) {
         if (layerModel.isVisible()) {
@@ -78,8 +77,7 @@ var CartoDBLayerGroup = Backbone.Model.extend({
       }.bind(this), []);
 
       return url.replace('{layerIndexes}', remoteIndexesOfVisibleLayers.join(','));
-    }
-    else if (type === 'mvt') {
+    } else if (type === 'mvt') {
       return url
         .replace('{layerIndexes}', 'mapnik')
         .replace('.png', '.mvt');
--- a/src/geo/cartodb-layer-group.js
+++ b/src/geo/cartodb-layer-group.js
@@ -46,15 +46,19 @@ var CartoDBLayerGroup = Backbone.Model.extend({
     return !!this.get('urls');
   },
 
-  getTileURLTemplates: function () {
+  getTileURLTemplates: function (type) {
+    type = type || 'png';
+
     if (this._areAllLayersHidden()) {
       return [];
     }
     var tileURLTemplates = (this.get('urls') && this.get('urls').tiles) || [];
-    return _.chain(tileURLTemplates)
-      .map(this._replaceLayerIndexesOfVisibleLayers.bind(this))
+    var urls = _.chain(tileURLTemplates)
+      .map(this._replaceLayerIndexesOfVisibleLayers.bind(this, type))
       .map(this._appendAuthParamsToURL.bind(this))
       .value();
+
+    return type === 'mvt' ? urls && urls[0] : urls;
   },
 
   _areAllLayersHidden: function () {
@@ -63,14 +67,23 @@ var CartoDBLayerGroup = Backbone.Model.extend({
     });
   },
 
-  _replaceLayerIndexesOfVisibleLayers: function (url) {
-    var remoteIndexesOfVisibleLayers = _.reduce(this.getLayers(), function (indexes, layerModel, layerIndex) {
-      if (layerModel.isVisible()) {
-        indexes.push(this.get('indexOfLayersInWindshaft')[layerIndex]);
-      }
-      return indexes;
-    }.bind(this), []);
-    return url.replace('{layerIndexes}', remoteIndexesOfVisibleLayers.join(','));
+  _replaceLayerIndexesOfVisibleLayers: function (type, url) {
+
+    if (type === 'png') {
+      var remoteIndexesOfVisibleLayers = _.reduce(this.getLayers(), function (indexes, layerModel, layerIndex) {
+        if (layerModel.isVisible()) {
+          indexes.push(this.get('indexOfLayersInWindshaft')[layerIndex]);
+        }
+        return indexes;
+      }.bind(this), []);
+
+      return url.replace('{layerIndexes}', remoteIndexesOfVisibleLayers.join(','));
+    }
+    else if (type === 'mvt') {
+      return url
+        .replace('{layerIndexes}', 'mapnik')
+        .replace('.png', '.mvt');
+    }
   },
 
   hasTileURLTemplates: function () {
--- a/src/geo/leaflet/leaflet-cartodb-webgl-layer-group-view.js
+++ b/src/geo/leaflet/leaflet-cartodb-webgl-layer-group-view.js
@@ -23,7 +23,9 @@ var LeafletCartoDBWebglLayerGroupView = L.Class.extend({
 
   initialize: function (layerGroupModel, map) {
     LeafletLayerView.call(this, layerGroupModel, this, map);
-    layerGroupModel.bind('change:urls', this._onURLsChanged, this);
+    layerGroupModel.bind('change:urls',
+      this._onURLsChanged(layerGroupModel.getTileURLTemplates.bind(layerGroupModel))
+    );
 
     this.tangram = new TC(map);
 
@@ -42,12 +44,11 @@ var LeafletCartoDBWebglLayerGroupView = L.Class.extend({
 
   setZIndex: function () {},
 
-  _onURLsChanged: function (e, res) {
-    var url = res.tiles[0]
-      .replace('{layerIndexes}', 'mapnik')
-      .replace('.png', '.mvt');
-
-    this.tangram.addDataSource(url);
+  _onURLsChanged: function (getUrl) {
+    var self = this;
+    return function () {
+      self.tangram.addDataSource(getUrl('mvt'));
+    };
   }
 });
 
--- a/test/spec/geo/leaflet/leaflet-cartodb-webgl-layer-group-view.spec.js
+++ b/test/spec/geo/leaflet/leaflet-cartodb-webgl-layer-group-view.spec.js
@@ -19,7 +19,6 @@ describe('src/geo/leaflet/leaflet-cartodb-webgl-layer-group-view.js', function (
 
     this.layerGroupModel.set('urls', { tiles: [] });
 
-
     this.vis = new VisModel();
 
     this.layer1 = new CartoDBLayer({ id: 'layer1', meta: {cartocss: '#layer {polygon-fill: black;}'} }, { vis: this.vis });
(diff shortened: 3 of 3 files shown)


src/geo/cartodb-layer-group.js
```diff
@@ -46,15 +46,19 @@ var CartoDBLayerGroup = Backbone.Model.extend({
     return !!this.get('urls');
   },
 
-  getTileURLTemplates: function () {
+  getTileURLTemplates: function (type) {
+    type = type || 'png';
```
ethervoid: There is no better way to check for type of tile than the extension? maybe it'll be useful to have some constant in case we have more extensions or whatever

src/geo/cartodb-layer-group.js
```diff
@@ -63,14 +67,21 @@ var CartoDBLayerGroup = Backbone.Model.extend({
...
+        if (layerModel.isVisible()) {
+          indexes.push(this.get('indexOfLayersInWindshaft')[layerIndex]);
+        }
+        return indexes;
+      }.bind(this), []);
+
+      return url.replace('{layerIndexes}', remoteIndexesOfVisibleLayers.join(','));
+    } else if (type === 'mvt') {
+      return url
+        .replace('{layerIndexes}', 'mapnik')
```
ethervoid: Just to know more about this, why are we doing this?

src/geo/cartodb-layer-group.js
```diff
@@ -46,15 +46,19 @@ var CartoDBLayerGroup = Backbone.Model.extend({
...
     }
     var tileURLTemplates = (this.get('urls') && this.get('urls').tiles) || [];
-    return _.chain(tileURLTemplates)
-      .map(this._replaceLayerIndexesOfVisibleLayers.bind(this))
+    var urls = _.chain(tileURLTemplates)
+      .map(this._replaceLayerIndexesOfVisibleLayers.bind(this, type))
       .map(this._appendAuthParamsToURL.bind(this))
       .value();
+
+    return type === 'mvt' ? urls && urls[0] : urls;
```
ethervoid: Why do we return only the index 0 for mvt tiles?

ethervoid:
Despite some comments, PR looks good to me

src/geo/cartodb-layer-group.js
donflopez: Because Tangram doesn't need an array of tiles URLs, it's always the same URL as a source and it controls internally the tile requests.

src/geo/cartodb-layer-group.js
donflopez: Raster tiles have an index of the layer if a map has multiples of them... MVT doesn't, all of them go in the protobufer and they are parsed by tangram.

src/geo/cartodb-layer-group.js
donflopez: Yeah, I agree with that, but maybe this is not the site... I think that the URL don't have to be formed with the png, and then replace it. It should be well formed from the root.

What do you think?

donflopez:
Thank you @ethervoid! Cartodb have a coupled behaviour with raster tiles and it is a little bit tricky to have an integration with vector, I'll make a research to know where is the root of the tile URL system and come back with a better approach (less tricky, having a more abstract source URL implementation).

donflopez:
After some research, I find that CDB.js generates tile URLs [here](https://github.com/CartoDB/cartodb.js/blob/v4/src/windshaft-integration/model-updater.js#L43). This implied that we only are creating png and grids tile urls, but no for other formats and that the libraries (leaflet & the grid loader) are coupled to the model.

What I think is a better approach is give to them or the wrappers, the resources to build his URLs and avoid to pre-format them even if don't use them.

But that implies majors changes to cartodb.js and others modules as contra.

alonsogarciapablo:
@donflopez as you found out... we're only generating tile urls for raster tiles. IMO this is a good opportunity to change [things](https://github.com/CartoDB/cartodb.js/blob/v4/src/windshaft-integration/model-updater.js#L54-L61) and give support to other formats from a "deeper" point in the app.

donflopez:
Sure @alonsogarciapablo, but that's gonna need time. It implies a change in the model scheme from my point of view and that seems to be pretty hard, IMHO.

donflopez:
What do you think we should do?

I see here two approaches:

- **Follow the same structure used:**

Add a vector property next to grid and raster, more easy but coupled and doing unnecessary work IMO.

-**Change the model schema:**

Avoid have in the model the formed urls and anything coupled with the modules, in this examples, we only need the domain, user and map id (I think), and have them into the modules so each one can build the URLs as they want to (we can have a shared library to easily build them).

alonsogarciapablo:
Tile urls are generated [here](https://github.com/CartoDB/cartodb.js/blob/v4/src/windshaft-integration/model-updater.js#L57) and they are only used [here](https://github.com/CartoDB/cartodb.js/blob/v4/src/geo/cartodb-layer-group.js#L49-L58). One idea:

We're generating url templates that look like ".../{layerIndexes}/{z}/{x}/{y}.png". What if we generate something like "...//{layerIndexes}/{z}/{x}/{y}.{format}' [here](https://github.com/CartoDB/cartodb.js/blob/v4/src/windshaft-integration/model-updater.js#L64)? Then we can replace that "{format}" [here](https://github.com/CartoDB/cartodb.js/blob/v4/src/geo/cartodb-layer-group.js#L54-L57).

donflopez:
But some of them need an array of URLs and others don't, some of them need the index === {layerIndexes} calculated with the subdomain and others don't, and we are precalculating (maybe) unnecessary things, what if we pass the raw info to them?

If sometime in the future we have to add a holographic source URLs we are gonna need to modify this and make the transpiler of the URL in the module integration and finally use the URL in our holographic library, what if we build the URL with the raw info?

src/geo/cartodb-layer-group.js
ethervoid: I agree with you, the `png` swap if a bit of bad smell

alonsogarciapablo:
> But some of them need an array of URLs and others don't, some of them need the index === {layerIndexes} calculated with the subdomain and others don't, and we are precalculating (maybe) unnecessary things, what if we pass the raw info to them?

Can you please elaborate a bit more? Who are "them"? Some links to code would help... thanks!

ethervoid:
Could be possible to add or have information about the tile format, even as a template, in the `windshaftmap` dependency [here](https://github.com/CartoDB/cartodb.js/blob/v4/src/windshaft-integration/model-updater.js#L41) and inside this function, call a tiles factory or something like that.

Sorry if I'm saying something dumb, I'm a bit out of my league xD

donflopez:
Yeah, sorry @alonsogarciapablo!

[Here](https://github.com/CartoDB/cartodb.js/blob/v4/src/windshaft-integration/model-updater.js#L44) we're precalculating the grids, tiles and attributes... all of them are based on some shared data, but others no AFAIK. For example:

[Grids](https://github.com/CartoDB/cartodb.js/blob/v4/src/windshaft-integration/model-updater.js#L83) needs the index predefined by [indexes of the mapnik layers](https://github.com/CartoDB/cartodb.js/blob/v4/src/windshaft-integration/model-updater.js#L71), however, tiles don't... they change the flag by [indexes of the visible layers](https://github.com/CartoDB/cartodb.js/blob/v4/src/geo/cartodb-layer-group.js#L73) in the layer group model.

Both of them needs an array of URLs, but tangram doesn't, it only needs one URL as a source.

So we are handling attributes, grids, tiles, and now vector... in a model updater making the model mutable and coupled with other external modules IMHO. What if our model doesn't know about which kind of technology is using it, what if we have the raw data and the glue between our core and the modules (tangram, leaflet, etc...) is in charge of transforming the raw into a proper URL or URLs?

I see the good point in your idea, maybe we can store that very general URL, but we don't know if that generic URL [is gonna be so generic](https://github.com/CartoDB/cartodb.js/blob/v4/src/windshaft-integration/model-updater.js#L97).

What do you think?

donflopez:
Ey, let's gonna push this!

layer-url-handling

If we have the raw information in the model, we can add any layer type glue without the needing of changing the model-updater every time we add or remove a new type.

It's the glue type the manager of adapt the viz information to the component glued.

Why do we need a model updater? What do you think guys? Does it make sense to you?

alonsogarciapablo:
It looks, from looking at the code, that we didn't make any change to append the authToken parameter to the URLs... Aren't we just moving the logic to "hack" the URL from ` src/geo/leaflet/leaflet-cartodb-webgl-layer-group-view.js` to `src/geo/cartodb-layer-group.js`?

As I said in https://github.com/CartoDB/cartodb.js/pull/1569#issuecomment-278020700, I would try to find a better solution, so that we don't need to "hack" a URL template that we're generating and have complete control over it. Otherwise, these changes make things a bit more complicated (IMHO).

@donflopez Could you please elaborate more what you're describing in https://github.com/CartoDB/cartodb.js/pull/1569#issuecomment-282680156? I too think there might be a better solution, so please, by all means, keep pushing on your ideas.

About your question:

> Why do we need a model updater? What do you think guys? Does it make sense to you?

The ModelUpdater class is basically something that "glues" Windshaft and CartoDB.js (some sort of facade between both systems). For example, when a "map" is instantiated it knows how to generate some URLs (tiles, grids, attributes, etc.) so that models (CartoDBLayerGroupModel in this case) doesn't need to know how to generate them.

donflopez:
> It looks, from looking at the code, that we didn't make any change to append the authToken parameter to the URLs

Yes, but is in the `cartodb-layer-group` where we know the token and where the auth logic is, I didn't want to duplicate the logic. I don't want to have that logic there, I think that the logic should be at the library glue level, passing to it not the model or models, but the necessary data, like domain, token, layers... I could simply get the token from the model in the webgl code, but it coupled more the code to the model, so I did in that way... But I didn't like it.

What I mean in my last comment is that IMHO could be a good idea to move the URL generation to each module and we don't have it represented in the model, just each one can generate is own URL based on his own necesities. Making it we don't need a model-updater and we don't also need a technology based url, stored in the model. I've represented that the changes happens subscribing the module to the on-event of the models, is like it is now, but what I proposed in the leaflet 1.0 is different.

By the way, make the things as I propose is a hard task that maybe we don't want to accomplish now, but when the new definition of Carto.js is available.

What do you think?

alonsogarciapablo:
Ok. I wasn't being aware of the fact that the view was using `url` directly, instead of using the method that we're changing here. Let's move ahead with this fix and revisit this to find a nicer and more flexible solution 👍

rafatower:
What does it take to release this?

I see there's a dependency declared here: https://github.com/CartoDB/cartodb/blob/25424e8481b05740eff9a21c6360d60ee66ce41c/package.json#L27

Is it enough to merge and wait for (or force) a deploy of CartoDB repo?

donflopez:
It is ready to merge! I'll do it :)

rafatower:
> It is ready to merge! I'll do it :)

thanks! :) :+1: 

I asked because I wanted to learn and deploy it myself :smile: 

> is it enough to merge and wait for (or force) a deploy of CartoDB repo?

so... am I right? just to know for the next time, buddy

donflopez:
Ah, sure!! You're right!

You can make the deploy now if want @rafatower :)

rafatower:
I just tested this in prod after deploying it and I'm getting this, following the STR:
```
{"errors":["Sorry, you are unauthorized (permission denied)"],"errors_with_context":[{"type":"unknown","message":"Sorry, you are unauthorized (permission denied)","http_status":403}]}
```

Either I did not deploy it rigth or it is not fixed :(

donflopez:
I think that it is not deployed correctly if you go to the source code, there is not any change.

How did you do the deploy @rafatower?

Shor explanation:

- Go cartodb project
- update npm version
- npm run-script update-internal-deps
- check the npm-shrinkwrap to see that everything is correct
     - if there is any other change, ask to the author
- push the update to master
- make the deploy to production with jenkins

rafatower:
Okay, I did not do that :sweat_smile: . I just assumed a new deploy would build and push the changes taking `v4` as the source. That reminds me of some wiki explaining just that... I'll try to locate it to keep it under my pillow.

Thanks!

donflopez:
You're welcome @rafatower! Whatever you need 🙂 

BTW: this is gonna be the most trending PR of Carto 😂

rafatower:
I'll do the deploy on Monday, unless somebody is in a rush.

PR #1573, opened 2017-02-10
Fix zoom loop

### LINKED ISSUE #1571: [Google Map Support] Zoom changes automatically after basemap provider changes and window is resized
**STR:**

- Open a map.
- Change the basemap (Leaflet -> Google Maps or viceversa).
- Resize the window.

**Expected results:**
- Zoom should not change.

**Actual results:**
- Zoom is changed (to a very hight number).

donflopez:
This fix #1571


src/geo/map-view.js
```diff
@@ -157,12 +157,12 @@ var MapView = View.extend({
 
   /** unbind model properties */
   _unbindModel: function () {
-    this.map.unbind('change:view_bounds_sw', this._changeBounds, this);
-    this.map.unbind('change:view_bounds_ne', this._changeBounds, this);
-    this.map.unbind('change:zoom', this._setZoom, this);
-    this.map.unbind('change:scrollwheel', this._setScrollWheel, this);
-    this.map.unbind('change:keyboard', this._setKeyboard, this);
-    this.map.unbind('change:center', this._setCenter, this);
+    this.map.unbind('change:view_bounds_sw', null, this);
```
alonsogarciapablo: Was the fact that we're specifying a callback the culprit of this?

```
 If no callback is specified, all callbacks for the event will be removed
```

Did you make sure no other objects should keep listening to these events?

donflopez:
Originally, we were cleaning all the events because there was a problem with the references.

This problem was pretty hard to fix, because of the hidden behaviour.

The problem was:

Binding some functions in a parent class that were being overwritten by the inherited class, that parent class has the property as a prototype property. The bind happens with the prototype function, but after that, we were binding the function to the child class, extended with the parent class, so the function is also a general property.

Then, we launch an event to set a property that needs to unbind the events, but the function is not the prototype reference, it is a common property of an object, so they don't match, and it is not unbound but also bound again, making a loop.

**Detailed Explanation**

fullsizerender

Steps:

1.- [Extend](https://github.com/CartoDB/cartodb.js/blob/switch-to-gmaps/src/geo/leaflet/leaflet-map-view.js#L7) from MapView and create a LeafletMapView instance
2.- [Call MapView (parent class) init](https://github.com/CartoDB/cartodb.js/blob/switch-to-gmaps/src/geo/leaflet/leaflet-map-view.js#L9) with the context of LeafletMapView instance
3.- MapView [bind some functions] from his child context, v8 look at the object properties, they don't exist so look at the prototype chain, they exist, binding.
4.- LeafletMapView inst, [bind some prototypes](https://github.com/CartoDB/cartodb.js/blob/switch-to-gmaps/src/geo/leaflet/leaflet-map-view.js#L11) functions as properties.
5.- Someone changes a property that [triggers a MapView event](https://github.com/CartoDB/cartodb.js/blob/switch-to-gmaps/src/geo/map-view.js#L135).
6.- This event [unbind the 3step](https://github.com/CartoDB/cartodb.js/blob/switch-to-gmaps/src/geo/map-view.js#L160), functions, but now it look at the object properties and there are them set by 4, so it is trying to unbind differents functions and then [binding them again](https://github.com/CartoDB/cartodb.js/blob/switch-to-gmaps/src/geo/map-view.js#L150).

So there are 2 same functions with different references.

xavijam:
👏  for the explanation man!

alonsogarciapablo:
Thanks for the research @donflopez! From what I see, we don't have [these bindings](https://github.com/CartoDB/cartodb.js/pull/1573/files#diff-aaf2180ce09425b310a1186ac37bab76R9) in [GmapsMapView#initialize](https://github.com/CartoDB/cartodb.js/blob/switch-to-gmaps/src/geo/gmaps/gmaps-map-view.js#L9-L11). Perhaps we could even get rid of them?

donflopez:
YW! Yep, we don't have them, but if there is a similar use for _ready, it will happen anyway. We should avoid those scenarios.

alonsogarciapablo:
I checked and I think we can get rid of `LeafletMapView#initialize` (all those methods are bound with the right context in the parent class). We should apply your approach to GMapsMapView#initialize:

```
var GoogleMapsMapView = MapView.extend({
  initialize: function () {
    _.bindAll(this, '_ready');
    this._isReady = false;
    MapView.prototype.initialize.apply(this, arguments);
  },
  ...
```

Do you agree @donflopez? Do you have any other alternative to solve this? Thanks!

donflopez:
Yeah, I agree! I'll apply that and we have to be aware of this kind of things... Why do we need [those bindings](https://github.com/CartoDB/cartodb.js/blob/switch-to-gmaps/src/geo/leaflet/leaflet-map-view.js#L11)?

alonsogarciapablo:
> Why do we need those bindings @alonsogarciapablo?

We don't need them from what I saw. Let's 🔥  them!

donflopez:
Done!! 🎉

alonsogarciapablo:
👍  Ship it!

src/geo/gmaps/gmaps-map-view.js
```diff
@@ -70,6 +70,11 @@ var GoogleMapsMapView = MapView.extend({
     this.projector.draw = this._ready;
   },
 
+  clean: function () {
+    google.maps.event.clearInstanceListeners(window);
+    google.maps.event.clearInstanceListeners(document);
```
alonsogarciapablo: We're overriding `MapView.prototype.clean` here, so we should have invoked `MapView.prototype.clean.call(this);` here. I'm fixing it, but I just wanted to pointed it out since we didn't notice.

donflopez: Perfect! Thanks @alonsogarciapablo!

Issue #1503, opened 2016-11-16
Migrate leaflet to 1.x series


chriswhong:
Hi there, are there any plans to update cartodb.js to use leaflet 1.0.1? I tried using `https://cartodb-libs.global.ssl.fastly.net/cartodb.js/v3/3.15/cartodb_noleaflet.js` today with leaflet v1.0.1 with an existing cartodb.js map and got the error:

`cartodb_noleaflet.js:16 Uncaught TypeError: this._adjustTilePoint is not a function(…)`

I would really like to use `map.flyTo()` and the new tooltips in my carto projects.

Thanks!

knownasilya:
Ping

knownasilya:
FYI if to work around the layer not working at all, just add a blank function, e.g.

```js
layer._adjustTilePoint = () => {};
```

allthesignals:
I would say this compatibility is pretty important for CartoDB.js users who need to use other Leaflet plugins that often require Leaflet 1+.

PR #1583, opened 2017-02-20
Migrate to leaflet 1.x

donflopez:
#1503


donflopez:
There are all minimum changes to make leaflet 1.x works on cartodb.js. Some of them are exclusively made for passing the tests.
They allow us to pass the tests but fire some errors when playing with the examples.

The event handling design of cartodb.js mixing Backbone events and Leaflet events, both of them working over `_events` property, make this migration *impossible* without a deep refactor of how Cartodb.js works with both of them.

## About Carto.js

I'd rethink the Cartodb.js architecture, avoiding inheritance (we have a double coupling with Leaflet and Backbone) and even avoiding the use of Backbone (do we need a "framework" in cartodb.js?).

From my point of view (having into account that I'm a noob with cartodb.js), cdb.js has to be an api to generalise the work with other opensource and the CARTO data management but actually, cartodb.js is owning that opensource and data, not using it.

### **Rendering**

We don't need cartodb.js painting anything, if we want legends, let's have a legend library to paint them with and input (data) and an output (dom rendering, even HTML). If we want to have legends out of a map (ex: a newspaper sidebar) we can use that library.

### **Data Handling**

Leave each module has his model, we don't need to have access to the layer model from the legend view. We can have an API to add, edit and remove legend rows (a bad example) and leave the knowledge of the Cartodb.js scheme to cdb.js, even more, leave the knowledge of the Cartodb.js scheme to a store, not coupled models, so anyone (another project) can easily use that store to get, set and observe data from there.

We shouldn't pass the model and use it as a source of events and everyone mutating them without rules. We use the models as communication channels between classes, instances or components, but we shouldn't, the models can be triggers of changes but all other components shouldn't know about them, only Cartodb.js

### **Why remove backbone?**

As I said before, Cartodb.js should be a master component, a component orchestrating the behaviour of others components, it doesn't need to render something or handle dom events at all, and the models can be not models but stores or instances states. So we can leave the inheritance behaviour *inherited* from Backbone.js

## Rethinking Carto

**How I see Cartodb.js?**

**Cartodb.js is a component orchestrator, is the ring to government all**

Let's see from an API point of view, all of this are examples.

Map:
`.addLayer`
`.removeLayer`
`.editLayer` -> `'HIDE', 'ID'` or `'SET_TITLE', 'Drug dealers locations'`
`.getLayers`

Legends:
`.addLegend`
`.removeLegend`
`.editLegend`

Events:
But... we have to listen to events.Yeah, we can expose a set of predefined events.

`.on` -> `zoom:changed`, `legends:toggled`, `legends:renamed`, `map:bounds`, `feature:selected`, `feature:hover` *We have to discuss the naming*

It's hard to explain all of it in here, but I can extend it as far as you wish. Let me know if you have some question.

What do you think guys? Specially @alonsogarciapablo.

cc/ @javisantana & @rochoa

javisantana:
I think you are missing a lot of context so let's organize a meeting so I can answer some of your "why's".

donflopez:
Close, we opened #1594

Issue #1517, opened 2016-11-29
Improve performance of expandable geometries feature


alonsogarciapablo:
@donflopez suggested a performance improvement in 
[this comment](https://github.com/CartoDB/cartodb.js/pull/1507#pullrequestreview-10335080). Basically, we don't need to repaint all middle points while dragging a vertex. Instead we can "move" (update their lat and long) of those who are related to the node that is being dragged.

alonsogarciapablo:
cc: @rochoa

PR #1584, opened 2017-02-20
Adds support for Google Maps basemaps

alonsogarciapablo:


src/geo/gmaps/gmaps-default-map-style.js
```diff
@@ -1,69 +0,0 @@
-var DEFAULT_MAP_STYLE = [{
```
xavijam: Why have you removed this? There will not be an option to create a google map instead a leaflet one?

alonsogarciapablo: All of the Google Maps basemaps available in BUILDER have some styles, so there was no need to provide a default map style.

xavijam: But I feel we should take into account when the user wants to create a map from the beginning using just the CARTO.js library, shouldn't we?

alonsogarciapablo: That's something we'll have to think we work on the API. Anyway, if someone wanted to create a map from a custom viz.json for example, they [should be able to specify the `baseType` and `style` attributes](https://github.com/CartoDB/cartodb.js/blob/ff9d12ad6f78035d42dd224cc37c268346e7a159/src/geo/gmaps/gmaps-base-layer-view.js#L31-L40) to customise their maps. And, [they will get the default style if they use any of the default map types](https://github.com/CartoDB/cartodb.js/blob/ff9d12ad6f78035d42dd224cc37c268346e7a159/src/geo/gmaps/gmaps-base-layer-view.js#L6-L11).

xavijam:
Hey, just found a small performance problem, in case we can fix it easily. Editing geometries with Google Maps as provider is not as smooth as in Leaflet. Here is an example (although a GIF is not the best example):

flickering-gmaps

Middle points flickers a lot, it is likely due to a performance problem, but it works 👌 .

alonsogarciapablo:
@xavijam True fact. Performance is not very good when editing geometries, and it's indeed worst in the case of Google Maps. There's already an issue about this : https://github.com/CartoDB/cartodb.js/issues/1517. Perhaps we should prioritise it.
cc: @rochoa @donflopez

xavijam:
That's your call guys, for me makes sense to tackle in a different issue, if not this is gonna be endless.

alonsogarciapablo:
Yes, totally.

xavijam:
Bugs fixed! 💃

Issue #1586, opened 2017-02-20
[Google Maps] On hover tooltip doesn't appear for the feature if click popup was opened and closed


xavijam:
Testing new changes with Leaflet provider.

## STR
- Create a map (with Leaflet) of polygons, for example.
- Enable click popups with all attributes selected, for example.
- Check that it is working 👌 .
- Enable on hover popups with all attributes selected, for example.
- Check that it is working 👌 .
- Now, click over one feature, click popup will appear (on hover one should disappear).
- Now close the previous click popup and make over that feature in order to see the on hover tooltip.
- It doesn't appear.

Looks like the "unbind" when the click popup is closed is not working properly.

cc @rochoa @alonsogarciapablo

xavijam:
In production is working properly.

Issue #1585, opened 2017-02-20
[Google Maps] Popup on hover doesn't work properly


xavijam:
## STR
1. Create a map (using GMaps as provider) with a layer of [points/lines/polygons].
2. Go to popup tab.
3. Select HOVER type.
4. Click over the light one.
5. Select all or several attributes.
6. Check the behaviour in the map

- For points: the popup works but it doesn't display any info, always the `No data is available`.
- For lines: same as points.
- For polygons: the popup doesn't show up 😱 .

Issue #1582, opened 2017-02-17
[Google Maps] Infowindow shows null values right after being setup


alonsogarciapablo:
From https://github.com/CartoDB/cartodb.js/issues/1527#issuecomment-280617099:

--
I've found a little issue, when I add a pop-up to a layer for click event, the pop-up shows `null` values: 

![screen shot 2017-02-17 at 11 37 19](https://cloud.githubusercontent.com/assets/1936888/23062374/d04ce060-f505-11e6-9db2-7693032f5ba9.png)

After a hard refreshing, it works as expected: 

![screen shot 2017-02-17 at 11 37 40](https://cloud.githubusercontent.com/assets/1936888/23062407/e9604254-f505-11e6-8d1b-8396978bfed7.png)

Not sure if it is related to basemaps, but I've checked it in prod and pop-up works without refreshing.

Apart from that, everything else works like a charm!!

Hope this helps!

--

PR #1587, opened 2017-02-21
Fixed several bugs related to infowindows and tooltips

alonsogarciapablo:
Fixes https://github.com/CartoDB/cartodb.js/issues/1586, https://github.com/CartoDB/cartodb.js/issues/1585, and https://github.com/CartoDB/cartodb.js/issues/1582.

@xavijam 🙏

### DIFF
--- a/src/geo/map/cartodb-layer.js
+++ b/src/geo/map/cartodb-layer.js
@@ -36,6 +36,7 @@ var CartoDBLayer = LayerModelBase.extend({
     this.unset('legends');
 
     this.bind('change', this._onAttributeChanged, this);
+    this.infowindow.fields.bind('reset', this._reloadVis, this);
 
     LayerModelBase.prototype.initialize.apply(this, arguments);
   },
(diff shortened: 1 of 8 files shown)


src/geo/map/cartodb-layer.js
```diff
@@ -36,6 +36,7 @@ var CartoDBLayer = LayerModelBase.extend({
     this.unset('legends');
 
     this.bind('change', this._onAttributeChanged, this);
+    this.infowindow.fields.bind('reset', this._reloadVis, this);
```
xavijam: `reset`, `add` and `remove` or only needed `reset`?

src/vis/vis.js
```diff
@@ -315,8 +315,21 @@ var VisModel = Backbone.Model.extend({
...
+    var successCallback = options.success;
+    var errorCallback = options.error;
+
+    options = _.extend({
+      includeFilters: true,
+      success: function () {
+        this.trigger('reloaded');
+        successCallback && successCallback();
+      }.bind(this),
+      error: function () {
```
xavijam: Don't we want to return the error in the `errorCallback`?

src/geo/map/cartodb-layer.js
alonsogarciapablo: Fields are only reset at the moment, but I'll add `add` and `remove` too 👍

src/vis/vis.js
alonsogarciapablo: We're not currently receiving any error here, so there's no error to "return"...

PR #1592, opened 2017-03-07
Layer view hierarchy

donflopez:


donflopez:
This fix #1593

src/geo/leaflet/leaflet-tiled-layer-view.js
```diff
@@ -2,43 +2,59 @@ var _ = require('underscore');
...
+    self.trigger('loading');
+  });
+
+  this.leafletLayer.onAdd = function (map) {
+    L.TileLayer.prototype.onAdd.apply(this, arguments);
+    self._onAdd();
+  };
+};
+
+LeafletTiledLayerView.prototype = _.extend(
```
rochoa: A silly thing, as an alternative to `_.extend` using an empty object, we could use `_.defaults`. Next.

rochoa:
A nice step forward, I would like to see some more changes similar to the way we inject the native map for the rest of the dependencies, instead of relying on `options` objects. But that's another story.

I would merge it and continue with similar approaches in next iterations 👍 .

alonsogarciapablo:
Thanks for the CR @rochoa!

PR #1594, opened 2017-03-09
Migrate to Leaflet 1.0.3

donflopez:
**Changes**
- Change L.Class for L.Layer
- Flatten array in getCoordinates to retrive paths
- Change wax to have the element and not the view

**Known Issues**
- [ ] Zoom with the buttons doesn't trigger the animation in tangram (use fly to, have a poor performance).
- [x] Click doesn't stop propagation when close popup, opening another one.
- [x] If user makes pan over a feature, it gets opened instead of pan the map.


donflopez:
Fix #1503

donflopez:
Yep @alonsogarciapablo, I've tried with a Surface!

We should also pass some smoke tests to make sure of everything.

donflopez:
We've fixed everything we've seen.

Are we gonna make smoke test of this?

cc/ @javisantana

vendor/wax.cartodb.js
```diff
@@ -3079,9 +3079,9 @@ wax.leaf.interaction = function() {
                         for (var tile in layers[layerId]._tiles) {
                             var _tile = layers[layerId]._tiles[tile];
                             // avoid adding tiles without src, grid url can't be found for them
-                            if(_tile.src) {
-                              var offset = wax.u.offset(_tile);
-                              o.push([offset.top, offset.left, _tile]);
+                            if(_tile.el.src) {
```
xavijam: Hey let's make this change in wax repository too!

alonsogarciapablo: Good point @xavijam! Also, we might want to bump Leaflet in wax.js.

donflopez: I updated wax https://github.com/CartoDB/wax/pull/12 and resumed the access to el property.

package.json
```diff
@@ -33,10 +33,10 @@
     "d3": "3.5.17",
     "tangram.cartodb": "cartodb/tangram.cartodb#master",
     "jquery": "2.1.4",
-    "leaflet": "0.7.3",
+    "leaflet": "1.0.3",
     "mustache": "1.1.0",
     "perfect-scrollbar": "git://github.com/CartoDB/perfect-scrollbar.git#master",
-    "torque.js": "CartoDB/torque#master",
+    "torque.js": "CartoDB/torque#leaflet-migration",
```
rochoa: Does this mean we need to keep two branches in torque.js?

src/geo/geometry-views/leaflet/leaflet-marker-adapter.js
```diff
@@ -41,10 +40,11 @@ LeafletMarkerAdapter.prototype.getIconURL = function () {
...
 
 LeafletMarkerAdapter.prototype.setIconURL = function (iconURL) {
-  var newIcon = L.icon({
-    iconUrl: iconURL
-  });
-  this._nativeMarker.setIcon(newIcon);
+  // Leaflet provides the `setIcon` method which generates a new img for the icon.
+  // We want to be able to update the markers icon while it's being dragged. That's
+  // why we're doing this little hack that changes the src of the existing marker img.
+  this._nativeMarker._icon.src = iconURL;
```
rochoa: What is the problem with generating a new img for the icon?

src/geo/leaflet/leaflet-cartodb-layer-group-view.js
```diff
@@ -101,11 +101,6 @@ LeafletCartoDBLayerGroupView.prototype = _.extend(
           this._onFeatureOver(latlng, containerPoint, waxEvent.data, waxEvent.layer);
           break;
         case 'click':
-        case 'touchend':
```
rochoa: Why don't we handle these events?

package.json
donflopez: No, this is only to have it working in staging... I shouldn't upload that line.

src/geo/leaflet/leaflet-cartodb-layer-group-view.js
donflopez: We want to handle the click event, so it is kind of smell when we are mixing events like pointermove on this, also the benefits of using touch-* or pointer-* is to have multitouch, control pointer pressure, etc... that we are not using.

Here, we only want to track the click event AFAIK, maybe in the past we needed them to browser compatibility or mobile support, but now they are carrying on with problems, for example... the click event is not fired when a touchmove is fired between the touchstart/pointerdown and touchend/pointerend, so rare things start to happens.

The problem comes to life when, apparently, leaflet doesn't stop some events propagations, this makes Leaflet more powerful allowing us to make things with those events but exposes problems like open the infowindow each panning of the map.

Maybe I am missing something, any kind of trick that I am not considering here.

What do you think?

src/geo/geometry-views/leaflet/leaflet-marker-adapter.js
donflopez: We are removing an img node to place another one in his place, the new node doesn't have the mousedown event fired, so it doesn't add the drag, drag start, etc...

Reusing the img node we are avoiding some extra work from the browser and also reusing all the attached events so it works perfectly.

Why is this happening now? It is hard to give up with an answer without make some deep debuging work over Leaflet, they changed the way their events works.

package.json
rochoa: OK, forget about it. I was thinking about the torque version we use in the old editor, but it's an old one 2.11.x, so this should be OK.

src/geo/geometry-views/leaflet/leaflet-marker-adapter.js
rochoa: Ok, fair enough. What about extending the comment in the code with the extra information provided here?

src/geo/leaflet/leaflet-cartodb-layer-group-view.js
rochoa: Don't know much, I wanted to have more details about why we were changing this.

package.json
rochoa: BTW, I guess we want to release a new torque.js version before merging this one, right?

donflopez: Yeah! We've to release a version with this PR https://github.com/CartoDB/torque/pull/284, but not until this it's merged.

src/geo/geometry-views/leaflet/leaflet-marker-adapter.js
donflopez: Sure!

alonsogarciapablo:
I've tested this in staging and seems to be working 👌 . I also don't have any comments about the code. I think we need to merge https://github.com/CartoDB/torque/pull/284 first, right @donflopez?

alonsogarciapablo:
- [ ] zIndexes are not working properly. I will look into it!

donflopez:
Yep, torque goes first.

About the zIndexes, they are not working because of the order attribute of the model.

Issue #1595, opened 2017-03-13
Provide a method to get the URL of a static map image


alonsogarciapablo:
Related to https://github.com/CartoDB/cartodb/issues/10822.

CartoDB.js should provide a way to get the URL of an static image of the map. The Static Maps API provides [this functionality](https://carto.com/docs/carto-engine/maps-api/static-maps-api#zoom--center). We'll basically need to generate something like:

```
.../api/v1/map/static/center/{token}/{z}/{lat}/{lng}/{width}/{height}.{format}
```

The new method should accept the following arguments:

- Center (lat and lng)
- Zoom
- Format (.png or .jpg)
- Size in pixels (width x height).

We can provide some default values to make some of them optional.

Tasks/requirements:

- [x] Base (tiled and plain) and labels layers should be sent to the Maps API so that they can be included in the image.
- [x] When tiled/plain layers change, map should be instantiated again.
- [x] When tiled/plain layers are added/removed from the map, it should be instantiated again.
- [x] Image url must include authentication params (api_key / auth_token)
- [x] Image should not include hidden layers

cc: @javierarce

alonsogarciapablo:
As part of this issue, the Static Maps API should accept a parameter to specify the layers that are included in the image (by index). CartoDB.js needs this so that the URL points to an image that only includes visible layers (cc: @rochoa).

xavijam:
Passed to acceptance, being tested in cartodb/cartodb#10822-export-image branch.

xavijam:
Acceptance passed! 💃 . Moved to deploy, and it will be done when it is deployed [here](https://github.com/CartoDB/cartodb/pull/11789#issuecomment-289392734).

xavijam:
💃

PR #1598, opened 2017-03-16
Implements vis.getStaticImageURL

alonsogarciapablo:
Fixes https://github.com/CartoDB/cartodb.js/issues/1595.

Usage example:

```
vis.getStaticImageURL({
  zoom: 10,
  lat: 180,
  lng: -180,
  width: 400,
  height: 600,
  format: 'jpg'
});
```

Default params are:

```
{
  zoom: 4,
  lat: 0,
  lng: 0,
  width: 300,
  height: 300,
  format: 'png'
}
```

cc: @javierarce


xavijam:
Hey guys, we need this in order to merge https://github.com/CartoDB/cartodb/issues/11789. Could we help in order to unblock it?

alonsogarciapablo:
cc: @rochoa @donflopez

rochoa:
> Hey guys, we need this in order to merge CartoDB/cartodb#11789. Could we help in order to unblock it?

@alonsogarciapablo and @donflopez are finishing with the leaflet migration, it would be super nice if you can help with the CR.

xavijam:
Sure, I'll do CR and acceptance 💪 . Invading your kanban!

rochoa:
Awesome! 😄 .

src/geo/cartodb-layer-group.js
```diff
@@ -18,24 +19,24 @@ var CartoDBLayerGroup = Backbone.Model.extend({
     this._layersCollection = options.layersCollection;
   },
 
-  contains: function (layerModel) {
-    return this.getIndexOf(layerModel) >= 0;
+  forEachGroupedLayer: function (iteratee, context) {
```
xavijam: Why grouped layer? If the method returns the CartoDB layers...

src/geo/map/layer-types.js
```diff
@@ -0,0 +1,36 @@
+var TILED_LAYER_TYPE = 'Tiled';
```
xavijam: I feel we should make use of this in Builder instead of having another similar to this there.

xavijam: cc @CartoDB/frontend

src/geo/map/plain-layer.js
```diff
@@ -11,6 +14,28 @@ var PlainLayer = LayerModelBase.extend({
...
+  },
+
+  initialize: function (attrs, options) {
+    attrs = attrs || {};
+    options = options || {};
+    if (!options.vis) throw new Error('vis is required');
+
+    this._vis = options.vis;
+
+    this.bind('change', this._onAttributeChanged, this);
```
xavijam: We are using `listenTo` in the new Builder code, and I feel it would be great if we make use of the same convention here (only for Backbone objects). How does it sound @alonsogarciapablo @donflopez?

xavijam:
Also, I'm trying to update Deep-insights with those last changes (in order to make a staging deploy for Builder), and the PR is failing, seems like CartoDB.js require is not working for any test. Any clue?

xavijam:
Forget my last comment, updating v4 in this branch and then in deep-insights, all 👌 .

src/geo/cartodb-layer-group.js
alonsogarciapablo: CartoDB layers are being grouped by this CartoDBLayerGroup, so I figured "grouped layers" would be a good name for those.

alonsogarciapablo:
Thanks for the CR @xavijam 💋 

You'll find more context on the issue (https://github.com/CartoDB/cartodb.js/issues/1595). Some of the changes you're seeing were required because we now need to also send the base and labels layers to Maps API so that they can also be included in the static image.

src/geo/map/plain-layer.js
alonsogarciapablo: In this project, we use `on` and `bind` and `listenTo`... I don't have a strong preference for any of them.

src/geo/cartodb-layer-group.js
xavijam: :+1:

src/geo/map/plain-layer.js
xavijam: take into account that, for views, if you have a custom model (not this.model), you need to keep using `.add_related_model(whatevermodel)`. But with `listenTo`, you don't need that.

alonsogarciapablo: That's interesting. But we still need to explicitly do `view.stopListening(model);` (http://backbonejs.org/#Events-stopListening), don't we?

xavijam:
Hey @alonsogarciapablo, I've tested:

- Create a layer and add it.
- Move layers in the layers-list.
- Add/remove interactivity.
- Tested image export functionality (with Leaflet provider).
- Embed map.

Any other thing you miss out?

alonsogarciapablo:
Thanks @xavijam! Some other stuff that I'd check to ensure the URL for the static image is right:

- Showing / hiding some layers (check the `layers=X,Y` param in the static image URL)
- Changing basemap (eg: with and without labels, plain color). Make sure the resulting image is correct.
- Torque layers: Check that they are part of the image.

xavijam:
Ok!, let me test them 💋 .

xavijam:
> Showing / hiding some layers (check the layers=X,Y param in the static image URL)
  It requests the proper url, but the image is not the correct, I guess it is related with the version of Windshaft in staging.

alonsogarciapablo:
Eso https://github.com/CartoDB/cartodb/pull/11789#issuecomment-289783703

xavijam:
Great, thanks man. Also, about torque layers. Would it be possible to export the image in the moment we want or it will take always the step 0?

alonsogarciapablo:
> Would it be possible to export the image in the moment we want

I don't think the [static maps api](https://carto.com/docs/carto-engine/maps-api/static-maps-api) supports that.

xavijam:
Yeah, let0s see if in the future it is possible. For me, ready to be merged!.

PR #1602, opened 2017-03-22
Tile urls subdomain

donflopez:


src/geo/cartodb-layer-group-view-base.js
```diff
@@ -65,7 +65,7 @@ CartoDBLayerGroupViewBase.prototype = {
         tilejson: '2.0.0',
         scheme: 'xyz',
         grids: this.model.getGridURLTemplates(layerIndexInLayerGroup),
-        tiles: this.model.getTileURLTemplates(),
+        tiles: [this.model.getTileURLTemplate()],
```
alonsogarciapablo: I don't think this is used by wax, so we could try to remove it.

donflopez: It is used to know how to get the tile coords from the image src

src/geo/cartodb-layer-group.js
```diff
@@ -91,7 +98,7 @@ var CartoDBLayerGroup = Backbone.Model.extend({
   },
 
   hasTileURLTemplates: function () {
-    return this.getTileURLTemplates().length > 0;
+    return this.getTileURLTemplate() ? true : false;
```
alonsogarciapablo: I guess we could do this here: `return !!this.getTileURLTemplate()`

donflopez: You're right!

src/geo/cartodb-layer-group-view-base.js
alonsogarciapablo: This method is only using other methods from `this.model`, so it could be moved to `src/geo/cartodb-layer-group.js`. I think it'd simplify things a bit... don't you think @donflopez?

alonsogarciapablo: We could then just do somethink like `this.model.toTileJSON()` in this class.

src/windshaft-integration/model-updater.js
```diff
@@ -40,7 +40,8 @@ ModelUpdater.prototype.updateModels = function (windshaftMap, sourceId, forceFet
 
 ModelUpdater.prototype._updateLayerGroupModel = function (windshaftMap) {
   var urls = {
-    tiles: this._calculateTileURLTemplatesForCartoDBLayers(windshaftMap),
+    tiles: this._generateTileURLTemplate(windshaftMap),
+    subdomains: this._getTileSubdomains(windshaftMap),
```
alonsogarciapablo: `subdomains` is not a url so perhaps it makes sense to set it as a separate attribute IMO.

src/geo/cartodb-layer-group-view-base.js
alonsogarciapablo: Ok. From what I understand `this.model.getTileURLTemplate()` will return the template with the `{s}` placeholder, right? Wouldn't we need to generate different URLs for each subdomain?

donflopez: Yes, that's another issue, thanks!!

src/geo/gmaps/gmaps-cartodb-layer-group-view.js
```diff
@@ -235,7 +235,7 @@ _.extend(
     _reload: function () {
       var tileURLTemplates;
       if (this.model.hasTileURLTemplates()) {
-        tileURLTemplates = this.model.getTileURLTemplates();
+        tileURLTemplates = [this.model.getTileURLTemplate()];
```
alonsogarciapablo: Have we tested this? We're not providing subdomains to the Google Maps layer so it probably doesn't know how to handle `{s}` properly... We'd probably need to replace `{s}` here.

src/geo/gmaps/gmaps-torque-layer-view.js
```diff
@@ -42,7 +42,7 @@ _.extend(
 
       if ('tileURLTemplates' in changed) {
         // REAL HACK
-        this.provider.templateUrl = this.model.getTileURLTemplates()[0];
+        this.provider.templateUrl = this.model.getTileURLTemplate();
```
alonsogarciapablo: As I mentioned in https://github.com/CartoDB/cartodb.js/pull/1602/files#r107705091, I'm not sure Google Maps will know how to handle `{s}` placeholders...

donflopez: You're right, he doesn't. I'll fix that

src/geo/cartodb-layer-group.js
```diff
@@ -46,17 +46,34 @@ var CartoDBLayerGroup = Backbone.Model.extend({
     return !!this.get('urls');
   },
 
-  getTileURLTemplates: function (type) {
+  getSubdomains: function () {
+    return this.get('urls') && this.get('urls').subdomains;
+  },
+
+  getTileURLGrid: function () {
```
alonsogarciapablo: This method is basically returning an array of urlTemplates with the `{s}` replaced by the subdomain... right? I don't quite understand the name... What about merging this method with getTileURLTemplate so we can do something like:

```
> model.getTileURLTemplates();
[
  "http://{s}.ashbu.cartocdn.com/documentation/api/v1/map/010b3e5e4b0e9997024b739bb538260a:0/0/{z}/{x}/{y}.png
]
> model.getTileURLTemplates({ replaceSubomains: true });
[
  "http://1.ashbu.cartocdn.com/documentation/api/v1/map/010b3e5e4b0e9997024b739bb538260a:0/0/{z}/{x}/{y}.png",
  "http://2.ashbu.cartocdn.com/documentation/api/v1/map/010b3e5e4b0e9997024b739bb538260a:0/0/{z}/{x}/{y}.png",
  "http://3.ashbu.cartocdn.com/documentation/api/v1/map/010b3e5e4b0e9997024b739bb538260a:0/0/{z}/{x}/{y}.png",
  "http://4.ashbu.cartocdn.com/documentation/api/v1/map/010b3e5e4b0e9997024b739bb538260a:0/0/{z}/{x}/{y}.png"
]

```

src/geo/gmaps/gmaps-cartodb-layer-group-view.js
alonsogarciapablo: We could use something like what I suggested in https://github.com/CartoDB/cartodb.js/pull/1602/files#r107707428 here.

src/geo/gmaps/gmaps-torque-layer-view.js
alonsogarciapablo: Ping ☝️

alonsogarciapablo:
Ok, so this is what we have so far in terms of the cartodb-layer-group.js:

TILES:

```

> layerGroupModel.getTileURLGrid(); // used for the TileJSON that is generated in cartodb-layer-group-view-base.js

[
  "http://0.ashbu.cartocdn.com/documentation/api/v1/map/010b3e5e4b0e9997024b739bb538260a:0/0/{z}/{x}/{y}.png",
  "http://1.ashbu.cartocdn.com/documentation/api/v1/map/010b3e5e4b0e9997024b739bb538260a:0/0/{z}/{x}/{y}.png",
  "http://2.ashbu.cartocdn.com/documentation/api/v1/map/010b3e5e4b0e9997024b739bb538260a:0/0/{z}/{x}/{y}.png",
  "http://3.ashbu.cartocdn.com/documentation/api/v1/map/010b3e5e4b0e9997024b739bb538260a:0/0/{z}/{x}/{y}.png"
]

> vis.layerGroupModel.getTileURLTemplate(); // Used by leaflet-cartodb-layer-group-view.js

"http://{s}.ashbu.cartocdn.com/documentation/api/v1/map/010b3e5e4b0e9997024b739bb538260a:0/0/{z}/{x}/{y}.png"


// We are also doing the following in gmaps-cartodb-layer-group-view.js:

this.model.getTileURLTemplate().replace('{s}', this.model.getSubdomains()[0]);

```

GRIDS:

```
> vis.layerGroupModel.getGridURLTemplates(0); // used for the TileJSON that is generated in cartodb-layer-group-view-base.js
[
  "http://0.ashbu.cartocdn.com/documentation/api/v1/map/010b3e5e4b0e9997024b739bb538260a:0/0/{z}/{x}/{y}.grid.json",
  "http://1.ashbu.cartocdn.com/documentation/api/v1/map/010b3e5e4b0e9997024b739bb538260a:0/0/{z}/{x}/{y}.grid.json",
  "http://2.ashbu.cartocdn.com/documentation/api/v1/map/010b3e5e4b0e9997024b739bb538260a:0/0/{z}/{x}/{y}.grid.json",
  "http://3.ashbu.cartocdn.com/documentation/api/v1/map/010b3e5e4b0e9997024b739bb538260a:0/0/{z}/{x}/{y}.grid.json"
]

```

Also, the model exposes a `getSubdomains` which basically returns the subdomains that can be used to fetch tiles and grids for the layerGroup.

alonsogarciapablo:
I fixed the tests and I found a couple of issues:

- [ ] URLs for attributes are not being generated correctly. They now include `{s}` (eg: `http://%7Bs%7D.ashbu.cartocdn.com/documentation/api/v1/map/b28fb73289e525bac025ee8a39729dad:0/1/attributes/3?callback=_cdbi_layer_attributes_0_2` and are not working.

- [ ] Torque layers are [using some default subdomains](https://github.com/CartoDB/torque/blob/086da527bde7aa1c53df351f951f7942da829c29/lib/torque/provider/windshaft.js#L287-L295), an not the ones that Maps API is returning. We should set them correctly when the TorqueLayerView is initialised.

alonsogarciapablo:
Ok, I renamed methods a bit and now we have:

`layerGroupModel.getTileURLTemplatesWithSubdomains` and `layerGroupModel.getTileURLTemplate()` for tiles.

and

`layerGroupModel.getGridURLTemplatesWithSubdomains(layerIndex)` for grids.

These names are a bit long, but I think they are more explicit and easier to understand. What do you think @donflopez?

donflopez:
I am not a long names friend, I'm gonna lose my fingers programming ;P, what do you think about:

- `layerGroupModel.getTileURLTemplate` and `layerGroupModel.getSubdomainsURLs`
- `layerGroupModel.getGridURLs()`

src/geo/leaflet/leaflet-torque-layer-view.js
```diff
@@ -45,6 +45,7 @@ LeafletTorqueLayer.prototype = _.extend(
 
       if (model.hasChanged('tileURLTemplates')) {
         this.leafletLayer.provider.templateUrl = model.getTileURLTemplates()[0];
+        this.leafletLayer.provider.options.subdomains = model.get('subdomains');
```
alonsogarciapablo: We'd need to something similar for Google Maps + Torque.

test/unit/windshaft-integration/model-updater.spec.js
```diff
@@ -59,11 +59,9 @@ describe('src/vis/model-updater', function () {
     spyOn(this.modelUpdater, '_getProtocol').and.returnValue('http');
   });
 
-  describe('.updateModels', function () {
+  fdescribe('.updateModels', function () {
```
donflopez: ⬆️ fdescribe!

alonsogarciapablo: Good catch! Thanks!

Issue #1606, opened 2017-03-29
getStaticImageURL method is not providing the valid URL with the subdomain


xavijam:
We have checked in production the behaviour of the image-export doesn't work because we are not replacing in the cdn url the subdomain, so image url that is provided is, for example:

`https://ourcdn_%7Bs%7D.morefromourcdn/xavijam/api/v1/map/static/center/xxxxxxxxxxxxx:xxxxxxxxxx/3/37.57941251343841/-35.68359375000001/300/200.png?layer=0,1,2,3&api_key=heymiapikey`

xavijam:
@alonsogarciapablo @donflopez

alonsogarciapablo:
:(

jorgesancha:
assigned to @alonsogarciapablo per @rochoa's request

alonsogarciapablo:
I'm fixing it already 👍

alonsogarciapablo:
## Acceptance

- Checkout the code in this branch
- Run `grunt dev` and visit: `localhost:9001/examples/v4/leaflet-layergroup.html`
- Open the console in your browser
- Type: `window.open(vis.getStaticImageURL())`
- You should see a static image of the map (and the URL should point to a valid subdomain)

xavijam:
Acceptance passed!

xavijam:
MERGED!

PR #1608, opened 2017-03-29
Handle subdomains properly in URLs for static images

alonsogarciapablo:
Fixes https://github.com/CartoDB/cartodb.js/issues/1606

@donflopez can you please take a look? Thanks!

### DIFF
--- a/src/geo/cartodb-layer-group.js
+++ b/src/geo/cartodb-layer-group.js
@@ -48,7 +48,7 @@ var CartoDBLayerGroup = Backbone.Model.extend({
   },
 
   getSubdomains: function () {
-    return (this.get('urls') && this.get('urls').subdomains) || ['0'];
+    return (this.get('urls') && this.get('urls').subdomains) || [];
   },
 
   getTileURLTemplatesWithSubdomains: function () {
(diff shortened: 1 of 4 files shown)


src/geo/cartodb-layer-group.js
```diff
@@ -142,6 +142,7 @@ var CartoDBLayerGroup = Backbone.Model.extend({
     if (staticImageURLTemplate) {
       staticImageURLTemplate = this._appendParamsToURL(staticImageURLTemplate, [ 'layer=' + this._getIndexesOfVisibleLayers() ]);
       staticImageURLTemplate = this._appendAuthParamsToURL(staticImageURLTemplate);
+      staticImageURLTemplate = staticImageURLTemplate.replace('{s}', this.getSubdomains()[0]);
```
donflopez: We were trying to avoid the use of a single subdomain, why add this here? Can't getSubdomains be an empty array?

alonsogarciapablo: Yes. You're right, but... 

1. Since this link is probably going to be used a single time I figured we could just use the first available subdomain.

2. `getSubdomains` can return an empty array, but in that case the template won't have a `{s}` so the replace will have no effect (and it'll work fine with `undefined`).

If you have a better suggestion... I'm all 👂

donflopez: Can that image have '{s}' at any point?

alonsogarciapablo: Yes, we use the CDN template that Maps API returns, which can indeed include a `{s}`. This is actually the problem that we're solving with this PR.

donflopez: Ah, Ok, I didn't know where the .image came from.

PR #1609, opened 2017-03-29
Clean up subdomains

alonsogarciapablo:
This is a follow up on https://github.com/CartoDB/cartodb.js/pull/1607 (I had to revert one cleanup commit which broke some stuff). Since Maps API is now returning info about subdomains that should be used for tiles, grids, and attributes, we should just use those. If this info is not available, no subdomains will be used.

cc: @rochoa @donflopez


donflopez:
Tried in ded02 and it looks like is working like a charm :) 🇪🇸

PR #1612, opened 2017-04-03
Add subdomains to tangram

donflopez:


src/geo/leaflet/leaflet-cartodb-webgl-layer-group-view.js
```diff
@@ -7,7 +7,7 @@ var LeafletCartoDBWebglLayerGroupView = function (layerGroupModel, leafletMap) {
   LeafletLayerView.apply(this, arguments);
 
   layerGroupModel.bind('change:urls',
```
alonsogarciapablo: We can probably "store" this as `this.model` so that we don't need to pass it around.

PR #1614, opened 2017-04-04
Exposed some method and info through map model.

nobuti:
This PR exposes some internal through map model in order to avoid consuming them globally. Related to https://github.com/CartoDB/cartodb/issues/11846

### DIFF
--- a/src/geo/gmaps/gmaps-map-view.js
+++ b/src/geo/gmaps/gmaps-map-view.js
@@ -10,7 +10,6 @@ var GoogleMapsMapView = MapView.extend({
     this._isReady = false;
 
     MapView.prototype.initialize.apply(this, arguments);
-    this._linkMapHelpers();
   },
 
   _createNativeMap: function () {
--- a/src/geo/gmaps/gmaps-map-view.js
+++ b/src/geo/gmaps/gmaps-map-view.js
@@ -3,6 +3,7 @@ var _ = require('underscore');
 var MapView = require('../map-view');
 var Projector = require('./projector');
 var GMapsLayerViewFactory = require('./gmaps-layer-view-factory');
+var Utils = require('../../core/util');
 
 var GoogleMapsMapView = MapView.extend({
   initialize: function () {
@@ -195,6 +196,10 @@ var GoogleMapsMapView = MapView.extend({
       lng: latlng.lng()
     };
   }
+});
+
+_.extend(Utils, {
+  pixelToLatLng: GoogleMapsMapView.prototype.containerPointToLatLng
 });
 
 module.exports = GoogleMapsMapView;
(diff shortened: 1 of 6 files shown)


src/geo/map-view.js
```diff
@@ -121,6 +121,13 @@ var MapView = View.extend({
     return this;
   },
 
+  _linkMapHelpers: function () {
+    this.map.set({
```
alonsogarciapablo: Why set these as attributes? Here's an idea:

```
this.map.setPixelToLatLngConverter(this.containerPointToLatLng.bind(this));
this.map.setLatLngToPixelConverter(this.latLngToContainerPoint.bind(this));
```

```
// in map.js
setPixelToLatLngConverter: function (pixelToLatLngConverter) {
  this._pixelToLatLngConverter = pixelToLatLngConverter;
},

setLatLngToPixelConverter: function (latLngToPixelConverter) {
  this._latLngToPixelConverter = latLngToPixelConverter;
},

pixelToLatLng: function (point) {
  return this._pixelToLatLngConverter && this._pixelToLatLngConverter(point);
},

latLngToPixel: function (latLng) {
  return this._latLngToPixelConverter && this._latLngToPixelConverter(point);
}

```

This way, those methods will be exposed directly. They could raise an exception perhaps if "converters" haven't been set... Just an idea...

src/geo/leaflet/leaflet-map-view.js
```diff
@@ -5,6 +5,10 @@ var MapView = require('../map-view');
 var LeafletLayerViewFactory = require('./leaflet-layer-view-factory');
 
 var LeafletMapView = MapView.extend({
+  initialize: function () {
+    MapView.prototype.initialize.apply(this, arguments);
+    this._linkMapHelpers();
```
alonsogarciapablo: Since both subclasses (`leaflet-map-view` and `gmaps-map-view.js`) are invoking this ☝️, we can just move it up to the `initialize` method of the parent class (`map-view.js`)

src/geo/gmaps/gmaps-map-view.js
```diff
@@ -65,6 +66,10 @@ var GoogleMapsMapView = MapView.extend({
       self.trigger('dblclick', e);
     });
 
+    google.maps.event.addListener(this._gmapsMap, 'bounds_changed', function (e) {
```
xavijam: What about adding a test for google and leaflet when "bounds" has changed and the function is triggered?

src/geo/map-view.js
```diff
@@ -117,10 +119,15 @@ var MapView = View.extend({
 
     // Enable geometry management
     this._geometryManagementController.start();
-
+    this.map.setMapViewSize(this.getSize());
     return this;
   },
 
+  _linkMapHelpers: function () {
```
xavijam: :+1:, worth it a test.

src/geo/map.js
```diff
@@ -477,7 +477,33 @@ var Map = Model.extend({
     }
 
     return zoom - 1;
+  },
+
+  setPixelToLatLngConverter: function (pixelToLatLngConverter) {
```
xavijam: What about if we check if those functions are available with a test?

xavijam:
Merging! 👏

PR #1615, opened 2017-04-05
It is not necessary to reCenter the map after initialization

xavijam:
Related ticket: https://github.com/CartoDB/support/issues/605

Basically we have noticed that Leaflet 1.0.3 doesn't need to set the center of the map when it is invalidated. It means, for example, when widgets list component appears and map gets stretched.
I'll review the code and make the acceptance with @ivanmalagon.

Then we will need to make changes in deep-insights in order to stop using the `centerMapToOrigin` deprecated function.

cc @alonsogarciapablo @donflopez @ivanmalagon

### DIFF
--- a/src/vis/vis.js
+++ b/src/vis/vis.js
@@ -399,11 +399,6 @@ var VisModel = Backbone.Model.extend({
     this.trigger('invalidateSize');
   },
 
-  centerMapToOrigin: function () {
-    this.invalidateSize();
-    this.map.reCenter();
-  },
-
   _flattenLayers: function (vizjsonLayers) {
     return _.chain(vizjsonLayers)
       .map(function (vizjsonLayer) {
--- a/src/geo/leaflet/leaflet-map-view.js
+++ b/src/geo/leaflet/leaflet-map-view.js
@@ -222,14 +222,7 @@ var LeafletMapView = MapView.extend({
   },
 
   invalidateSize: function () {
-    // there is a race condition in leaflet. If size is invalidated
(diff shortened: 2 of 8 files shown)


src/vis/vis.js
```diff
@@ -399,11 +399,6 @@ var VisModel = Backbone.Model.extend({
     this.trigger('invalidateSize');
   },
 
-  centerMapToOrigin: function () {
```
alonsogarciapablo: This method is being used by deep-insights.js ([here](https://github.com/CartoDB/deep-insights.js/blob/7f106bf59c969999fd635789b100b2a311a26d2b/src/api/create-dashboard.js#L111)).

src/vis/vis.js
```diff
@@ -399,11 +399,6 @@ var VisModel = Backbone.Model.extend({
...
-    this.invalidateSize();
-    this.map.reCenter();
```
alonsogarciapablo: Perhaps we can 🔥  `Map#reCenter` too.

src/vis/vis.js
xavijam: Yep, I've already referenced it in the description.

src/vis/vis.js
xavijam: oh, yep!

src/vis/vis.js
alonsogarciapablo: 🔝

alonsogarciapablo:
👍

PR #1616, opened 2017-04-05
[WIP] Dynamic raster vs vector decision

alonsogarciapablo:


src/geo/leaflet/leaflet-layer-view-factory.js
```diff
@@ -5,12 +5,37 @@ var LeafletPlainLayerView = require('./leaflet-plain-layer-view');
 var LeafletCartoDBLayerGroupView = require('./leaflet-cartodb-layer-group-view');
 var LeafletTorqueLayerView = require('./leaflet-torque-layer-view');
 var LeafletCartoDBWebglLayerGroupView = require('./leaflet-cartodb-webgl-layer-group-view');
+// Should we expose a Tangram.cartodb method for doing this?
+var CCSS = require('tangram-cartocss');
```
alonsogarciapablo: Yes, perhaps we could expose a method to check if some given CartoCSS would be supported by tangram-cartocss. 👍

src/geo/leaflet/leaflet-layer-view-factory.js
```diff
@@ -5,12 +5,37 @@ var LeafletPlainLayerView = require('./leaflet-plain-layer-view');
...
 
-var LayerGroupViewConstructor = function (layerGroupModel, mapModel, options) {
-  if (options.vector) {
-    return new LeafletCartoDBWebglLayerGroupView(layerGroupModel, mapModel);
+// Make this configurable and crate a pipe of decision based on the configuration.
+var shouldLoadWithVector = function (metadata, cb) {
```
alonsogarciapablo: IMO, we should try to not "leak" the response from the Maps API into the rest of the app. Instead, I think we should set the relevant info into one of our models (`model-updater` should take care of this). Otherwise we're coupling this code to the response from Maps API. Do you agree @donflopez?

src/geo/map-view.js
```diff
@@ -241,7 +241,7 @@ var MapView = View.extend({
   },
 
   _createLayerView: function (layerModel) {
-    return this._getLayerViewFactory().createLayerView(layerModel, this.getNativeMap());
+    return this._getLayerViewFactory().createLayerView(layerModel, this.getNativeMap(), this._visModel._windshaftMap);
```
alonsogarciapablo: A suggested in a previous comment, we could pass a model with the info (IMO)

src/geo/leaflet/leaflet-layer-view-factory.js
```diff
@@ -5,12 +5,37 @@ var LeafletPlainLayerView = require('./leaflet-plain-layer-view');
 var LeafletCartoDBLayerGroupView = require('./leaflet-cartodb-layer-group-view');
 var LeafletTorqueLayerView = require('./leaflet-torque-layer-view');
 var LeafletCartoDBWebglLayerGroupView = require('./leaflet-cartodb-webgl-layer-group-view');
+// Should we expose a Tangram.cartodb method for doing this?
+var CCSS = require('tangram-cartocss');
 
-var LayerGroupViewConstructor = function (layerGroupModel, mapModel, options) {
-  if (options.vector) {
-    return new LeafletCartoDBWebglLayerGroupView(layerGroupModel, mapModel);
+// Make this configurable and crate a pipe of decision based on the configuration.
```
alonsogarciapablo: How would you do this ☝️ ?

donflopez: I'd do something that expects a configuration and generates the decision chain.

We have to know the data.

```javascript
let someConfig = { // this configuration could come from the server
  ...
  rules: [{
    pos: 1,
    body: `return meta.data.stats.total > 10e6;` // It have to return a boolean (true -> raster, false -> vector)
  },
  ...
  ]
  ```
With this we can configure this filters from a dashboard... or based on other circunstancies like for example the next one:

```javascript
let someConfig = { // this configuration could come from the server
  ...
  rules: [{
    pos: 1,
    body: `meta.stats.last.time > 10000 && meta.stats.last.total > meta.stats.total)
      // time spent for opening the last map in this computer with vector is greather than 10s
      // and the number of vertex is higher than the last one, open with raster`
      
  },
  ...
  ]
  ```

File: src/geo/leaflet/leaflet-layer-view-factory.js (hunk 2 above)
Comment (donflopez) [inline src/geo/leaflet/leaflet-layer-view-factory.js]: Yes, I agree! I've taken a fast way for prototyping purposes.

File: src/geo/map-view.js (hunk 3 above)
Comment (donflopez) [inline src/geo/map-view.js]: Yes, I agree! I've taken a fast way for prototyping purposes.

**Comment by donflopez** [comment]
Fix https://github.com/CartoDB/cartodb/issues/11783

File: src/geo/leaflet/leaflet-layer-view-factory.js [hunk 5]
```diff
@@ -5,12 +5,37 @@ var LeafletPlainLayerView = require('./leaflet-plain-layer-view');
...
+  else {
+    try {
+      metadata.layers.forEach(function (layer, i) {
+        if (layer.type === 'mapnik') {
+          CCSS.carto2Draw(layer.meta.cartocss, i);
+        }
+      });
+    }
+    catch (e) {
+      // TODO: show why it is not working and give feedback to the user
```
javisantana: I think this could be a good source of information to know what the properties we should implement are

src/geo/leaflet/leaflet-layer-view-factory.js
```diff
@@ -5,12 +5,37 @@ var LeafletPlainLayerView = require('./leaflet-plain-layer-view');
...
 var LeafletCartoDBWebglLayerGroupView = require('./leaflet-cartodb-webgl-layer-group-view');
+// Should we expose a Tangram.cartodb method for doing this?
+var CCSS = require('tangram-cartocss');
 
-var LayerGroupViewConstructor = function (layerGroupModel, mapModel, options) {
-  if (options.vector) {
-    return new LeafletCartoDBWebglLayerGroupView(layerGroupModel, mapModel);
+// Make this configurable and crate a pipe of decision based on the configuration.
+var shouldLoadWithVector = function (metadata, cb) {
+  // NOTE: the limit should be setted by the server?
```
javisantana: no, client knows if it can render or not based on the device and rendering capabilities

src/geo/leaflet/leaflet-layer-view-factory.js
javisantana: What about a regular function like: 
```
wouldThisProviderBeAbleToRender: funcion(meta, deviceMetrics) {
deviceMetrics.supportsWebGL && meta.total_points < 1000
}
```

donflopez: Yes, but can a function be easily parsed? We're talking about the possibility of request that configuration from the server, if not... a regular function is perfect @javisantana

src/geo/leaflet/leaflet-layer-view-factory.js
donflopez: Who and how stipulates that render capabilities?

src/geo/leaflet/leaflet-layer-view-factory.js
javisantana: the "configuration" should be stored client side, right? I think this is too complex and too detailed

src/geo/leaflet/leaflet-layer-view-factory.js
javisantana: the browser tells you if supports webgl or if it's working on a mobile device. We will find out when we start to see maps in production

donflopez: So we are taking this decisions:

- Is mobile -> raster, others -> vector
- have less than 1M points -> vector, other -> raster
- Is cartocss supported -> vector, other -> raster

@javisantana

src/geo/leaflet/leaflet-layer-view-factory.js
donflopez: Perfect!

src/geo/leaflet/leaflet-layer-view-factory.js
alonsogarciapablo: Sounds good to me. We can always change this later.

donflopez: As we talked offline the final filters will be:

- Have less than 100k -> vector, else -> raster
- Is cartocss supported -> vector, else -> raster

cc/ @javisantana & @alonsogarciapablo

javisantana: the rationale is: let's start with a low number of vertices and then raise it if the metrics are fine.

For me the important metrics are:
- render time 
- Time To Map (TTM) which is the time from opening the map to see all the tiles on the screen
 which is a mix or rendering, payload and library size

donflopez: That's perfect!

Should we start to count when the html is parsed? @alonsogarciapablo

alonsogarciapablo:
Closing this for now. Please check out https://github.com/CartoDB/cartodb.js/pull/1627.

PR #1620, opened 2017-04-07
11910 builder issue

xavijam:
Related ticket in Builder: https://github.com/CartoDB/cartodb/issues/11910.

Trying to fix the problem above ^^^, we discovered that action parameters were not working properly when a layer was added or removed. So we have reviewed that and here we go!

CR: @alonsogarciapablo


xavijam:
VAAAAMOS!

PR #1623, opened 2017-04-17
Filter GMapsBase layer before to set styles.

nobuti:
This PR implements https://github.com/CartoDB/cartodb/issues/11838.

We are filtering the google basemap from the layers collection in order to set the json styles. This will work because we are assuming the first layer is always the basemap. See [this comment](https://github.com/CartoDB/support/issues/577#issuecomment-286424774) for more context.

### DIFF
--- a/src/windshaft/named-map.js
+++ b/src/windshaft/named-map.js
@@ -1,4 +1,5 @@
 var MapBase = require('./map-base.js');
+var _ = require('underscore');
 
 var NamedMap = MapBase.extend({
   toJSON: function () {
@@ -13,13 +14,18 @@ var NamedMap = MapBase.extend({
     //       "2": "/** torque visualization */\n\nMap { ... }"
     //     }
     //   }
-    //
-    json.styles = this._layersCollection.reduce(function (p, c, i) {
-      var style = c.get('cartocss');
+
+    var layers = this._layersCollection.filter(function (layer) {
+      return layer.get('type') !== 'GMapsBase';
+    });
+
+    json.styles = _.reduce(layers, function (memo, layer, index) {
+      var style = layer.get('cartocss');
       if (style) {
-        p[i] = style;
+        memo[index] = style;
       }
-      return p;
+
+      return memo;
     }, {});
 
     return json;
--- a/src/windshaft/named-map.js
+++ b/src/windshaft/named-map.js
@@ -1,5 +1,6 @@
 var MapBase = require('./map-base.js');
 var _ = require('underscore');
+var LayerTypes = require('../geo/map/layer-types');
 
 var NamedMap = MapBase.extend({
   toJSON: function () {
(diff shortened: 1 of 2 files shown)


xavijam:
I ❤️  it.

src/windshaft/named-map.js
```diff
@@ -13,13 +14,18 @@ var NamedMap = MapBase.extend({
     //       "2": "/** torque visualization */\n\nMap { ... }"
     //     }
     //   }
-    //
-    json.styles = this._layersCollection.reduce(function (p, c, i) {
-      var style = c.get('cartocss');
+
+    var layers = this._layersCollection.filter(function (layer) {
+      return layer.get('type') !== 'GMapsBase';
```
alonsogarciapablo: @nobuti I added [these little helpers](https://github.com/CartoDB/cartodb.js/blob/v4/src/geo/map/layer-types.js#L25-L27) to avoid typing layer types directly... I'd be 🔝  if we could use it here.

alonsogarciapablo: Another idea (💡): instead of adding a new temporary variable (`layers`), we could filter Google Maps base layers inside the `_.reduce` with something like:

```
json.styles = this._layersCollection.reduce(function (memo, layer, index) {
  if (layer.get('type') === 'GMapsBase') { // we could use the helper I mentioned here
    return memo;
  }

  // append styles to memo
  ...
});
```

Or, now that I think about it a bit more... We can probably just get the [layers of type `CartoDB`](https://github.com/CartoDB/cartodb.js/blob/v4/src/geo/map/layers.js#L53-L55) which are the ones with a `cartocss` attribute (not sure if we should also send styles for `torque` layers)...

xavijam: Torque layers don't have autostyle enabled, for example.

nobuti: I don't think this will work, the point is the index should change when a map has a google basemaps. A style's layer with index 1 must end having an index 0 in the final object. We could create a new indexish variable to take the count, but I feel more complicated.

alonsogarciapablo: @nobuti TRUE! I forgot about the indexes... 😓 

But something like `_.reduce(this._layersCollection.getCartoDBLayers(), function (...) { ... })` should work, right?

nobuti: Mmm I don't think so, `getCartoDBLayers` will return only `CARTODB_LAYER_TYPE` layers. But the thing is, if the map doesn't have a google basemap, it should return the tile layer, otherwise the indexes will change for this case too, and the autostyle will break for embeds not using google basemap.

alonsogarciapablo: Also TRUE! 👍

alonsogarciapablo:
🚀

xavijam:
God bless both of you guys 💃  🚀

PR #1624, opened 2017-04-17
[SMOKES] Dynamic VS Vector decision + Vector Infowindows / Tooltips

alonsogarciapablo:
This PR implements 3 things:

1. VisView (and MapView) is now rendered after the map is first reloaded.
2. CARTO.js now decides whether to use vector or raster. DEPENDS ON https://github.com/CartoDB/Windshaft-cartodb/issues/664.
3. Infowindows & Tooltips for vector rendering.

We'll be running some SMOKE tests for these changes.

Fixes: https://github.com/CartoDB/cartodb/issues/11783


donflopez:
I did the acceptance steps and it works perfectly.

One thing, vector will not work because it needs some changes that are under `dynamic-raster-vs-vector-decision-2` and a new release of `tangram.cartodb`

alonsogarciapablo:
True fact @donflopez! Should we cherry-pick those changes so that we can release this?

donflopez:
Yes, sure!

donflopez:
I already did this! We should try it in staging.

donflopez:
It is working on ded02, if you want to try @alonsogarciapablo, I think we're ready to merge :)

donflopez:
Well, I just tried with a torque map and it fails with the next error:

```
vis-view.js:104 Uncaught TypeError: Cannot read property 'invalidateSize' of undefined
```

donflopez:
Torque is now working, it need this [changes](https://github.com/CartoDB/deep-insights.js/pull/539) on deep-insights.js

donflopez:
I've noticed that the time series widget is not being animated like the map content.

alonsogarciapablo:
retest this please

donflopez:
After this `retest please` I've found that the time series widget moves the time slider but it doesn't move correctly. It moves like 2-3px each step and it should move much more.

donflopez:
Another issue that I found is when clicking to set a range, if you simply click without drag it throw an exception:

```
Uncaught Error: start must be smaller than end
    at AnimatorStepsRange (animator-steps-range.js:6)
    at Animator.steps (animator.js:101)
    at Object.options.ready (torque.js:92)
    at windshaft._setReady (windshaft.js:270)
    at LeafletTorqueLayer.<anonymous> (torque-layer-view-base.js:174)
```

But this could not be related with this changes.

donflopez:
After debugging the tiny steps bug, it looks like it takes 256 steps instead of data_steps, If I create a new map it works.

donflopez:
It works randomly... after try in staging it works always, but I think it is because of the timing of the requests.

If I set a throttle in local it works more often than without.

It looks like a race condition, as we can see [here](https://github.com/CartoDB/cartodb.js/blob/v4/src/vis/vis.js#L240-L242) we don't control the loading flow.

donflopez:
I've found another error when hiding a torque layer:

```
Uncaught TypeError: this.hide is not a function
    at h._onModelChanged (torque-layer-view-base.js:157)
    at s (backbone.js:370)
    at r (backbone.js:356)
    at m (backbone.js:155)
    at N.d.k.trigger (backbone.js:346)
    at N.d.set (backbone.js:526)
    at N.d.update (layer-model-base.js:30)
    at t._onLayerDefinitionChanged (deep-insights-integrations.js:707)
    at s (backbone.js:370)
    at r (backbone.js:356)
```

donflopez:
Adding a new layer after a Torque layer launchs an exception:

```
Uncaught TypeError: Cannot read property 'replace' of undefined
    at wax.cartodb.js:2231
    at Object.e.getGrid (wax.cartodb.js:2309)
    at Object.o.screen_feature (wax.cartodb.js:2642)
    at HTMLDivElement.c (wax.cartodb.js:2494)
    at HTMLDivElement.<anonymous> (wax.cartodb.js:304)
```

donflopez:
I've created [this other branch](https://github.com/CartoDB/cartodb.js/tree/SMOKES-render-view-after-first-reload) to make the smokes deploy

src/geo/map.js
```diff
@@ -19,10 +19,9 @@ var Map = Model.extend({
     drag: true,
     keyboard: true,
     provider: 'leaflet',
-    // enforce client-side rendering using GeoJSON vector tiles
-    vector: false,
     popupsEnabled: true,
-    isFeatureInteractivityEnabled: false
+    isFeatureInteractivityEnabled: false,
+    featureCount: 0
```
alonsogarciapablo: Let's change this to `Infinity` before releasing this so that things continue working as before in case the Maps API doesn't return stats. (cc: @donflopez)

src/geo/leaflet/leaflet-layer-view-factory.js
```diff
@@ -1,24 +1,60 @@
...
+
+var LayerGroupViewConstructor = function (layerGroupModel, nativeMap, mapModel, settingsModel) {
+  if (canMapBeRenderedClientSide(mapModel, settingsModel)) {
+    return new LeafletCartoDBWebglLayerGroupView(layerGroupModel, nativeMap);
+  }
+
+  return new LeafletCartoDBLayerGroupView(layerGroupModel, nativeMap);
+};
+
+var canMapBeRenderedClientSide = function (mapModel, settingsModel) {
```
alonsogarciapablo: @donflopez could you please review latest changes to this logic? I feel like it should be extracted into it's own module / class so we can test it in isolation (it's currently untested)... what do you think? thanks!

src/geo/leaflet/leaflet-layer-view-factory.js
```diff
@@ -1,24 +1,60 @@
...
+  }
+
+  if (mapRenderMode === RenderModes.RASTER) {
+    return false;
   }
-  return new LeafletCartoDBLayerGroupView(layerGroupModel, mapModel);
+
+  // RenderModes.AUTO
+  var featureCount = mapModel.getFeatureCount();
+  return isWebGLSupported() &&
```
alonsogarciapablo: When deciding wether to use Vector or Raster the following conditions must be satisfied in order for CartoDB.js to go for Vector:

- Browser must support WebGL. 
- Map must have stats about the estimated number of features.
- The estimated number of features must be smaller than 10e4 (100000).
- CartoCSS of all CartoDB layers should be supported by tangram-cartocss library.

Sounds good @donflopez, @javisantana?

alonsogarciapablo:
Fixes https://github.com/CartoDB/support/issues/692.

donflopez:
Important reminder!

Some properties set by default are not supported by Tangram and the decision always goes for raster.

We have to think which solution is the best for this cases:

1.- Support those properties.

This can be a pain where most hurt. It is gonna take time and collaboration with Mapzen.

2.- Remove those properties from default ccss.

This has some inconveniences that @makella can explain better than me, some of them are that raster needs some of them to make the visualisations smoother and we have raster in Builder.

3.- Ignore those properties in Tangram.

IMO, this is the best option because we don't affect to the Builder visualisation and we can get rid of those properties that doesn't affect pretty much in visualisations terms in vector.

cc/ @javisantana, @jorgesancha, @rochoa, @makella

makella:
The one property that we have by default that isn't supported by Tangram is the `polygon-gamma` as @donflopez mentioned, this smooths out polygonal data and makes a big difference with things like choropleth maps. 

On the other hand, the only place that we have this default is when polygons are first added. As soon as the user modifies any styling property, `polygon-gamma` and any other custom styling properties are lost so in reality, it isn't a big deal. 

Paco and I did some testing and it seemed that the Tangram rendering did an ok job with the rendering so I'd agree with him that if we can ignore these properties and still get the map to vector rendering that would be preferable instead of something like `polygon-gamma` being something that stops people from getting vector maps.

donflopez:
I'd like to add some other properties that aren't fully supported like `line-comp-op: soft-light` we only support `multiply, add and overlay` at this moment.

makella:
@donflopez oh yes, that's another default cartocss property for polygons :)

alonsogarciapablo:
@makella thanks for your input! ❤️ 

So the idea is to ignore all of the default properties that are not supported by tangram? Should we make a list of these properties are they are just the ones you guys already mentioned (`polygon-gamma` and `line-comp-op: soft-light`)? Not sure if other types of geometries have some unsupported props. Here's the default CartoCSS:

```
#layer['mapnik::geometry_type'=1] {
  marker-width: 7;
  marker-fill: #FFB927;
  marker-fill-opacity: 0.9;
  marker-line-color: #FFF;
  marker-line-width: 1;
  marker-line-opacity: 1;
  marker-placement: point;
  marker-type: ellipse;
  marker-allow-overlap: true;
}
#layer['mapnik::geometry_type'=2] {
  line-color: #3EBCAE;
  line-width: 1.5;
  line-opacity: 1;
}
#layer['mapnik::geometry_type'=3] {
  polygon-fill: #374C70;
  polygon-opacity: 0.9;
  polygon-gamma: 0.5;
  line-color: #FFF;
  line-width: 1;
  line-opacity: 0.5;
  line-comp-op: soft-light;
}
```

Also, not sure if auto-styling might set some props that won't work with Tangram.

@makella you're the boss here! Thanks!

cc: @jorgesancha

javisantana:
If outostyle set a not suported property we should remove it if it's not that important or add support for it.

For example, in case of `line-comp-op: soft-light` we should remove it from autostyle or default style. 

I understand the visual quality is not the same but we should take into account we can't do everything and we need to cut scope at some point

alonsogarciapablo:
We are going to release this internally as is and we can then continue with this discussion and tune things a bit more.

alonsogarciapablo:
Retest this please!

PR #1625, opened 2017-04-18
Fix unnecessary multiple set view

donflopez:
Fix https://github.com/CartoDB/support/issues/636


src/geo/gmaps/gmaps-map-view.js
```diff
@@ -103,13 +103,22 @@ var GoogleMapsMapView = MapView.extend({
   _setScrollWheel: function (model, z) {
     this._gmapsMap.setOptions({ scrollwheel: z });
   },
-
-  _setZoom: function (model, z) {
+  _setView: function (model) {
+    if (model.changed) {
+      if (model.changed.zoom) {
```
alonsogarciapablo: Take a look at http://backbonejs.org/#Model-hasChanged. Perhaps we can use it here.

donflopez: Good point! I'll change it :)

src/geo/map-view.js
```diff
@@ -154,20 +154,18 @@ var MapView = View.extend({
     this._unbindModel();
     this.map.bind('change:view_bounds_sw', this._changeBounds, this);
     this.map.bind('change:view_bounds_ne', this._changeBounds, this);
-    this.map.bind('change:zoom', this._setZoom, this);
```
alonsogarciapablo: Why not have:

```
this.map.bind('change:zoom', this._setZoom, this);
this.map.bind('change:center', this._setCenter, this);
```

?

Perhaps we can even get rid of _setView which is a bit confusing?

src/geo/gmaps/gmaps-map-view.js
```diff
@@ -103,13 +103,20 @@ var GoogleMapsMapView = MapView.extend({
   _setScrollWheel: function (model, z) {
     this._gmapsMap.setOptions({ scrollwheel: z });
   },
-
-  _setZoom: function (model, z) {
+  _setView: function (model) {
```
alonsogarciapablo: Is this something we only need to do for Google Maps?

src/geo/map-view.js
donflopez: Because they have the same behaviour applying two times the center and the zoom, we just need one call for both attributes.

src/geo/gmaps/gmaps-map-view.js
donflopez: No, I'm just sharing behaviour here.

donflopez:
cc/ @xavijam

alonsogarciapablo:
@donflopez What's the status of this? I just merged `v4` into this branch.

donflopez:
We stopped using flyTo because of weird behaviours when centring the map, after this fix, it is still broken because of some unnecessary recalls, we must improve this and invest more time on it.

PR #1630, opened 2017-04-26
Vector tooltips infowindow

donflopez:
It depends on #1624 and fix https://github.com/CartoDB/tangram.cartodb/issues/8

### DIFF
--- a/src/geo/cartodb-layer-group.js
+++ b/src/geo/cartodb-layer-group.js
@@ -39,6 +39,12 @@ var CartoDBLayerGroup = Backbone.Model.extend({
     return this._getGroupedLayers()[index];
   },
 
+  getCartoLayerById: function (id) {
+    return this._layersCollection.filter(function (ly) {
+      return ly.id === id;
+    });
+  },
+
   isEqual: function () {
     return false;
   },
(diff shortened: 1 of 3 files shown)


src/geo/leaflet/leaflet-cartodb-webgl-layer-group-view.js
```diff
@@ -20,7 +21,44 @@ LeafletCartoDBWebglLayerGroupView.prototype = _.extend(
...
+        },
+        click: function (e) {
+          if (e.feature) {
+            self.trigger('featureClick', self._getFeatureObject(e));
+          }
+        }
+      });
     },
+    _getFeatureObject: function (e) {
+      var index = 0;
```
alonsogarciapablo: Why is this always `0`?

donflopez: I forgot to change it in the refactor! Good catch :)

src/geo/cartodb-layer-group.js
```diff
@@ -39,6 +39,12 @@ var CartoDBLayerGroup = Backbone.Model.extend({
     return this._getGroupedLayers()[index];
   },
 
+  getCartoLayerById: function (id) {
+    return this._layersCollection.filter(function (ly) {
```
alonsogarciapablo: Perhaps we can use `Collection#get` -> http://backbonejs.org/#Collection-get

alonsogarciapablo:
:+1:

PR #1631, opened 2017-04-27
Optional secondary node

ivanmalagon:
Related to https://github.com/CartoDB/cartodb/issues/12018

In this PR I've made two changes:

- Know if a node parameter of an analysis is optional. In this case, we don't follow the existing workflow. Prior to this SQL ticket there were no optional nodes.

- In `model-updater` we were swallowing error of analysis that didn't get even instantiated. The change check if the error has been set to the layer or analysis. If not, it gets bubbled to a visualization error.  This is related to this [comment](https://github.com/CartoDB/cartodb/issues/12018#issuecomment-297727115)

Please @alonsogarciapablo take a look on this code to validate that it's the right way to tackle this two issues.


src/analysis/analysis-factory.js
```diff
@@ -51,9 +51,14 @@ AnalysisFactory.prototype.analyse = function (analysisDefinition) {
 AnalysisFactory.prototype._getAnalysisAttributesFromAnalysisDefinition = function (analysisDefinition) {
   var analysisType = analysisDefinition.type;
   var sourceNamesForAnalysisType = this._camshaftReference.getSourceNamesForAnalysisType(analysisType);
+  var optionalSourceNamesForAnalysisType = this._camshaftReference.getOptionalSourceNamesForAnalysisType(analysisType);
   var sourceNodes = {};
   _.each(sourceNamesForAnalysisType, function (sourceName) {
-    sourceNodes[sourceName] = this.analyse(analysisDefinition.params[sourceName]);
+    var isOptional = _.indexOf(optionalSourceNamesForAnalysisType, sourceName) > -1;
```
alonsogarciapablo: `_.contains` would work 👌  here too.

src/analysis/analysis-factory.js
```diff
@@ -51,9 +51,14 @@ AnalysisFactory.prototype.analyse = function (analysisDefinition) {
...
+    var sourceParams = analysisDefinition.params[sourceName];
+    if (sourceParams || !isOptional) {
```
alonsogarciapablo: What if `sourceParams` is null / undefined and param is NOT optional?

src/analysis/analysis-factory.js
alonsogarciapablo: 💡💡💡
 
```
this._canshaftReference.isSourceNameOptional(analysisType, sourceName)
```

src/analysis/analysis-factory.js
ivanmalagon: It'll break just as it breaks right now before this PR. If the param is required but it has no value, it should break.

In practice this doesn't happen because we didn't have analysis with optional node as a param (until now).

src/analysis/analysis-factory.js
ivanmalagon: done.

ivanmalagon: :+1:

ivanmalagon:
retest this please

src/windshaft/error.js
```diff
@@ -14,7 +14,7 @@ var WindshaftError = function (error) {
   }
   if (this.isAnalysisError(error.type)) {
     this.context = error.analysis && error.analysis.context;
-    this.analysisId = error.analysis && error.analysis.node_id;
+    this.analysisId = error.analysis && (error.analysis.node_id || error.analysis.id);
```
alonsogarciapablo: I'm not sure if `error.analysis.id` will give us the id of the analysis that has some errors. If I recall correctly, this would be the id of the root node of the analysis. @dgaubert or @rochoa can't comment on this. It'd be nice if Maps API could categorise this error correctly (i.e: setting `error.analysis.node_id`) so we don't need to anything in cartodb.js.

ivanmalagon:
Just beware that this issue should be merged just before https://github.com/CartoDB/cartodb/issues/12018

src/analysis/analysis-model.js
```diff
@@ -88,10 +88,14 @@ module.exports = Model.extend({
...
       return this;
     }
-    var sources = _.chain(this._getSourceNames())
+    var sourceNames = this._getSourceNames();
+    var sources = _.chain(sourceNames)
       .map(function (sourceName) {
         var source = this.get(sourceName);
-        return source.findAnalysisById(analysisId);
+        var isOptional = this._camshaftReference.isSourceNameOptionalForAnalysisType(this.get('type'), sourceName);
+        if (source || !isOptional) {
```
alonsogarciapablo: I would only try to traverse the source if there's indeed a source.
```
if (source) {
  source.findAnalysisById(analysisId)
}
```

src/analysis/analysis-model.js
```diff
@@ -123,10 +127,15 @@ module.exports = Model.extend({
...
-    _.each(this._getSourceNames(), function (sourceName) {
+    var sourceNames = this._getSourceNames();
+    _.each(sourceNames, function (sourceName) {
       var source = {};
-      source[sourceName] = this.get(sourceName).toJSON();
-      _.extend(json.params, source);
+      var isOptional = this._camshaftReference.isSourceNameOptionalForAnalysisType(this.get('type'), sourceName);
+      var sourceInfo = this.get(sourceName);
+      if (sourceInfo || !isOptional) {
+        source[sourceName] = sourceInfo.toJSON();
```
alonsogarciapablo: Same here, I'd only invoke `sourceInfo.toJSON()` if there's a `sourceInfo`. If `sourceInfo` is not defined but it's optional we can just set it to `null` / `undefined` (or whatever the Maps API is waiting for in those cases...):

```
if (sourceInfo) {
  source[sourceName] = sourceInfo.toJSON();
} else {
  if (!isOptional) {
    source[sourceName] = null;  
  }
}
```
```

**Comment by xavijam** [comment]
Shouldn't we merge this one? @ivanmalagon

**Comment by ivanmalagon** [comment]
I'd wait until the acceptance of the mother issue is done. Then, both repos will be merged and deployed.

File: src/windshaft/error.js (hunk 3 above)
Comment (alonsogarciapablo) [inline src/windshaft/error.js]: Talked to @dgaubert and we shouldn't be using `error.analysis.id` cause that's the id of the "root" analysis node.

@ivanmalagon why did we add this in the first place?

Comment (ivanmalagon) [inline src/windshaft/error.js]: Because `error.analysis.node_id` is not populated but `error.analysis.id` is. And it points to the right node in this particular node.

Comment (alonsogarciapablo) [inline src/windshaft/error.js]: @dgaubert do you know why this (:point_up:) might happen? Shouldn't `error.analysis.node_id` always be populated?

Comment (dgaubert) [inline src/windshaft/error.js]: Those identifiers come from the client. For instance in map instantiation:

```
"layers": [ ... ],
"dataviews": { ... },
"analyses": [
    {
        "id": "a1",
        "type": "buffer",
        "params": {
            "source": {
                "id": "a0",
                "type": "source",
                "params": {
                    "query": "SELECT * FROM populated_places_simple"
                }
            },
            "radius": 100000,
            "isolines": 1,
            "dissolved": false
        }
    }
]
```

If something went wrong in `source` node, camshaft tries to get the `id` from `buffer` (`a1`) node as `error.analysis.id` because it is the root node and the `id` from `source` (`a0`) as `error.analysis.node_id` because it is the failing node.

Camshaft does not enforce to have an `id` because it handles its own internal identifiers for each node and windshaft relies on `id`s that needs to replace the queries for layers in layergroup.

Said that, if you have a STR  when `error.analysis.node_id` is not populated, please, file a ticket in camshaft's repo.

alonsogarciapablo: Thanks for the clarification @dgaubert!

@ivanmalagon IMHO, we should not rely on `error.analysis.id` then. Instead, I agree with @dgaubert. Let's try to find an specific case where `error.analysis.node_id` is not being set correctly and fix it elsewhere.

ivanmalagon: Issue opened at camshaft [camshaft/issues/292](https://github.com/CartoDB/camshaft/issues/292)

xavijam:
Hey mates, we need this PR in order to have the SQL function analysis in master. Not sure if the issue in camshaft will be fixed soon. Could we go on with this solution and then get back to the previous state when it is done?

xavijam:
cc @jorgesancha

Issue #1407, opened 2016-09-08
Cleanups


alonsogarciapablo:
Keeping track here of code that is no longer used in the `v4` release:
- Delete `cartodb.js/src/cartodb.js`
- Delete `cartodb.js/src/ui/common/table/`
- Delete `cartodb.js/src/geo/ui/layer-selector.js`

PR #1633, opened 2017-05-03
Removes unused UI components + most of cdb.XXX stuff 🔥 

alonsogarciapablo:
This PR removes (🔥 ) some UI components that live in `src/ui/common` and are not currently used:

- Annotation
- Dialog
- Dropdown
- Infobox
- Image
- Layer View
- Notification
- Table (Row, TableData, TableProperties, RowView)
- TabPane
- Text

We can always "recover" them from CartoDB.js v3.X if we need to.

This PR also removes a lot of stuff from `window.cdb` object (see https://github.com/CartoDB/cartodb.js/pull/1642 for more info).

cc: @donflopez @rochoa 

Fixes https://github.com/CartoDB/cartodb.js/issues/1407.


alonsogarciapablo:
cc: @xavijam

xavijam:
I'd not remove the annotation because we should provide it in Builder like any other overlay. But 👍 .

alonsogarciapablo:
We can always bring back the code to render annotation overlays when we start supporting them in BUILDER.

alonsogarciapablo:
cc: @IagoLast

PR #1636, opened 2017-05-04
Removes usages of global variable `cdb` in layer-model-base

alonsogarciapablo:
Trying to get rid of usages of global variable `cdb` (thare are defined [here](https://github.com/CartoDB/cartodb.js/blob/e355ae1a16db3602a3564f80f2449f8d9dacf1e6/src/cartodb.js)).

This PR removes a couple of usages of `cdb` (`cdb.geo.TorqueLayer` and `cdb.geo.CartoDBNamedMapLayer`) in `src/geo/map/layer-model-base.js` by refactoring  `layerModel#isEqual` method.

cc: @donflopez @rochoa


alonsogarciapablo:
Thanks for the CR @donflopez. It just discovered that `Layer#isEqual` was only being used by `Map#isBaseLayerAdded`, which is no longer used and can be removed. I'll open another PR where I'll be able to 🔥  more code.

Issue #1638, opened 2017-05-05
[CARTO Milestone] Generate a documentation page


donflopez:
We should have a pretty good documentation and it has to be generated easily.

To make this, we are gonna use JSDOC, but we have to use a web page generator from the JSDOC output designing our page using the CARTO style and making it automatically with a bash command.

See [lodash](https://lodash.com/docs/) and [ramda](ramdajs.com/docs/) examples.

PR #1640, opened 2017-05-05
Setup JSDoc

alonsogarciapablo:
- [x] something

This PR introduces a task / script to generate documentation based on JSDoc comments:

```
$ npm run docs
```

This script will remove and re-create the `doc/` directory (using the [jsdoc package](https://www.npmjs.com/package/jsdoc)) with all the docs.

We're using the default template for now, but we can customise it in the future (visit [this link](http://usejsdoc.org/about-configuring-default-template.html) for more info) to integrate in https://carto.com/docs once it's ready.

screen shot 2017-05-05 at 16 44 20

screen shot 2017-05-05 at 16 44 29

The `index.html` includes the context of the `README.md` file. I removed most of the contents of this file because it was pretty outdated. We'll need to revisit it soon. I also added the `@private` tag to most of classes / methods that were documented so that we can start only displaying public API classes / methods in these docs. I only left a couple of `@public` classes for testing / demoing purposes.

This documentation is also available at http://localhost:9001/doc/ when using `grunt dev`.

Part of https://github.com/CartoDB/cartodb.js/issues/1638.

cc: @donflopez @rochoa @javisantana


javisantana:
I do think there should be clear guidlines on how to document stuff because there will be different levels of documentation:

- public methods/classes
- private or non-so-important public methods

For the first ones the quality level should be as high as they could be rendered to a public website right away. And when I mean public website I'm not talking abou the default jsdoc server (which is not valid as documentation), I'm talking about docs.carto.com/carto.js

alonsogarciapablo:
Yes! OpenLayers has solved this by declaring a custom JSDoc tag named @api that will indicate that a class/method/function is part of the API (eg: https://github.com/openlayers/openlayers/blob/master/src/ol/map.js#L87). I will add this a requirement to the issue. Thanks!

PR #1642, opened 2017-05-09
🔥 Clean up `cdb` object and remove unused stuff 🔥

alonsogarciapablo:
There's an internal `cdb` object which is "exported" to window as `window.carto` and `window.cdb` ([here](https://github.com/CartoDB/cartodb.js/blob/cdb-cleanup/src/cartodb.js#L9)). This object used to be a mechanism to namespace things, but it's no longer necessary. We'll have to figure out something similar when we start working on the new API.

Also, this PR removes `Map#isBaseLayerAdded` and `LayerModelBase#isEqual` and `Loader.loadModule`, which are no longer used.

cc: @rochoa


donflopez:
Also, I'd wait to have our main task merged to avoid side effects

alonsogarciapablo:
cc: @xavijam it'd be great if you could confirm / double-check that only the following "namespaced" classes / functions are used in BUILDER:

cdb.core.sanitize
cdb.core.Template
cdb.core.Model
cdb.core.View
cdb.SQL
cdb.createVis

 Thanks!

xavijam:
Checked code in Builder and these are the functions we use:

- cdb.Core.View
- cdb.core.Template
- cdb.core.sanitize
- cdb.core.Model

alonsogarciapablo:
Merging into `remove-old-code` branch.

alonsogarciapablo:
Thanks @xavijam!

Issue #1610, opened 2017-03-31
Using google maps provider, widget dataview first petition could request an invalid bbox


xavijam:
Basically, fixing the bug related to https://github.com/CartoDB/cartodb/issues/11894, I have checked that sometimes widgets are not working properly due to they request the data using the not correct bbox, check image below:

![screen shot 2017-03-31 at 14 12 52](https://cloud.githubusercontent.com/assets/132146/24550035/8b335666-161c-11e7-8f75-c8d24b9b9c07.png)

Seems like Google Map is not ready and it returns `[[0, 0], [0, 0]]` for the current bounding box, and return data for that is empty, that's why the widget appears without anything :(:

```js
…
  _getBoundingBoxFilterParam: function () {
    var result = '';
    var boundingBoxFilter;

    if (this.get('sync_on_bbox_change')) {
      boundingBoxFilter = new WindshaftFiltersBoundingBoxFilter(this._map.getViewBounds());
      result = 'bbox=' + boundingBoxFilter.toString();
    }

    return result;
  },
…
```

It is likely we will need to "wait" until map provider is ready before requesting data from dataview models.

This bug is related to https://github.com/CartoDB/cartodb/issues/11894.

cc @rochoa @jorgesancha

xavijam:
It is a bit random, because sometimes it works. Depends on if GMaps initialization is fast or slow.

alonsogarciapablo:
This is hopefully fixed by https://github.com/CartoDB/cartodb.js/pull/1653. Let's reopen it if we see this problem again.

xavijam:
🤞

PR #1653, opened 2017-05-22
Do not fetch dataviews until view bounds are available if sync_on_bbox_change is true

alonsogarciapablo:
Required for https://github.com/CartoDB/cartodb/issues/11783.

Hopefully fixes https://github.com/CartoDB/cartodb.js/issues/1610.

@xavijam mind taking a look at this 🙏 ? 💋 
cc: @donflopez

### DIFF
--- a/src/dataviews/dataview-model-base.js
+++ b/src/dataviews/dataview-model-base.js
@@ -48,11 +48,15 @@ module.exports = Model.extend({
     var boundingBoxFilter;
 
     if (this.get('sync_on_bbox_change')) {
-      boundingBoxFilter = new WindshaftFiltersBoundingBoxFilter(this._map.getViewBounds());
+      boundingBoxFilter = new WindshaftFiltersBoundingBoxFilter(this._getMapViewBounds());
       result = 'bbox=' + boundingBoxFilter.toString();
     }
 
     return result;
+  },
+
+  _getMapViewBounds: function () {
+    return this._map.getViewBounds();
   },
 
   /**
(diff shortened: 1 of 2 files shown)


src/dataviews/dataview-model-base.js
```diff
@@ -109,9 +113,14 @@ module.exports = Model.extend({
       this.listenTo(layerDataProvider, 'dataChanged', this.fetch);
     } else {
       this.listenToOnce(this, 'change:url', function () {
-        this.fetch({
-          success: this._onChangeBinds.bind(this)
-        });
+        if (this.get('sync_on_bbox_change') && !this._getMapViewBounds()) {
+          // wait until map gets bounds from view
+          this._map.on('change:view_bounds_ne', function () {
```
xavijam: Don't we need to unbind the map listener?

Idea below:

```js
var self = this;
var initialFetch = function () {
  self._map.off('change:view_bounds_ne', this);
};

if (…) {
  this._map.on('change:view_bounds_ne', initialFetch);
}  else {
  initialFetch();
}
```

xavijam:
Fine by me if we make the test pass 💃

src/dataviews/dataview-model-base.js
```diff
@@ -120,6 +129,12 @@ module.exports = Model.extend({
     }
   },
 
+  _initialFetch: function () {
+    this.fetch({
+      success: this._onChangeBinds.bind(this)
```
xavijam: I know it can't fail, but if so...

alonsogarciapablo: We were not handling errors before. I just extracted this to a method so that I was able to reuse it. Having said this, I notice that in the following `fetch`s, we set the state of the dataview and trigger an `error` event if request fails ([here](https://github.com/CartoDB/cartodb.js/blob/df5ec1cd78ef442bb463ee4cea137ae3e6a474f0/src/dataviews/dataview-model-base.js#L302-L312)). We could do that in this initial fetch too...

alonsogarciapablo: Actually, errors will be handled by `DataviewModel#fetch`.

xavijam: :+1:

PR #1657, opened 2017-05-29
Fix gmaps examples

karydja:


Cartofante:

PR #1665, opened 2017-05-31
Time series context menu

ivanmalagon:
Related to https://github.com/CartoDB/cartodb/issues/12177

There was a 'race condition' when changing the number of bins in a timeseries widget. If a change response delays and then we request another number of bins, the datamodel enters a spiral of request the data for those two numbers till infinity.

This fix is independent of the original issue cited but has appeared while developing it.

Related PR:
- cartodb: https://github.com/CartoDB/cartodb/pull/12246
- cartodb.js: https://github.com/CartoDB/cartodb.js/pull/1665
- deep-insights.js: https://github.com/CartoDB/deep-insights.js/pull/547

### DIFF
--- a/src/dataviews/histogram-dataview/histogram-data-model.js
+++ b/src/dataviews/histogram-dataview/histogram-data-model.js
@@ -1,5 +1,5 @@
 var _ = require('underscore');
-var BackboneCancelSync = require('../../util/backbone-abort-sync');
+var BackboneAbortSync = require('../../util/backbone-abort-sync');
 var Model = require('../../core/model');
 
 /**
@@ -33,7 +33,7 @@ module.exports = Model.extend({
   },
 
   initialize: function () {
-    this.sync = BackboneCancelSync.bind(this);
+    this.sync = BackboneAbortSync.bind(this);
     this.bind('change:url change:bins', function () {
       this.fetch();
     }, this);
--- a/src/dataviews/histogram-dataview/histogram-data-model.js
+++ b/src/dataviews/histogram-dataview/histogram-data-model.js
@@ -1,4 +1,5 @@
 var _ = require('underscore');
+var BackboneCancelSync = require('../../util/backbone-abort-sync');
 var Model = require('../../core/model');
 
 /**
@@ -32,6 +33,7 @@ module.exports = Model.extend({
   },
 
   initialize: function () {
+    this.sync = BackboneCancelSync.bind(this);
     this.bind('change:url change:bins', function () {
       this.fetch();
     }, this);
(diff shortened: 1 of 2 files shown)


ivanmalagon:
retest this please

src/dataviews/histogram-dataview/histogram-data-model.js
```diff
@@ -1,4 +1,5 @@
 var _ = require('underscore');
+var BackboneCancelSync = require('../../util/backbone-abort-sync');
```
alonsogarciapablo: Minor thing... we call this `BackboneAbortSync` [here](https://github.com/CartoDB/cartodb.js/blob/v4/src/dataviews/category-dataview/search-model.js#L3)...

src/dataviews/histogram-dataview/histogram-data-model.js
```diff
@@ -32,6 +33,7 @@ module.exports = Model.extend({
   },
 
   initialize: function () {
+    this.sync = BackboneCancelSync.bind(this);
```
alonsogarciapablo: Do you think it'd make sense to have this same behaviour in other dataviews (move this to `DataviewModelBase#initialize`)? I think it could be a good thing to avoid race conditions in other widgets too. What do you think?

src/dataviews/histogram-dataview/histogram-data-model.js
ivanmalagon: and `BackboneCancelSync` in [here](https://github.com/CartoDB/cartodb.js/blob/v4/src/dataviews/dataview-model-base.js#L3)

Which one should we adopt?

src/dataviews/histogram-dataview/histogram-data-model.js
ivanmalagon: I don't know if the same behaviour happens in another data models or if it's something particular on this one.

Let me peek what other data models do.

ivanmalagon: This has been added to a data model, not a dataview model. `Search model` does the same thing, applying the abort sync, so for those two it could make sense to set a `data-model-base`module as in dataviews.

I wouldn't add this abort behaviour blindly to another places if we don't know if they're suffering from the same problem. If ain't broke...

src/dataviews/histogram-dataview/histogram-data-model.js
alonsogarciapablo: I'd go with `BackboneAbortSync` (it matches name of the file) everywhere.

src/dataviews/histogram-dataview/histogram-data-model.js
alonsogarciapablo: I didn't realise this is not the dataview model, but a different model. Plus, [it seems like all DataviewModels are already using `backbone-abort-sync`](https://github.com/CartoDB/cartodb.js/blob/v4/src/dataviews/dataview-model-base.js#L89).

So, let's leave it as is 👍

PR #1676, opened 2017-06-15
Add vector decision log and geometries count

donflopez:
This PR needs this other to work https://github.com/CartoDB/tangram.cartodb/pull/13

### DIFF
--- a/src/geo/leaflet/leaflet-cartodb-webgl-layer-group-view.js
+++ b/src/geo/leaflet/leaflet-cartodb-webgl-layer-group-view.js
@@ -22,7 +22,7 @@ var LeafletCartoDBWebglLayerGroupView = function (layerGroupModel, leafletMap) {
       metric.end();
       metric = void 0;
 
-      log('Rendered Geometries Count: ', self.tangram.getTotalGeometries());
+      log.info('Rendered Geometries Count: ', self.tangram.getTotalGeometries());
     }
   });
 
--- a/src/geo/leaflet/leaflet-cartodb-webgl-layer-group-view.js
+++ b/src/geo/leaflet/leaflet-cartodb-webgl-layer-group-view.js
@@ -1,3 +1,4 @@
+var log = require('cdb.log');
 var _ = require('underscore');
 var L = require('leaflet');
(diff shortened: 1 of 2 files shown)


src/geo/leaflet/leaflet-cartodb-webgl-layer-group-view.js
```diff
@@ -20,6 +21,8 @@ var LeafletCartoDBWebglLayerGroupView = function (layerGroupModel, leafletMap) {
       self.trigger('load');
       metric.end();
       metric = void 0;
+
+      log('Rendered Geometries Count: ', self.tangram.getTotalGeometries());
```
alonsogarciapablo: You need to do something like `log.info`....

src/geo/leaflet/leaflet-layer-view-factory.js
```diff
@@ -14,9 +14,11 @@ var MAX_NUMBER_OF_FEATURES_FOR_WEBGL = 10e4;
 
 var LayerGroupViewConstructor = function (layerGroupModel, nativeMap, mapModel) {
   if (canMapBeRenderedClientSide(mapModel)) {
+    log('VECTOR RENDER MODE');
```
alonsogarciapablo: We should also use `log.info` here and below! Thanks @donflopez!

alonsogarciapablo:
👍

src/geo/leaflet/leaflet-layer-view-factory.js
```diff
@@ -13,30 +13,50 @@ var util = require('../../core/util');
 var MAX_NUMBER_OF_FEATURES_FOR_WEBGL = 10e4;
 
 var LayerGroupViewConstructor = function (layerGroupModel, nativeMap, mapModel) {
-  if (canMapBeRenderedClientSide(mapModel)) {
+  var renderModeResult = getRenderModeResult(mapModel);
+  log.info(renderModeResult);
```
alonsogarciapablo: We could add a prefix here to make it more clear (otherwise it will just log the object in the console):

```
log.info('Render mode:', renderModeResult);
```

PR #1677, opened 2017-06-16
12179 time series selected text

ivanmalagon:
Part of https://github.com/CartoDB/cartodb/issues/12179

In this PR we made two things:
- Add the calculation of the total amount of items and the filtered amount of items in the histogram dataview model..
- Added `stopListening` to the core `View` in case you use `listenTo` instead of `on` (One side effect spotted while doing the task).

### DIFF
--- a/src/dataviews/histogram-dataview-model.js
+++ b/src/dataviews/histogram-dataview-model.js
@@ -149,7 +149,7 @@ module.exports = DataviewModelBase.extend({
     }, 0);
   },
 
-  _calculateFilteredAmount:  function (filter, data) {
+  _calculateFilteredAmount: function (filter, data) {
     var filteredAmount = 0;
     if (filter && filter.get('min') !== void 0 && filter.get('max') !== void 0) {
       var indexes = this._findBinsIndexes(data, filter.get('min'), filter.get('max'));
--- a/src/dataviews/histogram-dataview-model.js
+++ b/src/dataviews/histogram-dataview-model.js
@@ -9,7 +9,9 @@ module.exports = DataviewModelBase.extend({
   defaults: _.extend(
     {
       type: 'histogram',
-      bins: 10
+      bins: 10,
+      totalAmount: 0,
+      filteredAmount: 0
     },
     DataviewModelBase.prototype.defaults
   ),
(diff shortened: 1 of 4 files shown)


test/spec/dataviews/histogram-dataview-model.spec.js
```diff
@@ -340,4 +340,70 @@ describe('dataviews/histogram-dataview-model', function () {
       expect(this.model.get('own_filter')).toBeUndefined();
     });
   });
+
+  describe('._sumBinsFreq', function () {
```
alonsogarciapablo: I'm not a bit fan of testing private methods (unless there's a very good reason to do it, and maybe there is)... In this case, why not just test this by setting different data on the `filter` and asserting that `totalAmount` and `filteredAmount` attributes are updated correctly?

ivanmalagon: I'm a big fan of testing everything as easy as possible and having private methods available eases a lot this process.

In this particular case, a couple of tests in `parse` can check all the paths, so I got rid of those private methods specs and added a couple testing `parse`.

"We may have lost the battle, but not the war"

test/spec/dataviews/histogram-dataview-model.spec.js
```diff
@@ -243,6 +250,47 @@ describe('dataviews/histogram-dataview-model', function () {
     expect(JSON.stringify(parsedData)).toBe('[{"bin":0,"start":55611,"end":70101.25,"freq":2,"max":70151,"min":55611},{"bin":1,"start":70101.25,"end":84591.5,"freq":2,"max":79017,"min":78448},{"bin":2,"start":84591.5,"end":99081.75,"freq":0},{"bin":3,"start":99081.75,"end":113572,"freq":1,"max":113572,"min":113572}]');
   });
 
+  it('should calculate total amount and filtered amount in parse when a filter is present', function () {
```
alonsogarciapablo: 🙇

src/dataviews/histogram-dataview-model.js
```diff
@@ -123,12 +125,56 @@ module.exports = DataviewModelBase.extend({
...
   },
 
+  _onFilterChanged: function (filter) {
+    this.set('filteredAmount', this._calculateFilteredAmount(filter, this._data));
+
+    DataviewModelBase.prototype._onFilterChanged.apply(this, arguments);
+  },
+
+  _calculateTotalAmount: function (buckets) {
+    return _.reduce(buckets, function (memo, bucket) {
```
matallo: reduce FTW

xavijam:
Reviewed and accepted by @alonsogarciapablo and @matallo. I'm in charge of merging it 😬 .

PR #1680, opened 2017-06-21
add layername to dataview model, and getsourceType method

matallo:
re: https://github.com/CartoDB/cartodb/issues/9956

added `layerName` to the dataview model as it will be needed in order to show the layer name in the deep-insights widgets (see attachment below)

screen shot 2017-06-21 at 14 51 55

also added a new `getsourceType` with the same purpose


nobuti:
there are some tests related with mapzen failing, any clue @alonsogarciapablo?

src/dataviews/dataviews-factory.js
```diff
@@ -96,6 +96,7 @@ module.exports = Model.extend({
   _generateAttrsForDataview: function (layerModel, attrs, whitelistedAttrs) {
     attrs = _.pick(attrs, whitelistedAttrs);
     attrs.source = attrs.source || { id: layerModel.id };
+    attrs.layerName = attrs.layerName || layerModel.get('layer_name');
```
alonsogarciapablo: There's a relationship between a DataviewModel and LayerModel...  Why not just add a method to DataviewModel that proxies the request to the layer?


```
// in dataview-model.js...
...
getLayerName: function () {
  return this.layer.get('layer_name');
}
...
```

alonsogarciapablo: Imagine that the name of the layer is changed... `layerName` attribute won't see these changes...

matallo: roger! made the changes and added the specs 👍

src/dataviews/dataview-model-base.js
```diff
@@ -322,6 +322,10 @@ module.exports = Model.extend({
     return sourceAnalysis && sourceAnalysis.get('type');
   },
 
+  getLayerName: function () {
+    return this.layer.get('layer_name');
```
alonsogarciapablo: @matallo it seems that `layer` is not required (we don't enforce setting it in the constructor) so you might want to change this to `return this.layer && this.layer.get('layer_name')`

matallo: awesome, in case there's no layer I have that case covered https://github.com/CartoDB/deep-insights.js/pull/552/files#diff-0c6eb42296959f824ed2b8c1e7927e94R39

alonsogarciapablo: I guess we could also return `''` here  if there's no layer. Not sure if it has any advantage though...

```
if (this.layer) {
  return this.layer.get('layer_name');
}

return '';
```

Issue #1652, opened 2017-05-12
Tap on iOS10 mobile embed jumps to page bottom


xavijam:
@noguerol commented on [Wed May 10 2017](https://github.com/CartoDB/cartodb/issues/12105)

### Context
One tap over an embed on mobile iOS10 jumps the browser (Safari) until the page bottom. If you scroll back to the embed, the normal double tap zoom interaction works as expected. This is not happening with Editor embeds. 

Comes from: https://github.com/CartoDB/support/issues/660

### Steps to Reproduce

1. Open[ this](http://www.michellechandra.com/lorem/) on iOS10 Safari.
2. Tap on the second embed, generated with Builder

### Current Result
Browser jumps to the bottom

### Expected result
Normal zoom double tap

---

@xavijam commented on [Fri May 12 2017](https://github.com/CartoDB/cartodb/issues/12105#issuecomment-301068171)

Moving to its right place.

xavijam:
cc @alonsogarciapablo

alonsogarciapablo:
This might be related: https://github.com/Leaflet/Leaflet/issues/5392 (thanks to @rjimenezda)

rjimenezda:
I've debugged it a bit with the iOS Simulator and the jump happens right after the focus call:

https://github.com/Leaflet/Leaflet/blame/master/src/map/handler/Map.Keyboard.js#L85

The window.scrollTo is a red herring, since it's on embed it doesn't really matter.

I don't see any decent way of disabling that call without modifying native object prototypes, which is a big no-no, or patching Leaflet to disable that line before building, which is also kind of uncool.

rjimenezda:
Maybe `HTMLElement.prototype.focus  = () => {}` isn't such a bad idea, but maybe doing it only for Safari.

rjimenezda:
Same thing happens on Chrome for iOS (which makes sense, afaik iOS browsers are just fancy webviews)

xavijam:
I've been checking the thing and it is true if we set the [`keyboard`](http://leafletjs.com/reference-1.0.3.html#map-keyboard) map attribute to false, the bug is not happening. Why not set that by default and then provide the possibility to enable/disable the keyboard map option?

nobuti:
Deployed

kevin-reilly:
Awesome guys.  Thanks for getting this done.

PR #1682, opened 2017-06-23
Disabling keyboard map navigation in order to fix iOS focus problem

xavijam:
Keyboard handling is provoking https://github.com/CartoDB/cartodb.js/issues/1652 issue. A reasonable fix is disabling that keyboard behavior until [problem is fixed in Leaflet library](https://github.com/Leaflet/Leaflet/issues/5392).

Sounds ok? @CartoDB/frontend @kevin-reilly 

Also, good thing is we could add the keyboard map option in Builder in the near future, because all logic is ready.


alonsogarciapablo:
👍  looks good codewise.

I'm not sure about the implications of disabling the keyboard for everyone though. Perhaps worth providing the Builder option before releasing this change so that maps keep working the way they were working before and users affected by this option can disable keyboard...

xavijam:
@kevin-reilly and @noguerol, I feel this is your call. Adding the option will take us a bit more of time.

nobuti:
how about to "detect" touch devices and disable keyboard only for those?

xavijam:
> how about to "detect" touch devices and disable keyboard only for those?

It could do the trick too, @alonsogarciapablo yay?

xavijam:
Something like:

```js
keyboard: 'ontouchstart' in window || navigator.msMaxTouchPoints, // #cartodb.js/1652
```

alonsogarciapablo:
Sounds like a good idea @nobuti. I would add a new method in `src/core/util.js`...

xavijam:
Ok! thanks guys, doing it!

xavijam:
Now? @nobuti @alonsogarciapablo

alonsogarciapablo:
ok

kevin-reilly:
Thanks guys.  Do you need me to test this at all?

--
*Kevin Reilly*
SVP Product
*CARTO <https://carto.com/> | kevin@carto.com <kevin@carto.com>*
(917) 375-2168

xavijam:
I'll let you know @kevin-reilly, but don't worry 🤗 .

PR #1686, opened 2017-06-27
Cancellable sql fetch.

nobuti:
This PR improves the SQL fetch queries to let them be cancellable. The default option is `false`, as the same as it is now.

```
var sql = new SQL({
    cancellable: true
});
```


nobuti:
@alonsogarciapablo @xavijam could you take a look? 🙏

xavijam:
SpecialK

alonsogarciapablo:
👍  I'm not sure if `cancelable` is better than `cancellable`... We could also go with `abortable`...

PR #1694, opened 2017-07-05
enhanced description about https support to include createLayer, as p…

csobier:
…er a confusion from a user

@alonsogarciapablo , got this request to include this in the docs- as it caused some confusion with one of Jeff's clients.  Okay to merge?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+3.15.12 (05/07/2017)
+-------
+* Same version number, no code changes. Doc enhancement only.
+
 3.15.12 (12/06/2017)
 -------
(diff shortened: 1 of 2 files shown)


NEWS.md
```diff
@@ -1,3 +1,7 @@
+3.15.12 (05/07/2017)
+-------
+* Same version number, no code changes. Doc enhancement only.
```
alonsogarciapablo: No need to change this IMO

csobier: ok, just following the best practices recommended in all the code repos so the build doesn't fail :-)

PR #1685, opened 2017-06-27
C12324 time series aggregations

ivanmalagon:
Related to: https://github.com/CartoDB/cartodb/issues/12324

____________
cartodb PR: https://github.com/CartoDB/cartodb/pull/12347
cartodb.js PR: https://github.com/CartoDB/cartodb.js/pull/1685
deep-insights.js PR: https://github.com/CartoDB/deep-insights.js/pull/555

PR #1695, opened 2017-07-13
time-series: add timezone selector to timeseries histogram

matallo:
re: https://github.com/CartoDB/cartodb/issues/12088

~~this will eventually point to v4, when https://github.com/CartoDB/cartodb.js/pull/1685 is merged~~

---

cartodb PR: https://github.com/CartoDB/cartodb/pull/12464
deep-insights.js PR: https://github.com/CartoDB/deep-insights.js/pull/562
cartodb.js PR: https://github.com/CartoDB/cartodb.js/pull/1695

---

- [x] add offset to filled bins
- [x] tests

~timezone attribute is set from CartoDB, which later is transformed with `moment` in data-model and dataview-model, maybe it will be interesting to change the name of this attribute to `offset` once the transformation is done and it is passed to Windshaft~

~also data-model could have the offset attribute directly, instead of the timezone, and prevent a repeated transformation~

as commented in https://github.com/CartoDB/cartodb.js/commit/a5c405b608647754a0dae9e6638b183094d16f2e some more work needs to be done in Windshaft-cartodb ir order to only pass `timezone` parameter, and `start` and `end` are calculated on the Engine side
some other issues have come up like being able to calculate bins with _weird_ timezones (those which doesn't start o'clock like "Asia/Kabul" +4:30

added the specs to the histogram-dataview-model directly rather than creating a new helpers suite

/cc @ivanmalagon @dgaubert


matallo:
the hack commented in https://github.com/CartoDB/cartodb.js/pull/1695#commitcomment-23073195 is already reverted and Windshaft is working like a charm

src/dataviews/histogram-dataview-model.js
```diff
@@ -165,15 +191,22 @@ module.exports = DataviewModelBase.extend({
...
+  },
+
+  _onOffsetChanged: function () {
     this._reloadVisAndForceFetch();
   },
 
   _calculateTotalAmount: function (buckets) {
     return _.reduce(buckets, function (memo, bucket) {
-      return memo + bucket.freq;
+      return bucket && memo + bucket.freq;
```
ivanmalagon: In case there is a null bucket in the middle of the `buckets` array the result of this reduce is wrong because it'll return null as the whole memo, reseting the result.

See this JSBin with the example https://jsbin.com/godimot/1/edit?html,js,console

src/dataviews/histogram-dataview-model.js
```diff
@@ -298,6 +296,7 @@ module.exports = DataviewModelBase.extend({
 
     if (columnType === 'number' && this.get('bins')) {
       options.bins = this.get('bins');
+<<<<<<< HEAD
```
rjimenezda: 🌝

src/dataviews/histogram-dataview-model.js
```diff
@@ -306,6 +305,10 @@ module.exports = DataviewModelBase.extend({
       if (offset) {
         options.offset = this.get('offset');
       }
+=======
```
rjimenezda: 🌚

ivanmalagon: image

package.json
```diff
@@ -35,6 +35,7 @@
     "jquery": "2.1.4",
     "leaflet": "1.0.3",
     "moment": "^2.18.1",
+    "moment-timezone": "^0.5.13",
```
xavijam: 😱

xavijam: Why aren't we specific with the version we want?

src/dataviews/histogram-dataview/histogram-data-model.js
```diff
@@ -65,6 +67,10 @@ module.exports = Model.extend({
         this.fetch();
       }
     }, this);
+
+    this.on('change:localTimezone', function () {
```
xavijam: ```this.on('change:localTimezone', this.fetch, this);``` ?

src/dataviews/histogram-dataview/histogram-data-model.js
```diff
@@ -120,5 +124,11 @@ module.exports = Model.extend({
     }
 
     return parsedData;
+  },
+
+  _getCurrentOffset: function () {
```
xavijam: Why isn't it "public"?

src/dataviews/histogram-dataview/histogram-data-model.js
matallo: we found problems with too deep call stack (probably because of the prototype method)

package.json
ivanmalagon: I specified the exact same semver that in deep-insights. There we already have the ^. We can modify both repos. As you wish.

src/dataviews/histogram-dataview/histogram-data-model.js
ivanmalagon: Yep. That's a prototype function and we were having bad problems. This way, it works. (But I don't know the real reason)

src/dataviews/histogram-dataview/histogram-data-model.js
ivanmalagon: It's a function only used in this module, to form the URL with the saved offset or the local timezone. It's not needed outta here.

Issue #1703, opened 2017-07-24
JS error is being thrown when infowindow is image type


xavijam:
## STR
- Create a map of points.
- Go to the layer, and add a string column.
- Add some image urls in those fields.
- Go to Popup tab and select the image type.
- Select all fields, but move the column containing image urls to the top.
- Open the browser console.
- Click over a point and check that the console throws an error.

xavijam:
Work done in ```underscore-problem``` branch.

xavijam:
Related ticket -> https://github.com/CartoDB/cartodb.js/issues/1314

rubenmoya:
In production 🚀

xavijam:
@rubenmoya, don't forget to close the ticket!

PR #1704, opened 2017-07-24
Fixed JS error for image infowindows/popup

xavijam:
Related ticket: https://github.com/CartoDB/cartodb.js/issues/1703.

There is a problem with the broserify-shim in Builder, so better to get rid of the problem and just taking a dependency from npm (y santas pascuas). Also, I've been reviewing the process of loading, so it should work propery now, but it is worth to check it. Things touched:

- [x] Infowindow-view spec linting problems.
- [x] Fix problem using a clip-path plugin/lib.
- [x] Fixing problems with loading image process.
- [x] Add new tests for new changes.
- [x] Review behavior in Builder (not only in CARTO.js).

And now a GIF 💃 

image-infowindow

cc @rubenmoya @piensaenpixel


rubenmoya:
This is how it looks now:

screen shot 2017-07-26 at 09 28 57

But in acceptance we found a problem related to the size of the image, if the image height is less than the content height the image hook still appearing:

screen shot 2017-07-26 at 09 31 55

The way we fixed it is removing the image hook in that case:

screen shot 2017-07-26 at 09 29 11

cc @CartoDB/design

arianaescobar:
Rubén, you broke my brain using that image as a test. I was SO confused! haha

rubenmoya:
Sorry! It was the first one I found when I was testing it xD

arianaescobar:
The fix looks good :) 

Although I don't fully get what's the problem with adapting the pop-up to the image size as in the first image.

rubenmoya:
The user is the one who chooses the pop-up width, so if the user selects a small width and because of the aspect ratio of the image, the height doesn't cover the content, we end up with that white space.

arianaescobar:
Ooooh right, got it. Then +1 from me.

piensaenpixel:
cool @rubenmoya :)

PR #1708, opened 2017-07-27
Allow overriding vizJSON with options in create vis

rubenmoya:
Related to: https://github.com/CartoDB/cartodb/issues/12513

We're having problems with the vectors in Builder, in the editor view we override the vizJSON in the rails controller to force raster, but in the datasets view we make a call to the API directly, so it doesn't work.

We were using a `vector: false` option in `deep-insights-integration` which we thought forced the raster mode, but debugging I've realised it does nothing, but I liked that approach so I implemented it to use it in the datasets view, and also, when we split the front/back it will work in the editor view too.

Pinging @rochoa since my knowledge about vector/raster is pretty limited

### DIFF
--- a/src/api/create-vis.js
+++ b/src/api/create-vis.js
@@ -131,6 +131,10 @@ var applyOptionsToVizJSON = function (vizjson, options) {
     vizjson.removeLogoOverlay();
   }
 
+  if (_.has(options, 'vector')) {
+    vizjson.setVector(options.vector);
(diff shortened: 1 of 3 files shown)


rochoa:
I'm not the expert here, but the changes look good to me. 👍


cc'ing @alonsogarciapablo, so he can review this also (although I wouldn't block this waiting for his review).

PR #1712, opened 2017-08-09
Support for mustache conditional statements

rjimenezda:
Sending 'null' instead of `null` makes it hard to use conditional statements if `showEmptyFields` is set to true

Sending `null` literal still renders as 'null' without conditionals, but advanced users are able to parse this in a more granular way.


alonsogarciapablo:
Makes sense 👍

Issue #1717, opened 2017-08-11
Display default error tile when user reaches its limits


xavijam:
When the user reaches its limits, it is likely errors will start coming from the Maps API. At that point, we need to display "error" tiles, and improve the user experience instead of leaving those ugly broken tiles.

## Caveats
- We have two map libraries, Leaflet and GMaps, both of them should support to display a default error tile.
- We should add the error tile image in the CARTO.js bundling (what about retina one? 😛 ).
- We have to decide with @CartoDB/engine team which is the proper 4XX error code returned by the API.

## Design
 
<img width="743" alt="screen shot 2017-08-11 at 11 21 18" src="https://user-images.githubusercontent.com/132146/29207622-3778fb2e-7e87-11e7-93ec-c6375e3d662a.png">

rochoa:
> - We should add the error tile image in the CARTO.js bundling (what about retina one? 😛 ).

One of the discussions we had was about the possibility of using an _artificial_ layer to display the errors themselves. If that is still an option, do we want to use an image or we should aim for something more friendly to the user, like a fake HTML tile where you can link to different docs/resources explaining the problem.

> - We have to decide with @CartoDB/engine team which is the proper 4XX error code returned by the API.

Related to my previous point, we need to check if we want to return a [429 error](https://tools.ietf.org/html/rfc6585#section-4) and we can catch it at the client side, or we have to return an image with a `200 OK`.

xavijam:
1. Well, I'd not add more logic or more layers when error appears. I'd bet for the simplest solution about the error image. Take into account we will display the "new overlay" warning the user about the problem.
2. True, @rubenmoya is on it, but I feel we will be ok with any of the error codes. 

Stay tuned.

rubenmoya:
I've been doing some research about error tiles, both, Leaflet and GMaps have a `onerror` event, which basically is the [img onerror attribute](https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XUL/image#a-onerror), the status code shouldn't be a problem if the response is an invalid image.

Jesus89:
Deployed in ded03 (also deployed Tiler `add-timeout-header` in ded03)

xavijam:
:_)

Jesus89:
Found unexpected behavior. Back to development :arrow_left:

dgaubert:
Let me explain a little the following issue we have found while fetching **raster** tiles (`.png`):

[...]

PR #1721, opened 2017-08-23
Display error tile

rubenmoya:
Related to: #1717

## Description

After a lot of talking and researching we have no way to communicate the front with the maps api, so we decided to show a generic error message instead of making ajax calls to retrieve it.

Now the `mapModel` has an `errors` collection, and both, `leaflet-cartodb-layer-group-view.js` and `gmaps-cartodb-layer-group-view.js` add a `{ type: 'tile' }` error in their respective `error` handlers. This approach is much simpler than making ajax requests.

New public methods added to the map model: `addError(error:obj)`, `getError(type:string)`, `removeError(type:string)`.

When `addError` is called it triggers an `error:[type]` event.

I named the generic error `tile`, but maybe we came up with a better name.

## Examples:

screen shot 2017-09-13 at 10 59 15
screen shot 2017-09-13 at 10 59 31 1
screen shot 2017-09-25 at 17 51 04


src/util/tile-checker.js
```diff
@@ -0,0 +1,27 @@
+var $ = require('jquery');
+var getValue = require('./get-object-value');
+var limitsSVG = require('../../themes/img/timeout.svg');
```
xavijam: For me it is not a timeout svg or limits svg, for the moment is an error image, that's it.

src/util/tile-checker.js
```diff
@@ -0,0 +1,27 @@
...
+
+var buildErrorTile = function (svg) {
+  return 'data:image/svg+xml;base64,' + window.btoa(svg);
```
rubenmoya: IE 10 according to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/btoa), but yep, I don't trust IE, so a manual check would be nice.

src/util/tile-checker.js
rubenmoya: 👍

xavijam:
What about the tests? 💃

xavijam:
@noguerol and @kevin-reilly. This is the first ticket in the frontend side about limits. It is true it is related but not implementing anything specific for that. So, if we merge this, any tile that fails (due to any circunstancy) will display that "error image". Sounds ok?

cc @alonsogarciapablo @rochoa

alonsogarciapablo:
Won't this cause a lot of extra requests (1 for each tile)? And... does that affect the performance?

Also, couldn't the Maps API serve these tiles directly when a limit is reached instead of triggering an error?

xavijam:
We discussed to make only one request (not one by tile) just for checking the root of the problem. If we haven't taken that into consideration, we should fix it @rubenmoya .

rochoa:
I guess the problem is that if the image success, you cannot access their headers, so it won't be possible to show any other information to the user apart from the image itself.

rubenmoya:
I've already changed it (the code is not pushed yet) and it does only one request now, basically the flow is:

- The tile is added as an `img` element, it fails to load and triggers the `onerror` method.
- We set the `src` attribute of the `img` to the error overlay.
- We check `hasError` flag to see if we've done a request already to get the error, if not, we set it to `true` and make the request.

Right now I'm having some problems since the tile fails to load, but when I make the second request it doesn't return an error anymore, so I'll mark the tiles as failed, but I have no idea what error it was. It feels weird that the tile throws an error because of the limits but if I request it again it works. How does the limit works? I think I need to know a little bit more about the whole picture.

rubenmoya:
Yesterday I talked to @rochoa to know better how limits work, and seems like we have 3 cases:

**Case 1**
We request the tile, the response time is within the limits, everything is perfect.

**Case 2**
We request the tile, it times out in the DB, so we don't have the tile. The next request will give us the error and we can work accordingly.

**Case 3**
We request the tile, it times out in the queue, but it returns from the DB and it's cached, so it fails in the front, but the next request to get the error the tile will be there, we end up with the tile using the error overlay, but we don't know the error, since the tile is available now. If this happens in the first (and only) request we make to check the error it means that we will end up with some tiles with the error overlay, but with no feedback about the error.

If we had only the first to cases, it will be all fine, the problem comes with the third case, were we could end up with error tiles but with no feedback about the error.

So, how should we approach this? From the top of my head I can think about 3 options:

- Make a request for every tile, but that means we could make 3 requests, the initial one, the one to check the error, and a third one if the tile is available.

- Make only one request, if the error is available we give more feedback, if it isn't we just show the tiles error overlay.

- Make a collection with the tiles which failed, and start making requests for each one, until any of them give us the error, this is similar to the first option, but we could avoid some requests, the problem is if we have `[1, 2, 3, 4, 5]` maybe we get the error in `3`, but `4` and `5` are cached, which means they are available and we will mark them with the error overlay.

I don't really like any of those options, so I'd love to hear what you think.

alonsogarciapablo:
When you say "the next request" (in case 2 and case 3), we're referring to the request triggered by `TileChecker`, right?

Also, I'm a bit confused by "We request the tile, it times out in the queue, but it returns from the DB and it's cached"... what queue are we talking about? who's caching the tile? Thanks!

rubenmoya:
So, after a few (lot) of changes, this is how it works now:

- In the onError handler it just adds the tile to the `tileErrorCollection`
- the `tileErrorCollection` changes the image `src` for the error overlay
- the tile is added to a queue, where we make one ajax call  for each tile get the error message / check if it was cached
- if the ajax call returns a `200`, the image src it's replaced again, if it returns an error, we add that error to the tile

What I don't know it's if we should go through all the tiles in the queue, since they could be cached and show as many tiles as we can, or if we should stop as soon as we find an error.

src/geo/gmaps/gmaps-cartodb-layer-group-view.js
```diff
@@ -223,13 +223,19 @@ _.extend(
...
-      im.onload = finished;
-      im.onerror = function () {
+
+      tile.onload = finished;
+
+      tile.onerror = function () {
         Profiler.metric('cartodb-js.tile.png.error').inc();
+        self.tileErrorCollection.add({
+          url: _.clone(this.src),
+          node: this
```
xavijam: this is a dom element?

xavijam: DOMel? TileDomNode? (de tu cosecha).

src/geo/gmaps/gmaps-map-view.js
```diff
@@ -67,6 +67,7 @@ var GoogleMapsMapView = MapView.extend({
     });
 
     google.maps.event.addListener(this._gmapsMap, 'bounds_changed', function (e) {
+      self.tileErrorCollection.resetErrorTiles();
```
xavijam: What about not adding this logic within gmaps or leaflet code? Check if it is feasible using it map attributes (center, bounds,...) change.

src/geo/leaflet/leaflet-cartodb-layer-group-view.js
```diff
@@ -53,6 +53,13 @@ var LeafletCartoDBLayerGroupView = function (layerModel, leafletMap) {
   this.leafletLayer.on('loading', function (e) {
     self.trigger('loading');
   });
+
+  this.leafletLayer.on('tileerror', function (layer, coords, error) {
```
xavijam: Same described above, what about "encapsulating" this logic in other place, like the map code (map-view?). As you suggested, you could add more "map libraries" and you would not need to add specific logic for each one.

src/geo/map-view.js
```diff
@@ -30,9 +32,12 @@ var MapView = View.extend({
     if (!deps.layerGroupModel) throw new Error('layerGroupModel is required');
 
     this._mapModel = this.map = deps.mapModel;
+    this._tileErrorCollection = this.tileErrorCollection = new TileErrorCollection();
     this._visModel = deps.visModel;
     this._cartoDBLayerGroup = deps.layerGroupModel;
 
+    window.tileErrorCollection = this.tileErrorCollection;
```
xavijam: 💥

src/geo/map-view.js
```diff
@@ -30,9 +32,12 @@ var MapView = View.extend({
     if (!deps.layerGroupModel) throw new Error('layerGroupModel is required');
 
     this._mapModel = this.map = deps.mapModel;
+    this._tileErrorCollection = this.tileErrorCollection = new TileErrorCollection();
```
xavijam: ?

src/geo/map-view.js
```diff
@@ -30,9 +32,12 @@ var MapView = View.extend({
...
     this._visModel = deps.visModel;
```
xavijam: deps?

src/geo/map-view.js
```diff
@@ -30,9 +32,12 @@ var MapView = View.extend({
...
     this._cartoDBLayerGroup = deps.layerGroupModel;
 
+    window.tileErrorCollection = this.tileErrorCollection;
+
     this.add_related_model(this.map);
```
xavijam: do we need this?

src/util/tile-error-collection.js
```diff
@@ -0,0 +1,77 @@
...
+      if (!isInQueue) this.queue.add(model);
+
+      model.get('node').src = TILE_ERROR_IMAGE;
+
+      if (!this.running) this._getTileErrors();
+    });
+  },
+
+  resetErrorTiles: function () {
+    var models = this.filter(function (model) {
```
xavijam: review this method, could we use proper "backbone" methods?

xavijam: Shouldn't we stop the "checker" when the map has changed?

src/util/tile-error-collection.js
```diff
@@ -0,0 +1,77 @@
...
+    });
+
+    if (!model) {
+      this.running = false;
+      return;
+    }
+
+    this.running = true;
+
+    if (this._deletedNode(model.get('node'))) {
```
xavijam: Do we need this?

src/util/tile-error-collection.js
rubenmoya: I've been doing some research and I can only see `remove` and `reset`, the main difference is remove triggers a lot more events than reset

xavijam:
Please, update the description with more information about these changes, and then @alonsogarciapablo, if you can take a look, it would be great 💋

rubenmoya:
Done!

alonsogarciapablo:
Taking a look!

src/geo/gmaps/gmaps-cartodb-layer-group-view.js
```diff
@@ -84,9 +84,9 @@ var GMapsCartoDBLayerGroupView = function (layerModel, gmapsMap) {
   // lovely wax connector overwrites options so set them again
   // TODO: remove wax.connector here
   _.extend(this.options, opts);
-  GMapsLayerView.call(this, layerModel, gmapsMap);
+  GMapsLayerView.apply(this, arguments);
```
nobuti: I don't like this, I'd rather to be explicit.

test/spec/util/tile-error-collection.spec.js
```diff
@@ -0,0 +1,196 @@
...
+    };
+
+    it('should set the error tile overlay when a new model is added', function () {
+      spyOn(document.body, 'contains').and.returnValue(true);
+      this.collection.add(tile);
+      var model = _.last(this.collection.models);
+      expect(model.get('tileDomNode').src.indexOf('data:image/svg+xml;base64,')).not.toBe(-1);
+    });
+
+    it('should add the model to the qeue if is not there', function () {
```
alonsogarciapablo: Perhaps worth adding a `describe('when running is true'...` to keep the test more consistent?

test/spec/util/tile-error-collection.spec.js
```diff
@@ -0,0 +1,196 @@
...
+  describe('.getError', function () {
+    it('should return an error if any tile have it', function () {
+      this.collection.models[0].set('error', this.error);
+
+      expect(this.collection.getError()).toEqual(this.error);
+    });
+  });
+
+  describe('.resetErrorTiles', function () {
+    it('should remove the tiles if they\'re not in the DOM', function () {
```
alonsogarciapablo: This test is a bit tricky to understand without having to know what `_deletedNode` does... Why not add some tiles to the collection and some of them to the DOM and calling `resetErrorTiles`?

test/spec/util/tile-error-collection.spec.js
```diff
@@ -0,0 +1,196 @@
...
+      spyOn(this.collection, '_deletedNode').and.returnValue(true);
+      this.collection.resetErrorTiles();
+
+      expect(this.collection.models).toEqual([]);
+    });
+  });
+
+  describe('._getTileErrors', function () {
+    describe('when we alerady have the error or there are no models to check', function () {
```
alonsogarciapablo: s/alerady/already

src/util/tile-error-collection.js
```diff
@@ -0,0 +1,80 @@
...
+
+    this.on('add', this._onAdd, this);
+  },
+
+  resetErrorTiles: function () {
+    this.running = false;
+    this.queue.reset();
+
+    var models = this.filter(function (model) {
+      this._deletedNode(model.get('tileDomNode'));
```
alonsogarciapablo: Don't we need a `return` here?

src/util/tile-error-collection.js
```diff
@@ -0,0 +1,80 @@
...
+    if (!this.running) {
+      this.running = true;
+      this._getTileErrors();
+    }
+  },
+
+  _getTileErrors: function () {
+    if (!this.running) return;
+
+    var model = this.queue.find(function (model) {
```
alonsogarciapablo: Why not just remove the model from the queue to ensure is not checked again?

src/util/tile-error-collection.js
```diff
@@ -0,0 +1,80 @@
...
+    return model && model.get('error');
+  },
+
+  _onAdd: function (model) {
+    var isInQueue = this.queue.findWhere({ url: model.get('url') });
+    if (!isInQueue) this.queue.add(model);
+
+    model.get('tileDomNode').src = TILE_ERROR_IMAGE;
+
+    if (!this.running) {
```
alonsogarciapablo: Would `if (this.getError()) { ... } ` work here too?

src/util/tile-error-collection.js
```diff
@@ -0,0 +1,80 @@
...
+      success: function () {
+        model.get('tileDomNode').src = model.get('url');
+      },
+      error: function (jqXHR) {
+        var errors = getValue(jqXHR, 'responseJSON.errors_with_context', []);
+        model.set('error', errors[0]);
+      },
+      complete: function () {
+        model.set('checked', true);
+        this._getTileErrors();
```
alonsogarciapablo: Why do we need this here? We know this tile is either correct or has errors, why not just wait until another tile is added to this collection?

src/util/tile-error-collection.js
```diff
@@ -0,0 +1,80 @@
...
+    this.queue.reset();
+
+    var models = this.filter(function (model) {
+      this._deletedNode(model.get('tileDomNode'));
+    }.bind(this));
+
+    this.reset(models);
+  },
+
+  getError: function () {
```
alonsogarciapablo: Is this something that has to be public? Are we going to expose that via the public API (eg: to BUILDER)?

src/util/tile-error-collection.js
rubenmoya: Because it could fail when loading in the `img`, and then the ajax request be successful, but fail again, at this could go on forever, so I have to make sure we just make one ajax call.

src/util/tile-error-collection.js
rubenmoya: As said above, it could fail when loading the image, the ajax request be successful, but fail again, so we have to prevent another ajax request

src/util/tile-error-collection.js
rubenmoya: Yes, that was the idea, right now it has no use, but the idea was to use it in the next issues related to limits

src/geo/gmaps/gmaps-cartodb-layer-group-view.js
rubenmoya: We were using `arguments` in most of the places related to this, so I thought it was the best idea to keep it consistent through all the files

src/geo/leaflet/leaflet-layer-view.js
```diff
@@ -1,15 +1,18 @@
 var _ = require('underscore');
 var Backbone = require('backbone');
 
-var LeafletLayerView = function (layerModel, leafletMap) {
+var LeafletLayerView = function (layerModel, leafletMap, mapModel) {
```
alonsogarciapablo: This is the base class for all layer "views" (`LeafletTiledLayerView`, `LeafletCartoDBLayerView`, etc...). Do we want to track errors in all those types of views? Or... do we only want to check errors in the case of a `LeafletCartoDBLayerView` which is the one that will go against the Maps API? We could push this down to `src/geo/leaflet/leaflet-cartodb-layer-group-view.js` in that case.

(this would also apply to `src/geo/gmaps/gmaps-layer-view.js`).

src/util/tile-error-collection.js
alonsogarciapablo: We're always setting `checked` to `true` after the request is done and then we filter those "tiles"... isn't that the same as removing it from the queue so that it doesn't get "processed" again?

src/geo/leaflet/leaflet-layer-view.js
rubenmoya: I put it there because I wasn't sure which ones will track errors, if I'm not mistaken in the case of leaflet there are two views right? `LeafletCartoDBLayerGroupView ` and `LeafletTiledLayerView `

src/util/tile-error-collection.js
rubenmoya: Yes, but then in `_onAdd` we check it's not in the queue so we don't add it again

src/geo/leaflet/leaflet-layer-view.js
alonsogarciapablo: [There are more types](https://github.com/CartoDB/cartodb.js/tree/v4/src/geo/leaflet)... many of those won't trigger the `tileerror`. For example, we use a `LeafletTiledLayerView` to render the basemap (from CARTO, or other providers). I don't think we want to check tiles in that case because CARTO might not be serving those tiles...

rubenmoya: So, we should just check `LeafletCartoDBLayerGroupView` and `GMapsCartoDBLayerGroupView`, right?

alonsogarciapablo: I think so, yes.

rubenmoya: 👍

test/spec/util/tile-error-collection.spec.js
rubenmoya: How would you add them to the DOM here?

src/util/tile-error-collection.js
alonsogarciapablo: Why would a tile be added to this collection more than once?

src/util/tile-error-collection.js
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    $.ajax({
+      url: model.get('url'),
+      success: function () {
+        model.get('tileDomNode').src = model.get('url');
+      },
+      error: function (jqXHR) {
+        var errors = getValue(jqXHR, 'responseJSON.errors_with_context', []);
+        model.set('error', errors[0]);
```
alonsogarciapablo: Errors in `responseJSON.errors_with_context` are objects, in other cases ([here](https://github.com/CartoDB/cartodb.js/blob/v4/src/windshaft/map-base.js#L116-L118)) we wrap them inside [this class](https://github.com/CartoDB/cartodb.js/blob/v4/src/windshaft/error.js) so that we return a known object to users of CartoDB.js. Do you think it'd be nice to do the same / something similar here?

src/util/tile-error-collection.js
rubenmoya: Because how Maps API works, when requesting a tile, it can timeout in 2 places, in an "internal queue" they have or in the database, if it fails in the database it will always fail, but there is a case when it will timeout in the queue, the database will finish and cache the tile, and the next request (the AJAX one) will work, that's ok in most of the cases, because the cache will keep alive, but in local (or maybe in onpremise?) there is a configuration which resets the cache on hit, so the first request for loading the image will fail, the ajax one will return a 200, and the next try to load the image (via src) will fail again, because the cache has been removed, that leads to an infinite loop of adding to the queue, status 200, fail, adding to the queue, status 200, fail, so the solution I found was to mark them as "checked" so we don't add them multiple times when this happens.

test/spec/util/tile-error-collection.spec.js
alonsogarciapablo: Couldn't we create `` elements and append them to (or remove them from) `document.body`?

src/util/tile-error-collection.js
rubenmoya: That's a pretty good idea, I think we could use the same class, but I don't get why we do `this.type = error.subtype`, in the response I just have:

```
{
  "type":"limit",
  "message":"You are over platform's limits. Please contact us to know more details"
}
```

test/spec/util/tile-error-collection.spec.js
rubenmoya: I'm not sure, I was having problems with `document.body.containts` because `document.body` was undefined, so I had to mock that call, but I'll give it a try.

src/util/tile-error-collection.js
alonsogarciapablo: So far, we only had "layer", "analysis" and "unknown" errors. The first two were assigned to the corresponding layer or analysis model, and we used `subtype` as the type of the error (eg: `turbocarto`). Think about an error related to a layer... we might want to display the error differently if it's an error related to the "sql" or if it has something to do with the "turbocarto"... That's why needed to know the type of error (from the subtype).

"Unknown" errors were treated as "global" errors and that's the type of the error wasn't really important... ".

[This test](https://github.com/CartoDB/cartodb.js/blob/v4/test/spec/vis/model-updater.spec.js#L642-L695) might help you understand how errors from Maps API are assigned to layer models.

Perhaps the `type` is not that important in this case. We can come back to this once you implement the change that will use this error....

src/util/tile-error-collection.js
alonsogarciapablo: Thanks for the explanation. Still, isn't `tileerror` (or `tile.onerror` for Google Maps) only called once for each tile?

rubenmoya: No, because when the ajax call returns a `200`, I update the image `src` with the tile url, which could fail again, triggering another `tileeror`

alonsogarciapablo: Ok, that explains a lot of things! I thought `tileerror` was only being triggered when Leaflet tried to load each of the tiles, but I wasn't expecting it'd be triggered when we change the `src` of the tile afterwards.

xavijam:
So, are we ready for testing this in staging?

alonsogarciapablo:
I don't have much more to say now that I understand this a bit better.

I'd ❤️ to see those tests follow a more outside-inside approach (eg: ensuring AJAX requests are triggered when tiles are added to the collection under different circumstances) instead of relying on private methods so much (mainly, because we're coupling tests with the implementation and any change in the code would break the tests...). But I'll leave this up to @rubenmoya...

Only pending thing would be what we discussed in https://github.com/CartoDB/cartodb.js/pull/1721#discussion_r136111171. I'd just encapsulate the error in a WindshaftError and forget about the `type`.

rubenmoya:
Since I had to change the `.svg` file (with a svg loader in browserify) to a `.tpl` to avoid errors with webpack in builder (and theuse of a svg-loader just for this icon) I added a few more tests which check the requests amount.

alonsogarciapablo:
> About the error, if I encapsulate it in a WindshaftError, we will lose track of what kind of error it is, or are we assuming it will be a limits error and that's it? @alonsogarciapablo

We can assume it's a "global" error for now.

rubenmoya:
@alonsogarciapablo If you check my last commits I did some changes to  `WIndshaftError`, addidng a check for `limit` error and keeping the `type` in that case, what do you think?

src/constants.js
```diff
@@ -14,5 +14,11 @@ module.exports = {
     LEAFLET: 'leaflet'
   },
 
-  GMAPS_BASE_LAYER_TYPES: ['roadmap', 'gray_roadmap', 'dark_roadmap', 'hybrid', 'satellite', 'terrain']
+  GMAPS_BASE_LAYER_TYPES: ['roadmap', 'gray_roadmap', 'dark_roadmap', 'hybrid', 'satellite', 'terrain'],
```
alonsogarciapablo: This file can become a little bit of a disaster if we keep adding stuff to it. I'd prefer to declare constants in the files that they "belong to" (eg: https://github.com/CartoDB/cartodb.js/blob/v4/src/analysis/analysis-model.js#L145). But that's just my opinion!

rubenmoya: Yes, I talked about this in another PR, the idea if the file starts becoming a disaster is moving it to a `constants` folder, and then have different files for each type of constants, for exampe `constants/layer-types.js`, `constants/errors.js`, `constants/widget-types.js`, etc

alonsogarciapablo:
> @alonsogarciapablo If you check my last commits I did some changes to WIndshaftError, addidng a check for limit error and keeping the type in that case, what do you think?

The `type` of the error is only used to determine what model it should be associated to when a request to the Maps API fails ([here's the code](https://github.com/CartoDB/cartodb.js/blob/v4/src/windshaft-integration/model-updater.js#L239-L247)). We're introducing a new type, but since there's no specific model to associate it to, it'll go to the visModel (i.e: visModel will trigger an event) when a request fails. As I mentioned already, these "global" errors don't get a type.

In summary: it won't have any effects really and we'll have to think about errors again when release the new version of CARTO.js.

src/geo/map.js
```diff
@@ -87,6 +88,24 @@ var Map = Model.extend({
 
   // PUBLIC API METHODS
 
+  addError: function (error) {
```
nobuti: Have you thought how error events are going to be transmitted to other components (Builder, DI...)?

rubenmoya: I thought about doing something like:

```
model.errors.on('add', function () {
  // Do something here
});
```
in the map model in deep insights integration, but to be honest I don't know if that's the best approach...

nobuti: How about trigger an `error` event from the model with the error type?

rubenmoya: So, trigger an error Instead of having a collection of errors?

nobuti: no, whenever an error is added to the collection, instead of listening the `add` event's collection, we could trigger an event from the map model itself, with the error type. This way we can avoid finding the error type in the consumer. Something like:
```
model.trigger('error', {type: error.type});
or 
model.trigger('error:limits') (with namespace)
```

nobuti: Just an idea, we need to think what happens when the error is gone, could this happen in the middle of a session?

rubenmoya: +1 for the idea of trigger the error with the type.

About the error gone, I'm not sure, in the limits case, how do we know the error is not happening anymore?

alonsogarciapablo:
Sure! I'll take a look later!

src/geo/map.js
alonsogarciapablo: > Just an idea, we need to think what happens when the error is gone, could this happen in the middle of a session?

Is this something we need to worry about?

src/geo/leaflet/leaflet-cartodb-layer-group-view.js
```diff
@@ -43,18 +46,25 @@ var findContainerPoint = function (map, o) {
...
   });
 
-  this.leafletLayer.on('loading', function (e) {
+  this.leafletLayer.on('loading', function () {
     self.trigger('loading');
   });
+
+  this.leafletLayer.on('tileerror', function (layer) {
+    layer.tile.src = TILE_ERROR_IMAGE;
+    self.mapModel.addError({ type: C.WINDSHAFT_ERRORS.TILE });
```
alonsogarciapablo: I think I already mentioned this in a previous CR, but I think it might make sense so I'd like to bring it up again... 

Both `src/geo/leaflet/leaflet-cartodb-layer-group-view.js` and `src/geo/gmaps/gmaps-cartodb-layer-group-view.js` use the same model: https://github.com/CartoDB/cartodb.js/blob/v4/src/geo/cartodb-layer-group.js. This model (virtually) "groups" layers of type "CartoDB". Since the tile errors that we're handling in this PR are related to this group of layers, I think it'd make more sense to "store" those error in the `CartoDBLayerGroup` (i.e: I'd move `addError` and the `errors` collection from `MapModel` to `LayerGroupModel`.

I'm asking because we're changing CartoDB.js right now and the new feature that we're introducing will allow developers to use `src/geo/leaflet/leaflet-cartodb-layer-group-view.js` and `src/geo/gmaps/gmaps-cartodb-layer-group-view.js` without the need of a `MapModel` (developers will work directly with the Leaflet or GoogleMaps map). We'll need to expose tile/limit errors in there too.

It should be a pretty straightforward change. We'd do `self.model.addError(...)` instead of `self.mapModel.addError(...)` here.

src/geo/map.js
```diff
@@ -87,6 +88,28 @@ var Map = Model.extend({
...
+
+    var inCollection = this.getError(type);
+    return inCollection || this.errors.add(error);
+  },
+
+  getError: function (type) {
+    return this.errors.findWhere({ type: type });
+  },
+
+  removeError: function (type) {
```
alonsogarciapablo: This is not currently being used, right?

src/geo/leaflet/leaflet-cartodb-layer-group-view.js
rubenmoya: Do you think we should keep a collection of errors? Or it's better just to trigger the `error:limit` / `error:tile` and that's it?

src/geo/gmaps/gmaps-cartodb-layer-group-view.js
```diff
@@ -223,13 +224,17 @@ _.extend(
...
       };
-      im.onload = finished;
-      im.onerror = function () {
+
+      tile.onload = finished;
+
+      tile.onerror = function () {
         Profiler.metric('cartodb-js.tile.png.error').inc();
+        tile.src = TILE_ERROR_IMAGE;
+        self.model.addError({ type: C.WINDSHAFT_ERRORS.TILE });
```
alonsogarciapablo: Isn't this error related with limits too? Do we need `C.WINDSHAFT_ERRORS.TILE` and `C.WINDSHAFT_ERRORS.LIMIT`?

src/geo/cartodb-layer-group.js
```diff
@@ -19,6 +19,16 @@ var CartoDBLayerGroup = Backbone.Model.extend({
     this._layersCollection = options.layersCollection;
   },
 
+  addError: function (error) {
```
alonsogarciapablo: LGTM! We can always change this if we need to "store" the error or if we need to be notified when the error is gone.

src/geo/gmaps/gmaps-cartodb-layer-group-view.js
rubenmoya: Since we don't really know if the error is because the server is down, or because of limits, I used `TILE`, just to know that tiles are not loading, and be able to show a generic error.

xavijam:
How come is this not merged? @rubenmoya @alonsogarciapablo

nobuti:
We need to implement errors notification in the Builder, otherwise, the user won't be aware what's going on.

xavijam:
But we can release it, because if we don't change the user configuration it will not be displayed, so, in my opinion, the sooner the better...

xavijam:
Although the error tile will be displayed, I feel we are ok with that, aren't we?

rubenmoya:
We already have a tile with an error and no notifications in Builder... so it will be the same, and as you said, unless we change the engine configuration, it will be de same as it is right now, so it's a 🏁  for me.

xavijam:
@nobuti @noguerol ?

nobuti:
👍

PR #1722, opened 2017-08-24
Fix icons in category legends

Jesus89:
Related to: https://github.com/CartoDB/cartodb/issues/11630

Now the icons are shown in the category legend

category-legends

### DIFF
--- a/src/windshaft-integration/legends/rule-to-category-legend-adapter.js
+++ b/src/windshaft-integration/legends/rule-to-category-legend-adapter.js
@@ -18,10 +18,14 @@ var generateCategories = function (bucketsColor, bucketsIcon) {
     });
     return {
       title: bucketColor.filter.name,
-      icon: bucketIcon ? bucketIcon.value.slice(5, -2) : '',
+      icon: bucketIcon ? _extractURL(bucketIcon.value) : '',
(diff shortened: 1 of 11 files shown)


src/windshaft-integration/legends/rule-to-category-legend-adapter.js
```diff
@@ -11,12 +11,23 @@ var isEveryBucketValid = function (rule) {
...
+    return {
+      title: bucketColor.filter.name,
+      icon: bucketIcon ? _extractURL(bucketIcon.value) : '',
+      color: bucketColor.value
+    };
   });
 };
 
+var _extractURL = function (str) {
+  return str.slice(5, -2);
```
matallo: be wary of magic numbers, I understand why, but the source value not always come quoted, actually it doesn't appear quoted in the docs https://carto.com/docs/carto-engine/cartocss/properties/#uri

shouldn't we better match the url inside the `url()` rahter than rely on the position of the url in the string?

Jesus89: Yeah, I was hoping for a better method to do this. Do you recommend regex for the url?

PR #1724, opened 2017-08-29
Create geometries unexpected behaviour in IE

ivanmalagon:
Fixes one of the problems related to https://github.com/CartoDB/support/issues/915

IE caches Ajax GET requests and the state of the map and its current visualization are different.

Added a `nocache` param to the request URL to avoid caching.

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,6 +81,7 @@
     "semver": "~4.1.0",
     "source-map-support": "CartoDB/node-source-map-support#0.4.6-cdb1",
     "time-grunt": "~0.3.1",
+    "url": "0.11.0",
     "watchify": "3.4.0"
   },
   "browserify": {
--- a/src/windshaft/client.js
+++ b/src/windshaft/client.js
@@ -13,6 +13,15 @@ var validatePresenceOfOptions = function (options, requiredOptions) {
   }
 };
 
+var getInstantiationParams = function (params) {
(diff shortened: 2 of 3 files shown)


ivanmalagon:
Question to @rochoa and perhaps @alonsogarciapablo 

In Internet Explorer we (mainly BCG) run into problems with the map instantiation. IE caches Ajax GET requests and it's returning always the initial map when doing anything that needs a map reinstantiation. In the related ticket, for instance, when we add new features (geometry) the map doesn't reflect the just created geometries.

I've added the typical `nocache` query parameter with the timestamp and it works. My question is: Can we cause another problems because of doing this? I don't think so because we indeed reinstantiate the map so we don't need it cached but I don't know if there's any edge case I'm not thinking on.

Of course, the param is only attached in IE, as you can see in the commit.

rochoa:
I think this is already solved here (or it should). When triggering a map instantiation from Builder, the callback param is auto-incremented to bust the cache.

ivanmalagon:
After checking the issue with @rochoa it seems likely that is a matter of APIs versions or configurations. In the very same machine, I'm reproducing the error in dev mode but it works flawlessly in production. So perhaps it's a problem that can be solved with no changes in frontend code.

We'll do some checks tomorrow to get to the heart of the problem.

Issue #1727, opened 2017-09-05
Improve createVis tests


IagoLast:
As part of the next carto.js version the `createVis` method is going to be changed.

To ensure this changes wont break anything we decided to add several tests against the generated visModel.

rubenmoya:
Using Jest? :D

IagoLast:
@rubenmoya the original idea was to [compare](https://nodejs.org/api/assert.html#assert_assert_notdeepequal_actual_expected_message) the actual vs expected but since serialising the visModel is not trivial we are going to add regular tests.

Issue #1728, opened 2017-09-05
Allow creating an empty VisModel and setting its state programmatically


alonsogarciapablo:
- [ ] Decouple VisModel form VizJSON ([here](https://github.com/CartoDB/cartodb.js/blob/v4/src/vis/vis.js#L124))
- [ ] Define the visModel public API
- [ ] Update unit tests for this api

After this PR, you should be able to create a new VisModel and manage layers, dataviews, analysis, etc.

alonsogarciapablo:
Wrong approach. Closing in favor of https://github.com/CartoDB/cartodb.js/issues/1749.

PR #1740, opened 2017-09-05
Cartojs 1727

IagoLast:
#1727 **Add tests to createVis**

> As part of the next carto.js version the createVis method is going to be changed.
> To ensure this changes wont break anything we decided to add several tests against the generated visModel.


src/vis/vis.js
```diff
@@ -147,8 +147,9 @@ var VisModel = Backbone.Model.extend({
       windshaftSettings: windshaftSettings
     });
 
+    var allowScrollInOptions = (vizjson.options && vizjson.options.scrollwheel) || vizjson.scrollwheel;
```
alonsogarciapablo: I think we can safely ignore `vizjson.scrollwheel` now that we now it's not part of the viz.json.

IagoLast: If we ignore them several tests will fail...

alonsogarciapablo: We should fix them IMO.

IagoLast: agreed but is out of the scope of this issue

alonsogarciapablo: Then we should leave this as it was and open an issue to fix the bug. Otherwise we might never get back to it and change will remain outdated.

IagoLast:
This is the list with the current test suite for `createVis`.


screen shot 2017-09-07 at 13 04 07

* The Options suite should be updated once we have the spec given by #1743
* Some suites could be removed once we merge #1728 since they are low-level and very sensible.

test/spec/api/createVis/create-vis.spec.js
```diff
@@ -0,0 +1,344 @@
...
+        setTimeout(function () {
+          expect(visModel._instantiateMapWasCalled).toEqual(false);
+          done();
+        }, 25);
+      });
+    });
+  });
+
+  describe('VisModel.map', function () {
+    it('should have the right title', function () {
```
alonsogarciapablo: Isn't this very similar to: https://github.com/CartoDB/cartodb.js/pull/1740/files#diff-5707dcdc5d2b4ea40894b5c990ab7c97R58?

alonsogarciapablo: Same for some of 👇

test/spec/api/createVis/create-vis.spec.js
```diff
@@ -0,0 +1,344 @@
...
+    expect(Loader.get).toHaveBeenCalledWith('www.example.com/fake_vis.json', jasmine.any(Function));
+  });
+
+  describe('Default (no Options)', function () {
+    it('should get the map center from the visJson', function () {
+      var visJson = scenarios.load(0);
+      var visModel = createVis(this.containerId, visJson);
+      expect(visModel.map.get('center')).toEqual(visJson.center);
+    });
+    it('should get the title from the visJson', function () {
```
alonsogarciapablo: I would give these guys (the `it`s) a little bit of space to breath...

test/spec/api/createVis/create-vis.spec.js
```diff
@@ -0,0 +1,344 @@
...
+    }.bind(this)).not.toThrowError();
+
+    expect(function () {
+      createVis(this.containerId, 'vizjson');
+    }.bind(this)).not.toThrowError();
+  });
+
+  it('should use the given vis.json (instead downloading) when the visjson parameter is provided', function () {
+    spyOn(Loader, 'get');
+    var visJson = scenarios.load(0);
```
alonsogarciapablo: This is pretty cool but we only have one scenario right now so we don't really need it... In any case, what about naming the scenario instead of using a number?

PR #1742, opened 2017-09-06
Tangram.cartodb 0.5.x

rochoa:


IagoLast:
any special reason to freeze the version (`^0.5.3` instead `0.5.3`)?

rochoa:
Sorry, I didn't see this before.

When I rely on an API (tangram.cartodb and tangram-cartocss in this case) that is in not stable enough and that can change [0], I tend to use a fixed version to be sure about what API I'm consuming. However, in this scenario, where we control both libraries and we are having to fix a few issues during a short period of time, I decided to use any patch version. Does that make sense?

[0]
> [#4](http://semver.org/#spec-item-4) Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable.

IagoLast:
100% agreed

rochoa:
This is ready to be merged.

Issue #1737, opened 2017-09-05
Setup ESLint


alonsogarciapablo:
A lot of projects are now using [ESLint](https://eslint.org/).

It integrates nicely with many [editors](https://eslint.org/docs/user-guide/integrations#editors) and it supports Semistandard (via [this module(s)](https://www.npmjs.com/package/eslint-config-semistandard)).

As part of this, we should cleanup https://github.com/CartoDB/cartodb.js/blob/v4/.eslintignore and fix existing errors.

Finally, we'll have to revisit the script to run the linter (currently using grunt +  semistandard).

PR #1748, opened 2017-09-11
Improve lint task

Jesus89:
Related to: #1737

Changes:
- Use eslint grunt-task instead of semistandard binary
- Update semistandard package to 11.0.0
- Remove semistandard config from package.json
- Fix .eslintignore file
- Update .eslintrc.json file
- Fix various files lint
- Add eslint to hound


Gruntfile.js
```diff
@@ -1,6 +1,19 @@
 var _ = require('underscore');
 var jasmineCfg = require('./grunt/tasks/jasmine');
 
+function getTargetDiff () {
```
IagoLast: I would add a comment explaining what this method does and I think is better to use it instead saving a execution on a variable.

src/core/loader.js
```diff
@@ -32,7 +32,8 @@ var Loader = {
   getPath: function (file) {
     var scripts = document.getElementsByTagName('script');
     var cartodbJsRe = /\/?cartodb[\-\._]?([\w\-\._]*)\.js\??/;
-    for (i = 0, len = scripts.length; i < len; i++) {
+    var i;
+    for (i = 0; i < scripts.length; i++) {
```
IagoLast: since `i` is only declared in the loop I think we should do:

```javascript 
for (var i = 0; i < scripts.length; i++) {   
```

src/core/profiler.js
```diff
@@ -73,6 +73,7 @@ Profiler.new_value = function (name, value, type, defer) {
 };
 
 Profiler.print_stats = function () {
+  var k;
   for (k in Profiler.metrics) {
```
IagoLast: [same](https://github.com/CartoDB/cartodb.js/pull/1748/files/6b10c850d25d5905844aecd06e278bacb3a5ddaa..b7ee6bc2d51ea8ea935cb686891ebb96707447c5#r138163309) here

PR #1753, opened 2017-09-12
Client windshaft map split

IagoLast:
Move request logic from `map-base` to `windshaft client`


src/windshaft/client.js
```diff
@@ -169,4 +176,59 @@ WindshaftClient.prototype._jsonpCallbackName = function (payload) {
...
+};
+
+/**
+ * 
+ */
+WindshaftClient.prototype._performRequest = function (request) {
+  var payload = request.payload;
+  var params = request.params;
+  var options = request.options;
+  this.instantiateMap({
```
alonsogarciapablo: `WindshaftClient#instantiateMap` is no longer being used externally so I'd merge it with `WindshaftClient#performRequest`. There's no reason to have both of them IMO.

src/windshaft/client.js
```diff
@@ -169,4 +176,59 @@ WindshaftClient.prototype._jsonpCallbackName = function (payload) {
...
+  var request = new Request(payload, params, options);
+  if (this._canPerformRequest(request)) {
+    this._performRequest(request);
+  } else {
+    log.error('Maximum number of subsequent equal requests to the Maps API reached (' + MAP_INSTANTIATION_LIMIT + '):', payload, params);
+    options.error && options.error();
+  }
+};
+
+/**
```
alonsogarciapablo: Should we remove this? Perhaps we can add JSDoc to the public method and document it a bit.

alonsogarciapablo:
Also, [these tests](https://github.com/CartoDB/cartodb.js/blob/v4/test/spec/windshaft/map-base.spec.js#L115-L202) can now be moved to `test/spec/windshaft/client.spec.js`.

src/windshaft/client.js
```diff
@@ -36,9 +42,32 @@ var WindshaftClient = function (settings) {
...
 };
 
-WindshaftClient.prototype.instantiateMap = function (options) {
+WindshaftClient.prototype.instantiateMap = function (payload, params, options) {
+  var request = new Request(payload, params, options);
+  if (this._requestTracker.canRequestBePerformed(request)) {
+    this._instantiateMap({
+      mapDefinition: request.payload,
+      params: request.params,
+      success: function (response) {
```
alonsogarciapablo: As you said, it's kind of confusing that both `instantiateMap` and `_instantiateMap` have to deal with `success` and `error` options.

Here's something that could simplify this IMO:

```
if (this._requestTracker.canRequestBePerformed(request)) {
  this._instantiateMap(request, {
    complete: function (response) {
      this._requestTracker.track(request, response);
    }
  });
} else {
  ...
}
```
Another option could be to move all the `success` and `error` logic from `_instantiateMap` to `instantiateMap` so that `_instantiateMap` just "forwards" `success` and `error` options to `_get` and `_post`.

Also, I guess the following chunk could be extracted into it's own method to simplify things (`_performRequest(...)`?):

```
  var encodedURL = this._generateEncodedURL(mapDefinition, params);
  if (this._isURLValid(encodedURL)) {
    this._get(encodedURL, ajaxOptions);
  } else {
    this._generateCompressedURL(mapDefinition, params, function (compressedURL) {
      if (this._isURLValid(compressedURL)) {
        this._get(compressedURL, ajaxOptions);
      } else {
        var url = this._getURL(params, 'post');
        this._post(url, mapDefinition, ajaxOptions);
      }
    }.bind(this));
  }
```

What do you think?

test/spec/windshaft/client.spec.js
```diff
@@ -273,4 +273,30 @@ describe('windshaft/client', function () {
...
+    beforeEach(function () {
+      client = new WindshaftClient({
+        urlTemplate: 'https://{user}.example.com:443',
+        userName: 'rambo'
+      });
+    });
+
+    it('should make a request when the request service allows it', function () {
+      spyOn(client, '_instantiateMap').and.callFake(function () {});
+      spyOn(client._requestTracker, 'canRequestBePerformed').and.returnValue(true);
```
alonsogarciapablo: If we really need to spy on private methods, I'd at least abstract this (`client._requestTracker.canRequestBePerformed`) in a method (as it was before) so that we don't have to know about the internals of the `RequestTracker` here...

test/spec/windshaft/client.spec.js
```diff
@@ -22,7 +22,7 @@ describe('windshaft/client', function () {
     });
 
     it('should trigger a GET request to instantiate a map', function () {
-      this.client.instantiateMap({
+      this.client._instantiateMap({
```
alonsogarciapablo: Does this really make a difference? Why not keep it so that we test the public interface of this object?

examples/viz/viz2.json
```diff
@@ -1,203 +1 @@
...
... class=\"loader\" original-title=\"\"></div>"},{"type":"logo","order":9,"options":{"display":true,"x":10,"y":40},"template":""}],"title":"Untitled Map","updated_at":"2017-09-13T09:46:12+00:00","user":{"fullname":"Iago Lastra","avatar_url":"https://s3.amazonaws.com/com.cartodb.users-assets.production/production/iago-carto/assets/20170720105148Avatar250.png","profile_url":"https://team.carto.com/u/iago-carto"},"version":"3.0.0","widgets":[{"id":"1a0e4a48-1a43-4d4e-8fce-773a370b8041","type":"category","title":"population","order":0,"layer_id":"636838fb-db9a-46b6-91d5-74ab9884111d","options":{"column":"population","sync_on_data_change":true,"sync_on_bbox_change":true,"aggregation":"count","aggregation_column":"","column_type":"number","schema":{},"aggregate":{"attribute":"","operator":"count"},"suffix":"","prefix":""},"style":{"widget_style":{"definition":{"color":{"fixed":"#9DE0AD","opacity":1}},"widget_color_changed":false},"auto_style":{"custom":false,"allowed":true}},"source":{"id":"a0"}}],"zoom":6,"analyses":[{"id":"a0","type":"source","params":{"query":"SELECT * FROM \"iago-carto\".lugares"},"options":{"table_name":"\"iago-carto\".lugares","simple_geom":"point"}}],"vector":false}
```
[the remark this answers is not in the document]
IagoLast: Do we have any public account for testing?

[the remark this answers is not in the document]
alonsogarciapablo: No, we don't 😞 . We've typically point to datasets from https://documentation.carto.com/.

Issue #1749, opened 2017-09-11
Extract "Engine" class


alonsogarciapablo:
To be used internally in the VisModel and also in the upcoming `carto.leaflet.CartoLayerGroup` and `carto.googleMaps.CartoLayerGroup`.

![carto js engine idea](https://user-images.githubusercontent.com/390398/30275332-4aa1b2e4-9701-11e7-9e3f-5e003b0325f1.jpg)

An "Engine" will expose a method that will:

- "Serialise" given layers (and associated queries / analyses / dataviews) and send requests to Maps API.
- Update given models when request is received.
- Raise errors if something goes wrong.

## Tasks

- [x] Move request limiting responsibility from `src/windshaft/map-base.js` to `src/windshaft/client.js` (https://github.com/CartoDB/cartodb.js/pull/1753). 
- [ ] CartoLayer has a `source` that should be an Analysis node (source or analysis). We're using `sql` and `source` attributes before. Can remove support for `sql` attribute (https://github.com/CartoDB/cartodb.js/issues/1754).
- [ ] Merge `LayersCollection`, `DataviewsCollection`, and `AnalysisCollection` into CartoLayer hierarchy (see drawing above).
  - [ ] Source has `dataviews` and `filters` (might require changes to Windshaft Map).
  - [ ] TBD.

![image](https://user-images.githubusercontent.com/390398/30313681-794157a6-979e-11e7-8309-dd5c80333fd8.png)

- [ ] Extract classes that know how to generate payload for Maps API from an array / collection of layers (this is currently implemented in `src/windshaft/anonymous-map.js#toJSON` and `src/windshaft/named-map.js#toJSON`.
- [ ] Extract "Engine" class:
  - [ ] Mutate `src/windshaft/map-base.js` into the new "Engine" classes that uses `src/windshaft/client.js` and `src/windshaft-integration/model-updater.js` internally.
  - [ ] `VisModel#reload` uses the new "Engine".
  - [ ] `VisModel#setOk` and `VisModel#setErrors` are "handled" by `VisModel` instead of by `src/windshaft-integration/model-updater.js`.

alonsogarciapablo:
I exploded this issue into smaller ones so this one doesn't make sense anymore.

PR #1755, opened 2017-09-13
Layer's source MUST be an AnalysisModel

alonsogarciapablo:
Part of https://github.com/CartoDB/cartodb.js/issues/1749. Changes:

### Changes to `CartoDBLayer` and `TorqueLayer`

- These models had a `source` attribute that was the `id` of an analysis  node. In this PR we're changing things so that the attribute must point to an instance of `AnalysisModel`.

- Both classes now expose `getSource`, `getSourceId` to keep the `source` attribute "private".

While it'll still be possible to set a new source via the `update` method, a deprecation warning will be displayed when trying to do so. Newly added `setSource` is the recommended way of doing this.

### Changes to `Map#createCartoDBLayer` and `Map#createTorqueLayer`

These two methods were expecting an analysis id as the `source` attribute and now know how to work with AnalysisModels directly. Deprecation warnings are displayed when an id is given as the `source`.

### Support for `sql` option in `CartoDB` and `torque` layers

`sql` option for `CartoDB` and `torque` layers in viz.json will still be supported (see https://github.com/CartoDB/cartodb.js/issues/1754 for more info), but instead of using that attribute internally, the library will generate a "source" type analysis when the viz.json is loaded and layer models are generated.

### Changes to Dataview models

At some point in time, dataview models could not have a "source" and they would "inherit" the source from the layer they were associated too. This is no longer true and we can assume that deep-insights.js will always set a valid "source" when a dataview is created. This PRs removes some of that logic (I couldn't resist myself).

### Setup:
1. Checkout `layer-sources` branch locally.
2. Link BUILDER <-> deep-insigths.js <-> cartodb.js locally (via `npm|yarn link`).
3. Build BUILDER using `grunt dev` and launch the VM / server.
4. Open the UI in a browser and create an empty map.

### Things to test:

- [ ] Try adding a new layer (empty and from dataset).
- [ ] Add an analysis to the layer (this changes its source).
- [ ] Remove the analysis.
- [ ] Animate the layer (Style -> Animated) (this will replace the CartoDB layer for a Torque layer in cartodb.js)
- [ ] Add an analysis to the animated layer.
- [ ] Add a widget (uses a dataview internally).

You might see some warnings in the console, but you shouldn't see any errors and map should always look right! Thanks!

Related PRs
* https://github.com/CartoDB/cartodb/pull/12866
* https://github.com/CartoDB/deep-insights.js/pull/599


src/geo/map/cartodb-layer.js
```diff
@@ -120,6 +120,15 @@ var CartoDBLayer = LayerModelBase.extend({
...
     return stats && stats.estimatedFeatureCount;
+  },
+
+  getSourceId: function () {
+    var source = this.getSource();
+    return source && source.id;
+  },
+
+  getSource: function () {
+    return this.get('source');
```
alonsogarciapablo: Something to think about here is wether we want to keep using a Backbone attribute to store the the "source" or we should go with an instance variable (eg: `this._source`)... Good thing about Backbone attributes is that:
1. They can be changed "in bulk".
2. Changing an attribute will trigger a `change` event.

On the other hand, if we go with an instance variable, we'll need to implement a setter (`setSource`) and we'll have to be more explicit about side-effects (eg: reloading the visualisation).

src/geo/map.js
```diff
@@ -96,11 +97,17 @@ var Map = Model.extend({
 
   createCartoDBLayer: function (attrs, options) {
     this._checkProperties(attrs, ['sql|source', 'cartocss']);
+    if (attrs.source && !(attrs.source instanceof AnalysisModel)) {
+      throw new Error('CartoDB layers must have a valid source');
```
IagoLast: I think the error should be more specific
```javascript
throw new Error('CartoDB layers must have a AnalysisModel  as source');
```

src/geo/map.js
```diff
@@ -96,11 +97,17 @@ var Map = Model.extend({
...
+    }
     return this._addNewLayerModel('cartodb', attrs, options);
   },
 
   createTorqueLayer: function (attrs, options) {
     this._checkProperties(attrs, ['sql|source', 'cartocss']);
+    if (attrs.source && !(attrs.source instanceof AnalysisModel)) {
+      throw new Error('CartoDB layers must have a valid source');
```
IagoLast: [same](https://github.com/CartoDB/cartodb.js/pull/1755/files#r138819670) here

alonsogarciapablo:
@IagoLast I addressed your feedback in https://github.com/CartoDB/cartodb.js/pull/1755/commits/385f7b0a95f74442f103e222b2811766e345052c. I didn't mention `AnalysisModel` in the error message because it's not part of the actual API (analysis models are created via `vis.analysis.analyse` method) and also because we'll have to introduce a new class for "source" analyses). Thanks!

test/spec/api/createVis/scenarios/index.js
```diff
@@ -6,10 +6,14 @@ function load (index) {
   // We use a switch because our current build system doesn't support variables in the require.
   switch (index) {
     case 'basic':
-      return Object.assign({}, require('./basic_vis.json.js'));
+      return clone(require('./basic_vis.json.js'));
   }
 }
 
+function clone (object) {
```
IagoLast: Maybe we should move this methods to the `util` class

src/vis/vis.js
```diff
@@ -142,9 +142,18 @@ var VisModel = Backbone.Model.extend({
 
     var windshaftClient = new WindshaftClient(windshaftSettings);
 
+    // Create the public Analysis Factory
+    this.analysis = new AnalysisFactory({
```
IagoLast: IMO `this.analysisFactory` would be a more descriptive name

IagoLast: just realized that the AnalysisFactory is not a factory (software pattern) maybe we should look for a better name in the future

alonsogarciapablo: Agree and agree. Any suggestions? `AnalysisService`?

IagoLast: AnalysisService will do the work!

alonsogarciapablo: Ok. `vis.analysis` is part of the API that BUILDER is using now so I wouldn't rename it for now (given that the API is going to change anyway). I also agree about renaming `AnalysisFactory` to `AnalysisService` but let's do it in a separate PR to keep this one small.

alonsogarciapablo:
@IagoLast I pushed some changes since I asked you for the review (sorry!). I've updated the description of this PR to better reflect everything we're changing here. Thanks!

src/geo/map/cartodb-layer.js
```diff
@@ -129,6 +129,10 @@ var CartoDBLayer = LayerModelBase.extend({
 
   getSource: function () {
     return this.get('source');
+  },
+
+  setSource: function (source, options) {
+    this.set('source', source, options);
```
IagoLast: I assume when the source is changed a `change` event is thrown and the visModel/Engine is listening in order to reload the visualization, right?

IagoLast:
LGTM

src/geo/map/cartodb-layer.js
```diff
@@ -120,6 +120,26 @@ var CartoDBLayer = LayerModelBase.extend({
...
+  getSource: function () {
+    return this.get('source');
+  },
+
+  setSource: function (source, options) {
+    this.set('source', source, options);
+  },
+
+  update: function (attrs) {
+    if (attrs.source) {
```
ivanmalagon: Curious about one thing here. If `source` is set to `null` or `undefined` the error won't trigger but Backbone would set it to that value. That's the same as setting a new value.

I could be wrong but it's worth to check it and use `hasOwnProperty` better than Backbone's `has` (that answers no if it's null or undefined).

src/geo/map/torque-layer.js
```diff
@@ -164,7 +164,28 @@ var TorqueLayer = LayerModelBase.extend({
...
+    return this.get('source');
+  },
+
+  setSource: function (source, options) {
+    this.set('source', source);
+  },
+
+  update: function (attrs) {
+    if (attrs.source) {
+      throw new Error('"source" must be set via setSource');
```
ivanmalagon: [Same](https://github.com/CartoDB/cartodb.js/pull/1755/files#r139922676) here.

IagoLast:
TODO: https://github.com/CartoDB/cartodb.js/pull/1755#pullrequestreview-63920277

src/geo/map/cartodb-layer.js
```diff
@@ -100,6 +101,58 @@ var CartoDBLayer = LayerModelBase.extend({
...
+    var source = this.getSource();
+    return source && source.id;
+  },
+
+  getSource: function () {
+    return this.get('source');
+  },
+
+  setSource: function (source, options) {
+    this.set('source', source, options);
```
alonsogarciapablo: @IagoLast since this will be an API method... do you think we should check that `source` is a valid source here?

src/geo/map/cartodb-layer.js
```diff
@@ -100,6 +101,58 @@ var CartoDBLayer = LayerModelBase.extend({
...
+  },
+
+  /**
+   * Check if an analysis node is the layer's source.
+   * Only torque and cartodb layers have a source otherwise return false.
+   */
+  hasSource: function (analysisModel) {
+    if (!(analysisModel instanceof AnalysisModel)) {
```
alonsogarciapablo: `CartoDBLayer#equals` already throws a similar error. We should agree on what methods should trigger these kind of errors.

src/geo/map/cartodb-layer.js
IagoLast: I believe in order to make this method bakwards compatible (BC) it should accept both (string-id and AnalysisNode).

Once builder gets fixed we can remove the BC changes and add a check here

src/geo/map/cartodb-layer.js
IagoLast: * **pros:**  Easy to catch bugs in production, clients will be aware of their errors thanks to a clear trace.

* **cons:** Hard to test, no duck typing for mocks.

IMO we should do a type-check in every method which requires specific parameters

src/geo/map/cartodb-layer.js
alonsogarciapablo: This is still not being used externally so there's no need to make it backwards compatible IMO.

src/geo/map/cartodb-layer.js
alonsogarciapablo: > IMO we should do a type-check in every method which requires specific parameters

Every method / function requires specific parameters, right?

Jesus89:
There are lots of errors testing deep-insights.js: https://github.com/CartoDB/deep-insights.js/tree/layer-sources.
@alonsogarciapablo @IagoLast could you [take a look](https://travis-ci.org/CartoDB/deep-insights.js/builds/279445866?utm_source=email&utm_medium=notification)?

Jesus89:
Also 41 failures in Builder (`layer-sources` branch). I think we need to move it to development

Jesus89:
Deployed in ded10

Jesus89:
It works without errors!

PR #1760, opened 2017-09-19
Make the API backwards compatible

IagoLast:
- Add a static method in cartoDBLayer to get the source from the attrs.
    this method allows the source to be a Analysis node or an ID.
  - Make the getSourceId method common to all layer types, otherwise some
    errors will appear.
  - This commit does not include torque layers and should be fixed.


src/geo/map/cartodb-layer.js
```diff
@@ -137,10 +138,29 @@ var CartoDBLayer = LayerModelBase.extend({
...
+      attrs.source = CartoDBLayer.getLayerSourceFromAttrs(attrs, this._vis.analysis);
     }
     LayerModelBase.prototype.update.apply(this, arguments);
   }
 });
 
+/**
+  * Return the source analysis node from given attrs object.
+  */
+CartoDBLayer.getLayerSourceFromAttrs = function (attrs, analysis) {
```
alonsogarciapablo: FYI, Backbone gives you a way to define static methods/properties:

```
var MyModel = Backbone.Model.extend({
  initialize: function () { ... },
  ...
}, {
  myStaticProperty: 1,
  myStaticMethod: function () { ... }
  ...
});
```

More info: http://backbonejs.org/#Model-extend

src/geo/map/cartodb-layer.js
```diff
@@ -137,10 +138,29 @@ var CartoDBLayer = LayerModelBase.extend({
 
   update: function (attrs) {
     if (attrs.source) {
-      throw new Error('"source" must be set via setSource');
+      console.warn('Deprectaed: Use ".setSource" to update a layar\'s source instead the update method');
```
alonsogarciapablo: There're a couple of typos here...

src/geo/map/cartodb-layer.js
```diff
@@ -137,10 +138,29 @@ var CartoDBLayer = LayerModelBase.extend({
...
+    var source = analysis.findNodeById(attrs.source);
+    if (source) {
+      return source;
+    }
+    throw new Error('No analysis found with id: ' + attrs.source);
+  }
+  if (attrs.source instanceof AnalysisModel) {
+    return attrs.source;
+  }
+  throw new Error('Invalid layer source. Source must be an ID or a Analysis node but got: ' + attrs.source);
```
alonsogarciapablo: '... an Analysis...'

src/geo/map/layer-model-base.js
```diff
@@ -55,6 +55,12 @@ var MapLayer = Model.extend({
 
   setError: function (error) {
     this.set('error', error);
+  },
+
+  /**
+   * @Abstract method only needed in Carto and Torque Layers
+   */
+  getSourceId: function () {
```
alonsogarciapablo: I wouldn't personally add this here given that it's only for CartoDB and Torque layers. Perhaps those two should inherit from a new base class. Something like: CartoDBLayerModel -> DataModel -> LayerModelBase. What do you think?

test/spec/geo/map/cartodb-layer.spec.js
```diff
@@ -221,18 +221,12 @@ describe('geo/map/cartodb-layer', function () {
...
-    it('should raise an error when "source" attribute is specified', function () {
-      expect(function () {
-        layer.update({
-          id: 3,
-          source: 'foo'
-        });
-      }).toThrowError('"source" must be set via setSource');
+      expect(this.vis.analysis.findNodeById).not.toHaveBeenCalled();
+      layer.update({ id: 3, source: 'a0' });
+      expect(this.vis.analysis.findNodeById).toHaveBeenCalledWith('a0');
```
alonsogarciapablo: We're only testing that  `this.vis.analysis.findNodeById` was called... Why not ensure that `layer.getSource` returns the node with id `a0`?

src/geo/map/layer-model-base.js
IagoLast: I'm not a fan of inheritance and I think creating intermediate classes complicates the design a lot.

As pointed [here](https://github.com/CartoDB/cartodb.js/pull/1760/files#diff-f99ed960498d5ea8717b3e74d54bfb01R295) I see 2 options.

* Make the method "abstract" so all `layer.getSourceId()` wont fail: This is useful in case we got several loops assuming all the layers have a .getSourceId.

* Just check if the method exists: `layer.getSourceId && layer.getSourceId()`: This is useful if only used in a few places since we don't modify the class inheritance structure.

alonsogarciapablo: > Make the method "abstract" so all layer.getSourceId() wont fail: This is useful in case we got several loops assuming all the layers have a .getSourceId.

I'm not aware of any loop where we iterate over all layers and do something common for all of them, that's why I think we don't need this method.

> Just check if the method exists: layer.getSourceId && layer.getSourceId(): This is useful if only used in a few places since we don't modify the class inheritance structure.

Again, no need for that so far.

We can always add it to the base class in we case we need it... right?

src/vis/layers-factory.js
```diff
@@ -105,7 +96,7 @@ var LAYER_CONSTRUCTORS = {
 
   cartodb: function (attrs, options) {
     checkProperties(attrs, ['source', 'cartocss']);
-    checkValidSource(attrs);
+    attrs.source = CartoDBLayer.getLayerSourceFromAttrs(attrs, options.vis.analysis);
```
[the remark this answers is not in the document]
IagoLast: you mean accepting a string? I dont think it worth it since it´s a deprecated feature.

src/geo/map/cartodb-layer.js
```diff
@@ -136,6 +136,17 @@ var CartoDBLayer = LayerModelBase.extend({
...
 
+  /**
+   * Check if an analysis node is the layer's source.
+   * Only torque and cartodb layers have a source otherwise return false.
+   */
+  hasSource: function (analysisModel) {
+    if (!(analysisModel instanceof AnalysisModel)) {
+      throw new TypeError('CartodbLayer.hasSource must be called with an "AnalysisModel" but got: ' + analysisModel);
+    }
+    return this.getSourceId() === analysisModel.get('id');
```
alonsogarciapablo: Why not `return this.getSource() === analysisModel`? Another option could be to implement an `equals`  method on `AnalysisModel`:

```
return this.getSource().equals(analysisModel);
```

IagoLast: > Why not return this.getSource() === analysisModel?

Strict equal compares object references and this can lead to errors if for some reason the object is cloned and since `ids` must be unique I think that comparing them will do the job.

Should I change the code and use `this.getSource === analysisModel` ?

alonsogarciapablo: `return this.getSource().equals(analysisModel);` would be 👌 because we could probably reuse it in other places, but I leave it up to you at this point. PR is approved so these are only suggestions...

Issue #1761, opened 2017-09-20
Dataview models hold a reference to a source object instead of just the source id


alonsogarciapablo:
`DataviewModels` are currently storing the ID of an analysis node/model in their `source` attribute. Since we're moving towards an API that exposes objects that can be composed into other objects, `DataviewModels` should point to the actual `AnalysisModel`.

We're implementing a similar change for `CartoDBLayer` and `TorqueLayer` models in https://github.com/CartoDB/cartodb.js/pull/1755. Those classes expose a `getSource` and `setSource` methods to replace the source of the object. We could follow the same approach here...

The idea is to move towards something like this:

![image](https://user-images.githubusercontent.com/390398/30749296-39978cde-9fb3-11e7-8de5-a7ad006d35b0.png)

[This change](https://github.com/CartoDB/cartodb.js/pull/1755/commits/e29abd236adba4f40a27b05cb4aee00981daeff3
) in https://github.com/CartoDB/cartodb.js/pull/1755 is related to this, so let's make sure we take it into account to.

cc: @ivanmalagon

IagoLast:
### Acceptance

  🏳️  **LGTM**

Just loaded the following `viz.json` using the `examples/base.html` and compared the dataviews section in the maps-api response.



```json

{
  "bounds": [
    [
      38.994,
      -8.74622
    ],
    [
      42.3508,
      -1.86658
    ]
  ],
  "center": "[40.67241595,-5.306396485]",
  "datasource": {
    "user_name": "iago-carto",
    "maps_api_template": "https://{user}.carto.com:443",
    "stat_tag": "3174f83d-c574-42e9-bbdd-70c53c981a83",
    "template_name": "tpl_3174f83d_c574_42e9_bbdd_70c53c981a83"
  },
  "description": null,
  "options": {
    "dashboard_menu": true,
    "layer_selector": false,
    "legends": true,
    "scrollwheel": true
  },
  "id": "3174f83d-c574-42e9-bbdd-70c53c981a83",
  "layers": [
    {
      "id": "264a1297-409f-441e-a090-59bc0752434a",
      "type": "tiled",
      "options": {
        "default": "true",
        "url": "https://cartodb-basemaps-{s}.global.ssl.fastly.net/light_nolabels/{z}/{x}/{y}.png",
        "subdomains": "abcd",
        "minZoom": "0",
        "maxZoom": "18",
        "name": "Positron",
        "className": "positron_rainbow_labels",
[...]

PR #1771, opened 2017-09-21
Set dataview source as a AnalysisModel and get rid of layer

ivanmalagon:
Related to https://github.com/CartoDB/cartodb.js/issues/1761

This PR forces dataviews to hold a instance of AnalysisModel as source. Formerly, only an object with an id was needed.

Also, gets rid of layer object in dataviews.

Ready for acceptance.


src/dataviews/dataviews-factory.js
```diff
@@ -26,62 +26,55 @@ module.exports = Model.extend({
...
-    var categoryFilter = new CategoryFilter({
-      layer: layerModel
-    });
+    var categoryFilter = new CategoryFilter();
 
     return this._newModel(
       new CategoryDataviewModel(attrs, {
         map: this._map,
         vis: this._vis,
-        layer: layerModel,
```
alonsogarciapablo: We can also get rid of the `analysisCollection` option (and probably `vis` and `map` too) \o/.

Also, this factory probably doesn't need `analysisCollection` as a dependency.

src/dataviews/dataview-model-base.js
```diff
@@ -161,15 +160,14 @@ module.exports = Model.extend({
...
-      this._analysis.on('change:status', this._onAnalysisStatusChange, this);
+    if (this.getSource()) {
+      this.getSource().on('change:status', this._onAnalysisStatusChange, this);
     }
   },
 
   _removeExistingAnalysisBindings: function () {
-    if (!this._analysis) return;
-    this._analysis.off('change:status', this._onAnalysisStatusChange, this);
+    if (!this.get('source')) return;
```
alonsogarciapablo: I'd try to use `getSource` instead of accessing the `source` attribute as much as possible.

src/dataviews/dataviews-factory.js
ivanmalagon: `map` is used in dataviews for:
- Respond to changes like `view_bounds_ne`, `center`, `zoom`.
- Get the view bound.

`vis` is used in dataviews for:
- Reload the map (do an instantiation)

So i'm not sure we can eliminate those two properties unless you have something in mind that I'm not aware of, like `Engine` being responsible for triggering those changes to dataviews.

alonsogarciapablo: You're totally right! I misunderstood the diff. We can only get rid of `analysisCollection`.

ivanmalagon:
retest this please

src/dataviews/dataview-model-base.js
```diff
@@ -98,9 +99,21 @@ module.exports = Model.extend({
     this._setupAnalysisStatusEvents();
   },
 
+  set: function (attributes, options) {
```
alonsogarciapablo: The "public" way of updating a `DataviewModel` is by using `DataviewModel#update` (https://github.com/CartoDB/cartodb.js/blob/v4/src/dataviews/dataview-model-base.js#L258-L261).  I'd add this logic to that method instead of overriding `set`. Take into account that `update` only accepts an object, so we won't have to support "single attribute updates".

Also, for layers, we've introduced the `setSource` method so that users of the library can change the source more explicitly (https://github.com/CartoDB/cartodb.js/blob/layer-sources/src/geo/map/cartodb-layer.js#L128-L131). We should do something similar here.

src/dataviews/dataviews-factory.js
```diff
@@ -15,73 +15,59 @@ module.exports = Model.extend({
...
-    this._analysisCollection = opts.analysisCollection;
     this._dataviewsCollection = opts.dataviewsCollection;
-    this._analysisCollection = opts.analysisCollection;
   },
 
-  createCategoryModel: function (layerModel, attrs) {
-    _checkProperties(attrs, ['column']);
-    attrs = this._generateAttrsForDataview(layerModel, attrs, CategoryDataviewModel.ATTRS_NAMES);
+  createCategoryModel: function (attrs) {
+    _checkProperties(attrs, ['source', 'column']);
```
alonsogarciapablo: Should we ensure the `source` attrs is an instance of an `AnalysisModel` here? We're doing so in the `LayersFactory` [now](https://github.com/CartoDB/cartodb.js/blob/layer-sources/src/vis/layers-factory.js#L99).

src/analysis/analysis-factory.js
```diff
@@ -23,6 +23,8 @@ var AnalysisFactory = function (opts) {
  * may have one or more "source" params pointing to another node. If a node had been created
  * already, this method updates the attributes of the existing node. New nodes are added to
  * the collection of analyses that has been injected.
+ * 
+ * TODO: document what's the analysis definition
```
alonsogarciapablo: Should we?

src/util/required-opts.js
```diff
@@ -0,0 +1,12 @@
+var _ = require('underscore');
+
+module.exports = function checkAndBuildRequiredOpts (actualOpts, requiredOpts, context) {
+  if (requiredOpts === void 0) {
+    throw new Error('Opts are required');
+  }
+
+  _.each(requiredOpts, function (item) {
+    if (actualOpts[item] === void 0) throw new Error(item + ' is required');
+    context['_' + item] = actualOpts[item];
```
alonsogarciapablo: Do we need this?

test/helpers/fakeFactory.js
```diff
@@ -0,0 +1,24 @@
+var _ = require('underscore');
```
alonsogarciapablo: 👏  We should start using like this more frequently...

src/dataviews/dataview-model-base.js
ivanmalagon: +1 to add it to `update`.

I did this because in several specs we were modifying `source` via a simple `set`so I realised that's a backdoor, but I agree this can be a bit overkilling.

I'll left untouched the `set`function and will add the logic to `update`.

src/dataviews/dataviews-factory.js
ivanmalagon: I wouldn't because that check will be made by `dataview-model-base`, that it's called right here. Doing it in several places it's too much defensive, IMO.

src/analysis/analysis-factory.js
ivanmalagon: I missed it for the specs. Depends on what the public API will be. If we ask the users to provide an analysis definition we should.

src/util/required-opts.js
ivanmalagon: In `initialize` we use to do two things:
- Check that required options exist.
- Create a private property with each one of them.

This helpers does those two things in a single step. It's been very useful in Builder and we follow the same path in `cartodb.js` so I think it's convenient to keep it as it is.

test/helpers/fakeFactory.js
ivanmalagon: He´s phony... she's `fake`... that's the type of people I hate. https://www.youtube.com/watch?v=_T2yTDxJOdQ

alonsogarciapablo:
I'm resolving conflicts!

src/util/required-opts.js
```diff
@@ -0,0 +1,12 @@
+var _ = require('underscore');
+
+module.exports = function checkAndBuildRequiredOpts (actualOpts, requiredOpts, context) {
+  if (requiredOpts === void 0) {
```
IagoLast: since `void 0` is equivalent to `undefined` IMO we should use [isUndefined](http://underscorejs.org/#isUndefined)

```javascript 

if (_.isUndefined(requireOpts)){
    throw new Error('Opts are required');
}

File: src/util/required-opts.js [hunk 9]
```diff
@@ -0,0 +1,12 @@
...
+    throw new Error('Opts are required');
+  }
+
+  _.each(requiredOpts, function (item) {
+    if (actualOpts[item] === void 0) throw new Error(item + ' is required');
```
IagoLast: * I like to add curly braces to every conditional.
* [Same](https://github.com/CartoDB/cartodb.js/pull/1771/files#r141307949)

src/util/required-opts.js
ivanmalagon: I'd bet that line will always say `Opts are required but received: undefined`.

src/util/required-opts.js
```diff
@@ -0,0 +1,12 @@
+var _ = require('underscore');
+
+module.exports = function checkAndBuildRequiredOpts (actualOpts, requiredOpts, context) {
```
IagoLast: * We already had an [util](https://github.com/CartoDB/cartodb.js/blob/v4-next/src/core/util.js) (by the way with outdated methods should we merge them?

* Should we use this check in the other Models? (ie: [CartoDBLayer](https://github.com/CartoDB/cartodb.js/blob/v4-next/src/geo/map/cartodb-layer.js#L21) )

test/helpers/fakeFactory.js
IagoLast: yep, imo we should have a .spec.js and a .mock.js for every file :)

btw: I like to use [sinon](http://sinonjs.org/) for the spies/mocks it´s a nice tool!

test/helpers/fakeFactory.js
```diff
@@ -0,0 +1,24 @@
...
+var AnalysisModel = require('../../src/analysis/analysis-model');
+var camshaftReference = require('../../src/analysis/camshaft-reference');
+
+var createAnalysisModel = function (attrs) {
+  if (typeof attrs === 'string') throw new Error('BOOM!');
```
IagoLast: boom?

ivanmalagon: Oops... that shouldn't be there.

src/util/required-opts.js
ivanmalagon: We can move it to `util`, yes.

My advice is to use wherever we can. It eases the `initialize` checks.

src/dataviews/dataview-model-base.js
IagoLast: I dont understand this method :S

ivanmalagon: It's overriding Backbone's `set` to check that `source` is an AnalysisModel in case somebody takes the backdoor of using `set` after the widget is created. As talked with Pablo, this makes more sense in `update` function.

src/dataviews/dataview-model-base.js
```diff
@@ -161,15 +174,14 @@ module.exports = Model.extend({
 
   _setupAnalysisStatusEvents: function () {
     this._removeExistingAnalysisBindings();
-    this._analysis = this.getSource();
-    if (this._analysis) {
-      this._analysis.on('change:status', this._onAnalysisStatusChange, this);
+    if (this.getSource()) {
```
IagoLast: any special reason for not using `this.has('source')`? looks more descriptive to me

src/dataviews/dataviews-factory.js
```diff
@@ -15,73 +15,59 @@ module.exports = Model.extend({
...
     if (!opts.map) throw new Error('map is required');
     if (!opts.vis) throw new Error('vis is required');
-    if (!opts.analysisCollection) throw new Error('analysisCollection is required');
     if (!opts.dataviewsCollection) throw new Error('dataviewsCollection is required');
-    if (!opts.analysisCollection) throw new Error('analysisCollection is required');
 
     this._map = opts.map;
     this._vis = opts.vis;
-    this._analysisCollection = opts.analysisCollection;
     this._dataviewsCollection = opts.dataviewsCollection;
```
IagoLast: @alonsogarciapablo is this being used?

src/dataviews/dataviews-factory.js
```diff
@@ -15,73 +15,59 @@ module.exports = Model.extend({
...
+        filter: rangeFilter
       })
     );
   },
 
-  _generateAttrsForDataview: function (layerModel, attrs, whitelistedAttrs) {
+  _generateAttrsForDataview: function (attrs, whitelistedAttrs) {
     attrs = _.pick(attrs, whitelistedAttrs);
-    attrs.source = attrs.source || { id: layerModel.id };
+    attrs.source = attrs.source;
```
IagoLast: is this line needed?

src/dataviews/dataview-model-base.js
alonsogarciapablo: I think we should avoid working with attributes as much as possible so that we can change things easier in the future (eg: if we decide `source` is no longer an attribute). If `getSource` is not descriptive enough we can create a `hasSource` method...

Anyway, do we need to check if there's a source? Can't we assume there'll always be one?

IagoLast: > do we need to check if there's a source?

IMO is not necessary, but we should update the setters to prevent an empty/invalid source.

thoughts?

src/dataviews/dataviews-factory.js
alonsogarciapablo: :+1:

src/dataviews/dataviews-factory.js
ivanmalagon: 🤦‍♂️

src/core/util.js
```diff
@@ -147,4 +149,18 @@ util.endsWith = function (str, suffix) {
...
+    }
+  });
+};
+
+util.setAsPrivateProperties = function (opts, context) {
+  _.each(_.keys(opts), function (item) {
+    context['_' + item] = opts[item];
+  });
+};
+
```
ivanmalagon: ^^^
I've divided `checkAndBuildOptions` in two functions: one to check that all required options are present and another one to set them as private properties.

I've moved them to `util` as well. You can check how they're used at `dataviews-factory`, for instance.

src/dataviews/dataview-model-base.js
```diff
@@ -71,21 +78,16 @@ module.exports = Model.extend({
   initialize: function (attrs, opts) {
     attrs = attrs || {};
     opts = opts || {};
+    util.checkRequiredOpts(opts, REQUIRED_OPTS, 'DataviewModelBase');
+    util.setAsPrivateProperties(opts, this);
```
alonsogarciapablo: I still don't buy it... IMO, this makes it hard to know what are the "private" properties of this class unless you actually go look for them in the actual code... I'd rather be explicit and make sure we're only declaring the variables that are indeed being used...

ivanmalagon: OK. If you feel uncomfortable we could leave it only with `checkRequiredOpts` and then assign them manually.

But I must say that you concern is only caused by the novelty of seeing it. In practice, we haven't suffered any issues in Builder with this.

ivanmalagon: Reverted.

src/dataviews/dataview-model-base.js
```diff
@@ -71,21 +78,18 @@ module.exports = Model.extend({
   initialize: function (attrs, opts) {
     attrs = attrs || {};
     opts = opts || {};
+    util.checkRequiredOpts(opts, REQUIRED_OPTS, 'DataviewModelBase');
+
+    this._map = opts.map;
+    this._vis = opts.vis;
```
alonsogarciapablo: @ivanmalagon don't you think this makes it easier to know what are the "private" instance variables of this class (plus it only declares the variables we explicitly "extract" from `opts`?

ivanmalagon: In this case, you get used to it.

What I find worse is the lack of knowing what the attributes are. We don't tend to set the `attributes` property in Backbone classes and then knowing the different attributes is a mistery. But this is thing of how we use Backbone through all repos (and it has nothing to to with this PR).

Issue #1764, opened 2017-09-20
JSDoc: implement mechanism to separate internal docs from public facing docs


alonsogarciapablo:
We should support different levels of privacy and more specifically we need to have a way to specify which classes / methods / functions are part of the public facing API, which ones are "public" and which ones are "private". OpenLayers have solved this by defining their own @api tag (eg: https://github.com/openlayers/openlayers/blob/master/src/ol/map.js#L87).

cc: @Jesus89

Jesus89:
I think we could use this @api tag, so we could have:
```
npm run docs
npm run docs:all
```

PR #1772, opened 2017-09-21
Use @api tag for docs

Jesus89:
Related to: #1764

Now there are two commands to generate docs:
- `npm run docs`  (tagged with @api)
- `npm run docs:all` (except @ignored)

Also, Travis CI deploys to gh-pages the Public docs in "docs/v4" and All docs in "docs/v4-internal".

http://cartodb.github.io/cartodb.js/


.travis.yml
```diff
@@ -18,13 +18,4 @@ notifications:
...
-
-#deploy:
-#  provider: pages
-#  skip_cleanup: true
-#  local_dir: out
-#  on:
-#    branch: jsdoc-ghpages
-#  github-token:
-#    secure: ...
+    secure: lcdjTpuv2imFSgcYWMLV4odyZ4jOTjJhyViGg9hOFzggoPGFAjplz6bPgNJsxq20ZsFDw4UN6UlZ8lPGlByIUBhf2OXs0cEJDcNSoYAV7CuTMSdZsLDb+TrWQvjtSZMBEkeVENfJjyfiJo7PCt2/cIQXSVSnhZ5Uiir+Eus5x/s=
```
IagoLast: What about [this](https://docs.travis-ci.com/user/environment-variables/#Defining-Variables-in-Repository-Settings) instead using encrypted vars?

doc/index.html
```diff
@@ -0,0 +1,16 @@
...
+<html lang="en">
+<head>
+  <meta charset="utf-8">
+  <title>CARTO.js</title>
+  <link type="text/css" rel="stylesheet" href="docs/v4/styles/jsdoc-default.css">
+</head>
+<body>
+  <h1 class="page-title">CARTO.js <span style="color:#0095dd;">v4</span></h1>
+  <ul style="margin:24px;">
+    <li><h3><a href="docs/v4">API docs</a></h3></li>
```
IagoLast: should this be doc/v4?

.travis.yml
Jesus89: It can be done this way, but for this purpose the encrypted tokens are perfectly fine

doc/index.html
Jesus89: This file is the main index.html in gh-pages: https://github.com/CartoDB/cartodb.js/blob/gh-pages/index.html. It is copied with the "docs" directory. I think we could move it to another directory, may be "config/docs"

Jesus89:
In gh-pages deploy out is moved to docs. I can generate always the "docs" directory, but this will be git-ignored.

IagoLast:
should we change the .gitignore? :D

config/jsdoc/plugins/api.js
```diff
@@ -0,0 +1,148 @@
+/**
```
alonsogarciapablo: It looks like this file still contains some stuff that is specific to OpenLayers (`olx.js`)... It'd be nice to cleanup a bit so and fit it to what we need. Although it doesn't look something very straight forward.

Jesus89: Super cleanup done!

alonsogarciapablo: 🔝 !

PR #1778, opened 2017-09-25
Improve widgets errors

rubenmoya:
Related to: https://github.com/CartoDB/deep-insights.js/issues/574

This PR triggers a custom error when the status code is `429`, this is used in `deep-insights` to handle the widget error view.


src/dataviews/dataview-model-base.js
```diff
@@ -117,6 +117,15 @@ module.exports = Model.extend({
     if (this.filter) {
       this.listenTo(this.filter, 'change', this._onFilterChanged);
     }
+
+    this.bind('error', function (model, response, options) {
+      if (response.status === 429) {
+        var error = response.responseJSON.errors_with_context[0];
+        setTimeout(function () {
```
nobuti: why we need to defer this?

rubenmoya: Because otherwise, the `error` event arrives after `error:_custom_` and could mess up the `listenTo` overriding the specific error with a generic one.

nobuti: It's kind a hack, and maybe it indicates a code smell. Is there any way to avoid this situation?

rubenmoya: The only way to avoid it would be not to trigger an error inside another error... or think about another way to communicate with Builder without triggering errors...

xavijam: A stack of errors?

xavijam:
I have a crazy idea, but let me know what you think @nobuti @rubenmoya. 
So, seems like the problem is we are triggering two error events (one custom and another standard) from the same model. So what about if we override the error "callback" or place where it is used and we send only one event, but depending the error, we send `error:whatever` or `error`.

nobuti:
It could work, but not sure if I'd focus on adding some kind of a type like the error requests...

xavijam:
Tell me more...

xavijam:
Yes! That should be the idea!

xavijam:
I've tried in staging again, but the histogram is not updated when the map is moved.

Issue #1763, opened 2017-09-20
Extract classes to generate Maps API payload 


alonsogarciapablo:
Part of https://github.com/CartoDB/cartodb.js/issues/1762.

Right now we have the following hierarchy of classes:

![untitled drawing](https://user-images.githubusercontent.com/390398/30636833-d5f32ff2-9df6-11e7-981b-4af5366fa23f.jpg)

Those classes have at least two very clear responsibilities:

1. They hold a reference to the collections of layers, dataviews and analysis and expose a method (`createInstance`) that:
    - Uses `toJSON` (more on this in 2.)
    - Knows how to communicate with Maps API (via `WindshaftClient`)
    - Uses an instance of `ModelUpdater` to update many internal models with info from the response from Maps API.

2. Generate the payload that gets sent to the Maps API from the collection of layers, dataviews and analysis ( this is done by the `toJSON` method).

This issue is about extracting this second responsibility into a separate class and use composition instead of inheritance.

PR #1779, opened 2017-09-25
Extract map serializer

IagoLast:
Fixes https://github.com/CartoDB/cartodb.js/issues/1763 (which is part of https://github.com/CartoDB/cartodb.js/issues/1763).


test/spec/windshaft/map-serializer/anonymous-map-serializer/layers-serializer.spec.js
```diff
@@ -0,0 +1,116 @@
...
+var CartoDBLayer = require('../../../../../src/geo/map/cartodb-layer');
+var LayersSerializer = require('../../../../../src/windshaft/map-serializer/anonymous-map-serializer/layers-serializer');
+var PlainLayer = require('../../../../../src/geo/map/plain-layer');
+var TileLayer = require('../../../../../src/geo/map/tile-layer');
+var TorqueLayer = require('../../../../../src/geo/map/torque-layer');
+var VisModel = require('../../../../../src/vis/vis');
+
+describe('layers-serializer', function () {
+  describe('.serialize', function () {
+    var visMock;
```
alonsogarciapablo: We tend to use `this.foo` inside the `beforeEach` instead of using variables. For example:
https://github.com/CartoDB/cartodb.js/pull/1779/files#diff-6c3f2d03eeffeb930a0da94cbd134a5fR11
I don't really prefer one over the other (perhaps using vars makes our lives simpler...), but we should keep it consistent.

@IagoLast It seem you prefer `var` over `this.`, right? Why?

package.json
```diff
@@ -48,6 +48,10 @@
     "babelify": "~7.3.0",
     "browserify": "13.0.0",
     "cartoassets": "CartoDB/CartoAssets#master",
+    "eslint-config-standard": "^10.2.1",
```
alonsogarciapablo: `eslint-config-semistandard`?

IagoLast: They are peer-dependencies of [eslint-confi-semistandard](https://github.com/Flet/eslint-config-semistandard#install)

test/spec/windshaft/map-serializer/anonymous-map-serializer/anonymous-map-serializer.spec.js
```diff
@@ -0,0 +1,112 @@
+var Backbone = require('backbone');
```
alonsogarciapablo: @iago I added this very basic tests to ensure anonymous-map-serializer is doing the right thing.

test/spec/windshaft/map-serializer/anonymous-map-serializer/anonymous-map-serializer.spec.js
```diff
@@ -0,0 +1,120 @@
...
+var DataviewModelBase = require('../../../../../src/dataviews/dataview-model-base');
+var AnonymousMapSerializer = require('../../../../../src/windshaft/map-serializer/anonymous-map-serializer/anonymous-map-serializer');
+
+var MyDataviewModel = DataviewModelBase.extend({
+  toJSON: function () {
+    return {};
+  }
+});
+
+fdescribe('anonymous-map-serializer', function () {
```
IagoLast: `fdescribe` should be `describe`

test/spec/windshaft/map-serializer/anonymous-map-serializer/layers-serializer.spec.js
IagoLast: when using `this.foo` the linter cannot determinate if the variable is declared or unused plus, using `this`we must handle the context manually (ie: inside a loop) so IMO using var is less error prone.

PR #1780, opened 2017-09-26
Filter log, to avoid treating "Unsupported CartoCSS" as errors

dv343:
We want to avoid logging "Unsupported CartoCSS" errors, since they are not really errors.


src/geo/leaflet/leaflet-layer-view-factory.js
```diff
@@ -62,7 +62,11 @@ var canLayerBeRenderedClientSide = function (layerModel) {
   var cartoCSS = layerModel.get('meta').cartocss;
   var result = TC.getSupportedCartoCSSResult(cartoCSS);
   if (!result.supported) {
-    log.error(new Error('[Vector] Unable to render due "' + result.reason + '". Full CartoCSS:\n' + cartoCSS));
+    if (result.reason.indexOf('Unsupported CartoCSS') === 0) {
```
IagoLast: I would create a separated function for this, but it´s ok

Issue #1787, opened 2017-09-28
Clean Model updater


IagoLast:
Instead keeping a list with all the created analyses, the modelUpdater should extract all the analyses from the layers and dataviews collections updating each one.

IagoLast:
Following the discussion, since looking for all the analysis and (removing dupes and nested ones) is a common behaviour I think we should encapsulate it in a shared method.

Maybe a static method in the `analisisFactory`?

```javascript
analysisFactory.prototype.getAnalyses = function(layersCollection, dataviewsCollection){
 // ...
}
```

PR #1795, opened 2017-10-02
1787 clean model updater

IagoLast:
https://github.com/CartoDB/cartodb.js/issues/1787 WIP
----


src/windshaft-integration/model-updater.js
```diff
@@ -266,4 +268,27 @@ ModelUpdater.prototype._getLayerLegends = function (layerModel) {
     layerModel.legends.choropleth
   ];
 };
+
+/**
+ * Return the analysis list without duplicates
+ * 
+ * (a dataview and a layer could share some analysis)
+ */
+ModelUpdater.prototype._getUniqueAnalyses = function () {
```
alonsogarciapablo: Perhaps we should differentiate between the terms "Analysis" and "Analysis node" in the code (we're already doing it in the body of this method). Since this method is returning "nodes", perhaps it should reflect that in its name, so I'd rename it to something like `_getAnalysisNodes`.

src/analysis/analysis-service.js
```diff
@@ -74,21 +75,52 @@ AnalysisFactory.prototype._getAnalysisAttributesFromAnalysisDefinition = functio
...
+    if (!layer.getSource()) {
+      throw new TypeError('CartoDB and Torque layers must have a source');
+    }
+    return layer.getSource();
+  });
+}
+
+function _getAnalysesFromDataviews (dataviewsCollection) {
+  return dataviewsCollection.map(function (dataview) {
+    return dataview.getSource();
```
Jesus89: I think we need to check if getSource is null here too

Jesus89: cc @alonsogarciapablo

alonsogarciapablo: Or we could assume both layers and dataviews will always have a source [here](https://github.com/CartoDB/cartodb.js/pull/1795/files#diff-b4b64c4fcc702683c1df5ba40b85006bR107).

src/analysis/analysis-service.js
```diff
@@ -74,21 +75,52 @@ AnalysisFactory.prototype._getAnalysisAttributesFromAnalysisDefinition = functio
...
+  var layerAnalyses = _getAnalysesFromLayers(layersCollection);
+  var dataviewsAnalyses = _getAnalysesFromDataviews(dataviewsCollection);
+  return layerAnalyses.concat(dataviewsAnalyses);
+};
+
+function _getAnalysesFromLayers (layersCollection) {
+  var layers = _getCartoDBAndTorqueLayers(layersCollection);
+  return layers.map(function (layer) {
+    if (!layer.getSource()) {
+      throw new TypeError('CartoDB and Torque layers must have a source');
```
alonsogarciapablo: I wouldn't consider this a `TypeError`... Anyway, perhaps we could get rid of this check (see comment below).

src/vis/layers-factory.js
```diff
@@ -95,17 +95,31 @@ var LAYER_CONSTRUCTORS = {
...
-    checkProperties(attrs, ['source', 'cartocss']);
-    attrs.source = TorqueLayer.getLayerSourceFromAttrs(attrs, options.vis.analysis);
+    // TODO: Once https://github.com/CartoDB/cartodb/issues/12885 is merged,
+    // we should make 'source' attribute required again and make sure it's
+    // populated:
+    // checkProperties(attrs, ['source', 'cartocss']);
+    // attrs.source = CartoDBLayer.getLayerSourceFromAttrs(attrs, options.vis.analysis);
+    checkProperties(attrs, ['cartocss']);
+    if (attrs.source) {
+      attrs.source = CartoDBLayer.getLayerSourceFromAttrs(attrs, options.vis.analysis);
```
alonsogarciapablo: We're using `TorqueLayer.getLayerSourceFromAttrs`...

`CartoDBLayer.getLayerSourceFromAttrs` and `TorqueLayer.getLayerSourceFromAttrs` are identical so perhaps we could unify them somewhere. Perhaps they could live in the `AnalysisService` and here we could just do something like `options.vis.analysis.getSourceOrFindById(attrs.source)`.

Since this epic will be merged after `sources-as-references`, perhaps this is a good opportunity to get rid of the deprecation warnings in `getLayerSourceFromAttrs`.

src/windshaft-integration/model-updater.js
```diff
@@ -266,4 +268,22 @@ ModelUpdater.prototype._getLayerLegends = function (layerModel) {
     layerModel.legends.choropleth
   ];
 };
+
+/**
+ * Return the analysis list without duplicates
+ *
+ * (a dataview and a layer could share some analysis)
+ */
+ModelUpdater.prototype._getUniqueAnalyses = function () {
```
alonsogarciapablo: I'd include "nodes" in the name of this method somehow so that it's clear that it's not just returning "root" analyses but it's "flattening" the analysis trees instead.

IagoLast: 👍

Jesus89:
Tests must be fixed:

X vis/layers-factory attribute validation CartoDB should throw an error if no properties are given
   Expected function to throw an exception with message 'The following attributes are missing: cartocss', but it threw an exception with message 'The following attributes are missing: cartocss,source'.

X vis/layers-factory attribute validation torque should throw an error if no properties are given
   Expected function to throw an exception with message 'The following attributes are missing: cartocss', but it threw an exception with message 'The following attributes are missing: cartocss,source'.

test/spec/api/createVis/create-vis.spec.js
```diff
@@ -213,13 +213,13 @@ describe('create-vis:', function () {
       expect(visModel._windshaftMap.get('statTag')).toEqual(visJson.datasource.stat_tag);
     });
 
-    it('should have the _analysisCollection linked', function () {
+    xit('should have the _analysisCollection linked', function () {
```
alonsogarciapablo: What about these tests?

src/vis/layers-factory.js
```diff
@@ -95,12 +95,7 @@ var LAYER_CONSTRUCTORS = {
   },
 
   cartodb: function (attrs, options) {
-    // TODO: Once https://github.com/CartoDB/cartodb/issues/12885 is merged,
-    // we should make 'source' attribute required again and make sure it's
-    // always populated:
-    // checkProperties(attrs, ['source', 'cartocss']);
-    // attrs.source = CartoDBLayer.getLayerSourceFromAttrs(attrs, options.vis.analysis);
-    checkProperties(attrs, ['cartocss']);
+    checkProperties(attrs, ['cartocss', 'source']);
```
alonsogarciapablo: https://github.com/CartoDB/cartodb/issues/12885 is not closed yet so we cannot make "source" required.

src/vis/layers-factory.js
```diff
@@ -111,14 +106,9 @@ var LAYER_CONSTRUCTORS = {
   },
 
   torque: function (attrs, options) {
-    // TODO: Once https://github.com/CartoDB/cartodb/issues/12885 is merged,
-    // we should make 'source' attribute required again and make sure it's
-    // populated:
-    // checkProperties(attrs, ['source', 'cartocss']);
-    // attrs.source = CartoDBLayer.getLayerSourceFromAttrs(attrs, options.vis.analysis);
-    checkProperties(attrs, ['cartocss']);
+    checkProperties(attrs, ['cartocss', 'source']);
```
alonsogarciapablo: https://github.com/CartoDB/cartodb/issues/12885 is not closed yet so we cannot make "source" required.

PR #1797, opened 2017-10-02
Use eslint config

Jesus89:
Use `eslint-config-semistandard` (+deps) instead `semistandard` package.

This pull request removes some npm warnings during installation.

Also `jshint` deps have been removed.


IagoLast:
We decided to include eslint as a explicit dev-dependency

PR #1801, opened 2017-10-03
Add tracking code

ivanmalagon:
Temporary code to track in real usage if Dataviews' source is changed often (if ever) and in what circumstances.

### DIFF
--- a/src/dataviews/dataview-model-base.js
+++ b/src/dataviews/dataview-model-base.js
@@ -107,7 +107,7 @@ module.exports = Model.extend({
   _initBinds: function () {
     this.listenTo(this.layer, 'change:visible', this._onLayerVisibilityChanged);
     this.listenTo(this.layer, 'change:source', this._setupAnalysisStatusEvents);
-    // this.on('change:source', this._setupAnalysisStatusEvents, this);
+    this.on('change:source', this._setupAnalysisStatusEvents, this);
 
     // Temporary code to log changes to dataview's sources
     // TODO: to be removed when enough data is checked / 1761 gets merged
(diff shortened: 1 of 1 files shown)


ivanmalagon:
retest this please

src/dataviews/dataview-model-base.js
```diff
@@ -100,8 +106,8 @@ module.exports = Model.extend({
 
   _initBinds: function () {
     this.listenTo(this.layer, 'change:visible', this._onLayerVisibilityChanged);
-    this.listenTo(this.layer, 'change:source', this._setupAnalysisStatusEvents);
-    this.on('change:source', this._setupAnalysisStatusEvents, this);
+    this.listenTo(this.layer, 'change:source', this._onLayerSourceChanged);
+    this.on('change:source', this._onSourceChanged, this);
```
alonsogarciapablo: 💡  We could also leave existing bindings as they were and add another one with a comment explaining why we're adding this:
```
// Temporary code to log changes to dataview's sources.
// TODO: To be removed after xxx is merged.
this.on('change:source', this._logSourceChange, this);
```

src/dataviews/dataview-model-base.js
```diff
@@ -256,6 +274,10 @@ module.exports = Model.extend({
   },
 
   update: function (attrs) {
+    if (_.has(attrs, 'source')) {
+      var message = '[SOURCE] Source present in UPDATE attrs. ' + JSON.stringify(attrs.source);
```
alonsogarciapablo: I'm sure you're aware, but just in case... We'll get two error messages in this case: one when `update` is invoked, and another one when `change:source` will be triggered. Nothing bad with it...

src/dataviews/dataview-model-base.js
ivanmalagon: :+1:

src/dataviews/dataview-model-base.js
ivanmalagon: I know, but we can receive the `update` message and not the `change` one if the source is the same.

alonsogarciapablo: :+1:

PR #1803, opened 2017-10-03
Fix gmaps tests

Jesus89:
Related to: #1798

The issue is that gmaps projection in the projector is not ready when `latLngToPixel`and `pixelToLatLng`methods are called and they return always (0, 0) failing some tests.

More precisely: in the polygons edition tests (6 tests) the middle points are not added in the coordinates because the `computeMidLatLng` method calls `mapView.latLngToContainerPoint` that returns (0, 0) when the projection is not ready. So no middle points in the coordinates list, no tests passed.

It depends on the platform. In Chrome when all the tests are executed they pass. If you run some tests individually they fail. In Firefox the 6 tests fail when all tests are executed, as in Jenkins.

Changes to solve the issue:
- Refactor of Projector. Remove _ready method trick, use idle event instead
- Added 'Projector has no projection' warning when projection is not loaded
- Wait async for the map to be loaded in the failing tests

### DIFF
--- a/src/geo/gmaps/gmaps-map-view.js
+++ b/src/geo/gmaps/gmaps-map-view.js
@@ -82,6 +82,12 @@ var GoogleMapsMapView = MapView.extend({
     MapView.prototype.clean.call(this);
   },
 
+  listenOnce: function (name, callback) {
+    google.maps.event.addListenerOnce(this._gmapsMap, name, function (event) {
+      callback(event);
+    });
+  },
+
   _getLayerViewFactory: function () {
     this._layerViewFactory = this._layerViewFactory || new GMapsLayerViewFactory();
 
--- a/src/geo/gmaps/gmaps-map-view.js
+++ b/src/geo/gmaps/gmaps-map-view.js
@@ -1,12 +1,10 @@
 /* global google */
-var _ = require('underscore');
 var MapView = require('../map-view');
 var Projector = require('./projector');
 var GMapsLayerViewFactory = require('./gmaps-layer-view-factory');
 
 var GoogleMapsMapView = MapView.extend({
   initialize: function () {
-    _.bindAll(this, '_ready');
     this._isReady = false;
 
     MapView.prototype.initialize.apply(this, arguments);
@@ -70,9 +68,11 @@ var GoogleMapsMapView = MapView.extend({
       self.map.setMapViewSize(self.getSize());
     });
 
-    this.projector = new Projector(this._gmapsMap);
+    google.maps.event.addListenerOnce(this._gmapsMap, 'idle', function (e) {
+      this._isReady = true;
+    });
 
-    this.projector.draw = this._ready;
+    this.projector = new Projector(this._gmapsMap);
   },
 
   clean: function () {
@@ -86,12 +86,6 @@ var GoogleMapsMapView = MapView.extend({
     this._layerViewFactory = this._layerViewFactory || new GMapsLayerViewFactory();
 
     return this._layerViewFactory;
-  },
-
-  _ready: function () {
-    this.projector.draw = function () {};
-    this.trigger('ready');
-    this._isReady = true;
   },
 
   _setKeyboard: function (model, z) {
--- a/src/geo/gmaps/projector.js
+++ b/src/geo/gmaps/projector.js
@@ -5,7 +5,7 @@ function Projector (map) {
   if (!this._projection) {
     google.maps.event.addListenerOnce(map, 'projection_changed', function () {
       this._projection = map.getProjection();
-    });
+    }.bind(this));
   }
 }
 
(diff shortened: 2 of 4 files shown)


src/geo/gmaps/projector.js
```diff
@@ -1,21 +1,30 @@
...
-Projector.prototype.draw = function () {};
-Projector.prototype.latLngToPixel = function (point) {
-  var p = this.getProjection();
-  if (p) {
-    return p.fromLatLngToContainerPixel(point);
+
+function Projector (map) {
+  this._projection = map.getProjection();
+  if (!this._projection) {
+    google.maps.event.addListenerOnce(map, 'projection_changed', function () {
```
IagoLast: We updated this code adding a warning here

src/geo/gmaps/projector.js
```diff
@@ -1,21 +1,30 @@
...
+      this._projection = map.getProjection();
+    });
+  }
+}
+
+Projector.prototype.latLngToPixel = function (latlng) {
```
IagoLast: Transform coordinates into pixels without knowing the projection has no sense so I wont return anything is there is no projection.

```javascript
Projector.prototype.latLngToPixel = function (latlng) {
  return this._projection.fromLatLngToPoint(latlng);
};
```

This will throw a better error instead `cannot read property fromLatLngToPoint from undefined`
```javascript
Projector.prototype.latLngToPixel = function (latlng) {
  if(!this._projection) {
    throw new Error('Projector.latLngToPixel needs a projection to work'); 
  }
  return this._projection.fromLatLngToPoint(latlng);
};
```

src/geo/gmaps/projector.js
```diff
@@ -1,21 +1,30 @@
...
+    return this._projection.fromLatLngToPoint(latlng);
   }
-  return new google.maps.Point(0, 0);
+  // FIXME
+  console.warn('Projector has no projection');
+  return new google.maps.Point(latlng.lat(), latlng.lng());
 };
+
 Projector.prototype.pixelToLatLng = function (point) {
-  var p = this.getProjection();
```
IagoLast: same here

IagoLast:
@alonsogarciapablo can you define acceptance tests for this?

is there any way of testing the `polygon-drawing with google maps`, without using builder?

src/geo/gmaps/projector.js
Jesus89: @IagoLast the issue here is that the projection is not defined when the tests are executed. I have checked that tests are failing with your approach, so if we don't return the same point if there is no projection we must update the failing tests

IagoLast:
Aceptance 
------


Use the Example `gmaps-geometry-editor` and test with different geometries.

src/geo/gmaps/gmaps-map-view.js
```diff
@@ -82,6 +82,12 @@ var GoogleMapsMapView = MapView.extend({
     MapView.prototype.clean.call(this);
   },
 
+  listenOnce: function (name, callback) {
+    google.maps.event.addListenerOnce(this._gmapsMap, name, function (event) {
```
IagoLast: Remove uneccessary function
```javascript
  google.maps.event.addListenerOnce(this._gmapsMap, name, callback);
```

src/geo/leaflet/leaflet-map-view.js
```diff
@@ -194,6 +194,10 @@ var LeafletMapView = MapView.extend({
     }, this);
   },
 
+  listenOnce: function (name, callback) {
```
IagoLast: what is the purpose of this?

src/geo/gmaps/gmaps-map-view.js
Jesus89: This function is used in the tests to wait after map's idle and ensure that projection exists

src/geo/leaflet/leaflet-map-view.js
Jesus89: This method is implemented in gmaps. This is the equivalent leaflet method to pass the tests

src/geo/gmaps/gmaps-map-view.js
alonsogarciapablo: I think @IagoLast refers to the callback fn. There's no need to define an anonymous fn and you can pass `callback` instead.

src/geo/leaflet/leaflet-map-view.js
alonsogarciapablo: That's a bit weird (having to add a method just for testing purposes).

test/spec/geo/geometry-views/shared-tests-for-path-views.js
```diff
@@ -161,12 +161,17 @@ module.exports = function (Path, MapView, PathView) {
         mapView: this.mapView
       });
 
-      this.geometryView.render();
+      // Listen for the map to be loaded
+      this.mapView.listenOnce('idle', function () {
```
alonsogarciapablo: Not very elegant but perhaps we can check if `this.mapView` is an instance of `src/geo/gmaps/gmaps-map-view.js` and wait to the `idle` event only in that case? So that we can get rid of `listenOnce` which was only added for testing purposes...

Issue #1788, opened 2017-09-28
Decouple Analysis Polling from AnalysisCollection


IagoLast:
When a new instance of the map is created, CARTO.js resets analysis polling and restarts polling again for each of the analysis nodes present in the collection of analyses ([here](https://github.com/CartoDB/cartodb.js/blob/0daa66a19d9e61f0c1f2aabb71210d0d6b4f6ad2/src/vis/vis.js#L272-L282)). Since we're removing that internal collection, we should revisit how this works (i.e: not depend on the collection anymore).

Also, CARTO.js could be a bit smarter about polling (i.e: do not start polling if node depends on nodes that are not ready).

PR #1805, opened 2017-10-04
[WIP] Decouple Analysis Polling from AnalysisCollection

alonsogarciapablo:
Fixes https://github.com/CartoDB/cartodb.js/issues/1788.

Remaining work:

- [ ] Do not use analysisCollection when reseting models in the analysisPoller (extract sources from layers and dataviews instead).
- [ ] No need to initialise analysisCollection in `vis#load`
- [ ] Ensure sources are marked/unmarked as referenced when creating / updating / deleting layers and dataviews.

cc: @IagoLast


src/vis/vis.js
```diff
@@ -262,34 +259,20 @@ var VisModel = Backbone.Model.extend({
...
 
-  _onAnalysisStatusChanged: function (analysisModel) {
-    if (analysisModel.isDone()) {
-      this.untrackLoadingObject(analysisModel);
-      if (this._isAnalysisSourceOfLayerOrDataview(analysisModel)) {
-        this.reload();
-      }
-    }
+  _getAnalysisNodeModels: function () {
+    return this._analysisCollection.models;
```
alonsogarciapablo: @IagoLast instead of using `this._analysisCollection` here, we need to extract sources from `this._layersCollection` and `this._dataviewsCollection`.

src/analysis/analysis-model.js
```diff
@@ -25,6 +25,8 @@ var AnalysisModel = Model.extend({
     this._vis = opts.vis;
 
     this._initBinds();
+
+    this._referencedBy = {};
```
IagoLast: A little comment pointing the intention would be helpful:

`// Store every layer/dataview pointing to this analysis.`

src/analysis/analysis-model.js
```diff
@@ -68,6 +70,16 @@ var AnalysisModel = Model.extend({
     _.each(this.getParamNames(), function (paramName) {
       this.bind('change:' + paramName, this._reloadVis, this);
     }, this);
+
+    this.bind('change:status', function () {
+      if (this._hadStatus() && this.isReady() && this.isSourceOfAnyModel()) {
```
IagoLast: I suppose _hadStatus checks for a previous status, can an analysis start in "ready" status?

src/analysis/analysis-model.js
```diff
@@ -68,6 +70,16 @@ var AnalysisModel = Model.extend({
...
+        this._reloadVis();
+      }
+    }, this);
+  },
+
+  _hadStatus: function () {
```
IagoLast: `_hasPreviousStatus` would be a better name IMO

src/analysis/analysis-poller.js
```diff
@@ -1,22 +1,32 @@
 var _ = require('underscore');
 var BackbonePoller = require('backbone-poller');
 
-var AnalysisPoller = function () {};
+var AnalysisPoller = function () {
```
IagoLast: I prefer to use [function declarations](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function) instead [function expresions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/function) to keep anonymous functions away from the stack traces.

alonsogarciapablo: 👍

alonsogarciapablo:
I closed this PR by mistake. I reverted it and here's the new one: https://github.com/CartoDB/cartodb.js/pull/1807. @IagoLast addressing your comments in the new PR.

src/analysis/analysis-model.js
alonsogarciapablo: Analysis models don't have a `status` until they get one from Maps API.

PR #1807, opened 2017-10-04
Decouple Analysis Polling from AnalysisCollection

alonsogarciapablo:
Fixes https://github.com/CartoDB/cartodb.js/issues/1788.

Remaining work:

- [x] Do not use analysisCollection when reseting models in the analysisPoller (extract sources from layers and dataviews instead).
- [x] Ensure sources are marked/unmarked as referenced:
  - [x] CartoDB Layers (`initialize`, `update`, `remove`)
  - [x] Torque Layers (`initialize`, `update`, `remove`)
  - [x] Dataviews (`initialize`, `remove`)
- [x] Add unit tests for the [getUniqueAnalysisNodes](https://github.com/CartoDB/cartodb.js/pull/1812#discussion_r142947277)

cc: @IagoLast


src/analysis/analysis-model.js
```diff
@@ -110,6 +122,10 @@ var AnalysisModel = Model.extend({
     return this._anyStatus(STATUS.READY, STATUS.FAILED);
   },
 
+  isReady: function () {
+    return this._anyStatus(STATUS.READY);
```
IagoLast:
```javascript
  return this.get('status') === STATUS.READY;
```

src/analysis/analysis-model.js
```diff
@@ -110,6 +122,10 @@ var AnalysisModel = Model.extend({
     return this._anyStatus(STATUS.READY, STATUS.FAILED);
```
IagoLast: `_anyStatus` doesnt look a very descriptive method name to me, we could just use `_.contains`

```javascript
  _.contains([STATUS.READY, STATUS.FAILED], this.get('status'));

```

src/analysis/analysis-poller.js
```diff
@@ -1,22 +1,32 @@
 var _ = require('underscore');
 var BackbonePoller = require('backbone-poller');
 
-var AnalysisPoller = function () {};
+var AnalysisPoller = function () {
```
IagoLast: Use funcion declaration

src/analysis/analysis-poller.js
```diff
@@ -1,22 +1,32 @@
...
-    poller.start();
+AnalysisPoller.prototype.resetAnalysisNodes = function (analysisModels) {
+  this.reset();
+  _.each(analysisModels, function (analysisModel) {
+    this._poll(analysisModel);
+  }, this);
+};
+
+AnalysisPoller.prototype._poll = function (analysisModel) {
+  if (analysisModel.url()) {
```
IagoLast: What is the semantic meaning of this?

An analysis with an url is considered as ready?

src/analysis/analysis-poller.js
```diff
@@ -1,22 +1,32 @@
...
+    var poller = this._findOrCreatePoller(analysisModel);
+    if (analysisModel.hasChanged('url') && poller.active()) {
+      poller.stop();
```
IagoLast: why are we stopping the poller in this case?

src/geo/map/cartodb-layer.js
```diff
@@ -25,6 +25,11 @@ var CartoDBLayer = LayerModelBase.extend({
       this.set('initialStyle', attrs.cartocss);
     }
 
+    // TODO: Validate source?
+    if (attrs.source) {
```
IagoLast: are we going to admit this? I mean, I thought that sources can only be set with `setSource`

src/geo/map/cartodb-layer.js
```diff
@@ -112,8 +117,17 @@ var CartoDBLayer = LayerModelBase.extend({
     return this.get('source');
   },
 
-  setSource: function (source, options) {
-    this.set('source', source, options);
+  setSource: function (newSource, options) {
+    this._updateSourceReferences(newSource);
+    this.set('source', newSource, options);
```
IagoLast: options? for wat?

src/vis/vis.js
```diff
@@ -365,6 +349,14 @@ var VisModel = Backbone.Model.extend({
       includeFilters: true,
       success: function () {
         this.trigger('reloaded');
+        this._restartAnalysisPolling();
+
+        var analysisNodes = this._getAnalysisNodeModels();
+        if (this._isAnyAnalysisNodeLoading(analysisNodes)) {
+          this.trackLoadingObject(this);
```
IagoLast: what does `trackLoadingObject` do?

IagoLast: I mean, we keep a list of "loading objects" but we dont do anything with it :S

src/vis/vis.js
```diff
@@ -433,6 +431,10 @@ var VisModel = Backbone.Model.extend({
     overlayView.type = 'custom';
     this.overlaysCollection.add(overlayView);
     return overlayView;
+  },
+
+  isLoading: function () {
```
IagoLast: where is this method used?

IagoLast: I mean besides specs :P

src/analysis/analysis-poller.js
alonsogarciapablo: This means there is an active poller for this analysisModel to a previous url. Since we now have a new url, we have to stop the previous poller and create a new one.

src/geo/map/cartodb-layer.js
alonsogarciapablo: They should be mandatory when creating the layer and they can only be replaced via `setSource`. We have [this check to ensure dataviews get a `source` when they are initialised](https://github.com/CartoDB/cartodb.js/blob/v4/src/dataviews/dataview-model-base.js#L92). I think we should do the same here. Agree?

src/geo/map/cartodb-layer.js
alonsogarciapablo: For example, for setting the source silently with:

```
layer.setSource(newSource, { silent: true });
```

We're currently using this [here](https://github.com/CartoDB/cartodb/blob/master/lib/assets/core/javascripts/cartodb3/deep-insights-integration/layers-integration.js#L267). We might want to find a better name for that option when we polish the public API more.

src/vis/vis.js
```diff
@@ -261,34 +259,20 @@ var VisModel = Backbone.Model.extend({
...
   },
 
   /**
    * Check if an Analysis node is the source of a layer or a dataview.
    */
   _isAnalysisSourceOfLayerOrDataview: function (analysisModel) {
-    var isAnalysisLinkedToLayer = this._isAnalysisLinkedToLayer(analysisModel);
-    var isAnalysisLinkedToDataview = this._dataviewsCollection.isAnalysisLinkedToDataview(analysisModel);
-    return isAnalysisLinkedToLayer || isAnalysisLinkedToDataview;
+    return this._layersCollection.isAnalysisLinkedToAnyLayer(analysisModel) ||
```
IagoLast: I think we can use:

```javascript 
_isAnalysisSourceOfLayerOrDataview: function (analysisModel) {
  return analysisModel.isSourceOfAnyModel();
}
```

IagoLast: Maybe `isAnalysisLinkedToAnyLayer` is not longer needed

src/vis/vis.js
alonsogarciapablo: It keeps track of "things" that are loading. I admit this is kind of weird now... There're two types of objects that might be "loading":

1. `vis` (when any analysis is still not done). This one is now tracked / untracked every time the vis is reloaded, based on the status of analyses.
2. `leafletCartoDBLayerGroupView` or `gmapsCartoDBLayerGroupView`. These will track or untrack themselves as loading events when tiles are loading / loaded.

We want to mark the vis as `loading` (to display the blue loader/bar at the top of the map) whenever there's any of those two objects are loading.

We're tracking each individual `AnalysisModel` before but I realised it doesn't make sense because as soon as one is not done, the vis is "loading" and the vis will be reloaded when an analysis pointed by a layer or a dataview is ready.

src/vis/vis.js
alonsogarciapablo: Only in tests I believe.

src/vis/vis.js
alonsogarciapablo: :sparkles: :sparkles: :sparkles:

Good catch!

src/analysis/analysis-poller.js
alonsogarciapablo: Analyses don't have a `url` initially. Only after the first request to Maps API they are given a `url`. That's when polling starts so the truth is that every analysisModel should have a `url` at this point. Polling is only possible if there's a `url` and I guess this check is just ensuring polling won't start unless that's true. It's possible that this could work with this check, but it's also not harmful...

IagoLast: In this case maybe we should create a function for this conditional.
([something like this](https://sourcemaking.com/refactoring/decompose-conditional))

```javascript
AnalysisPoller.prototype._poll = function (analysisModel) {
  if (this._canBePolled(analysisModel)) {
    // ... 
  }
}

/**
 * We need the analysis to have an url in order to start polling
 */
AnalysisPoller.prototype._canBePolled = function (analysisModel) {
  return !!analysisModel.url();
}

Comment (IagoLast) [reply src/analysis/analysis-poller.js]: I dont understand the intetion of this:

```javascript
if (analysisModel.hasChanged('url') && poller.active()) {
      poller.stop();
    }
    if (!analysisModel.isDone()) {
      poller.start();
    }
```

src/analysis/analysis-poller.js
IagoLast: Since is a little tricky maybe we could add this exact answer as a comment in the code

src/geo/map/cartodb-layer.js
IagoLast: yep, we must check source for cartolayers and torquelayers

src/vis/vis.js
IagoLast: so IMO we should use the `get('loading')` instead

IagoLast:
@alonsogarciapablo  
> Do not use analysisCollection when reseting models in the analysisPoller (extract sources from layers and dataviews instead).

Solved in: 
https://github.com/CartoDB/cartodb.js/pull/1812

src/vis/vis.js
alonsogarciapablo: I try to hide Backone attrs when possible (so that things don't break if the attr is renamed for example).... Adding this method is useful for tests and it might encourage us developers to use it if we ever need to check if the `vis` is loading... I don't think it's that harmful but if you have a strong reason to remove it, I'll change it. Thanks!

alonsogarciapablo:
@IagoLast Thanks A LOT for all your feedback!

I addressed most of your comments. Would you please give it another look? Thanks!

src/vis/vis.js
IagoLast: :trollface:  https://github.com/CartoDB/cartodb.js/pull/1803#discussion_r142934308

alonsogarciapablo:
Thanks @IagoLast! 👏 👏 👏

PR #1809, opened 2017-10-05
Layers serializer ignores layers not supported by Maps API

alonsogarciapablo:
`layers-serializer` should ignore layers that are not supported by Maps API (eg: layers of type `GMapsBase` instead of throwing an error that breaks the map.

**Acceptance Criteria**

1. Checkout this branch and run `grunt dev`.
2. Visit http://localhost:9001/examples/gmaps-geometry-editor.html.
3. Map should load and you shouldn't see any errors in the console.

cc: @IagoLast @Jesus89


Jesus89:
:rocket:

PR #1810, opened 2017-10-05
Dataview sources cannot be updated

alonsogarciapablo:
We haven't seen any errors in Track.js so we can assume BUILDER is not updating the `source` of dataview models. It doesn't make sense conceptually either so this PR ensures it's not possible to do so.

### DIFF
--- a/src/dataviews/dataview-model-base.js
+++ b/src/dataviews/dataview-model-base.js
@@ -99,7 +99,6 @@ module.exports = Model.extend({
     }
 
     this._initBinds();
-    this._setupAnalysisStatusEvents();
   },
 
   _initBinds: function () {
@@ -117,6 +116,8 @@ module.exports = Model.extend({
     if (this.filter) {
       this.listenTo(this.filter, 'change', this._onFilterChanged);
     }
+
+    this.getSource().on('change:status', this._onAnalysisStatusChange, this);
   },
 
   _onChangeBinds: function () {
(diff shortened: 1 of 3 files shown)


src/dataviews/dataview-model-base.js
```diff
@@ -171,22 +159,11 @@ module.exports = Model.extend({
...
-      : '';
-    track(new Error('[SOURCE] _onSourceChanged [' + changedKeys + ']'));
-  },
-
   _setupAnalysisStatusEvents: function () {
-    this._removeExistingAnalysisBindings();
-    if (this.getSource()) {
-      this.getSource().on('change:status', this._onAnalysisStatusChange, this);
-    }
+    this.getSource().on('change:status', this._onAnalysisStatusChange, this);
```
ivanmalagon: This function `_setupAnalysisStatusEvents` is not called ever.

You should add this logic to `initBinds` in order to get the dataviews to react to analysis status changes.

src/dataviews/dataview-model-base.js
```diff
@@ -263,9 +240,8 @@ module.exports = Model.extend({
   },
 
   update: function (attrs) {
-    if (_.has(attrs, 'source')) {
-      var message = '[SOURCE] Source present in UPDATE attrs. ' + JSON.stringify(attrs.source);
-      track(new Error(message));
+    if (attrs.source) {
```
ivanmalagon: I'd stick to `_.has` because even in the case of someone updating `source` to `undefined` we should throw an error.

src/dataviews/dataview-model-base.js
alonsogarciapablo: It's actually being called in the `initialized` method, but I'll include this logic in `initBinds`. 👍

alonsogarciapablo:
> Have you tried theses changes using the latest master branches of CartoDB and Deep-Insights? Or will we create new PRs for integration?

I haven't tested the integration between these changes an CartoDB and deep-insights.js. Based on the stats that we gathered, I'm assuming these projects are not trying to update dataview sources.

PR #1812, opened 2017-10-05
Remove analysis collection analysis poller pr0

IagoLast:
* Do not use analysisCollection when reseting models in the analysisPoller (extract sources from layers and dataviews instead).

* Remove code-duplication, move getUniqueNodes to the AnalysisService

### DIFF
--- a/src/analysis/analysis-service.js
+++ b/src/analysis/analysis-service.js
@@ -94,11 +94,35 @@ AnalysisService.prototype._removeAnalsyisFromIndex = function (analysis) {
 
 /**
  * Return a list with all the analyses contained in the given collections.
+ * 
+ * @example
+ * We have the following analyses:  (a0->a1->a2), (b0->a2)
+ * This method will give us: (a0->a1->a2), (b0->a2)
  */
 AnalysisService.getAnalysisList = function (layersCollection, dataviewsCollection) {
   var layerAnalyses = _getAnalysesFromLayers(layersCollection);
   var dataviewsAnalyses = _getAnalysesFromDataviews(dataviewsCollection);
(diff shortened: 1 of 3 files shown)


alonsogarciapablo:
Retest this please

src/vis/vis.js
```diff
@@ -264,7 +264,7 @@ var VisModel = Backbone.Model.extend({
   },
 
   _getAnalysisNodeModels: function () {
-    return this._analysisCollection.models;
+    return this._getUniqueAnalysesNodes();
```
alonsogarciapablo: There's no reason to have both `_getAnalysisNodeModels` and `_getUniqueAnalysesNodes`. I'd just keep `_getUniqueAnalysesNodes`.

src/analysis/analysis-service.js
```diff
@@ -94,13 +94,37 @@ AnalysisService.prototype._removeAnalsyisFromIndex = function (analysis) {
...
 
+/**
+ * Return all the analysis nodes without duplicates.
+ * The analyses are obtained from the layers and dataviews collections.
+ * 
+ * @example
+ * We have the following analyses:  (a0->a1->a2), (b0->a2)
+ * This method will give us: (a0->a1->a1), (a1->a2), (a2), (b0->a2)
+ */
+AnalysisService.getUniqueAnalysesNodes = function (layersCollection, dataviewsCollection) {
```
alonsogarciapablo: Do we want to add some tests for this method?

IagoLast: `the current test suite already covers this` :P

alonsogarciapablo: Perhaps indirectly. I'm referring to adding unit tests for this new function...

Issue #1790, opened 2017-09-28
AnalysisFactory#findNodeById doesn't depend on internal AnalysisCollection


IagoLast:
The goal of this issue is to refactor `vis.analysis.findNodeById()` so it doesn't use analysisCollection internally and finds the node in sources from `layersCollection` and `dataviewsCollection`.

Also:

`vis#load` is reading analysis from viz.json and doing:

1. "Analysing" exiting analysis definitions.

```
    if (vizjson.analyses) {
      _.each(vizjson.analyses, function (analysis) {
        this.analysis.analyse(analysis);
      }, this);
    }
```
2. Finding previously generated nodes from a node ID.

```
  var layerModels = _.map(vizjson.layers, function (layerData, layerIndex) {
      ...
      if (layerData.source) {
        layerData.source = this.analysis.findNodeById(layerData.source);
      }
      ...
  });
```

`analysis.findNodeById` is changing in this issue so we can get rid of 1. and change 2. so that it finds the analysis definition in the viz.json and then uses  `this.analysis.analyse(...)` to "create" the analysis model.

Analysis models will now be referenced directly by some layer models and dataview models (implemented in `sources-as-references`). That's the reason why we're getting rid of the internal `AnalysisCollection` and it's the reason why [`AnalysisFactory#analyse`](https://github.com/CartoDB/cartodb.js/blob/v4/src/analysis/analysis-factory.js#L27) shouldn't need to add nodes to AnalysisCollection.

For now, we'll keep this method as the public (as in "API" public) way of generating an analysis tree from an analysis definition. We'll rethink the API to manage analysis in the near future.

PR #1815, opened 2017-10-06
Public findNodeById method using static AnalysisService method

Jesus89:
Related to: #1790

Publish analysis.findNodeById(id) using the static method findNodeById from the AnalysisService

### DIFF
--- a/src/analysis/analysis-model.js
+++ b/src/analysis/analysis-model.js
@@ -173,6 +173,7 @@ var AnalysisModel = Model.extend({
   /**
    * Return an Array with the complete node list for this analysis.
    */
+  // TODO: use backbone collection
   getNodes: function () {
     // Add current node to the list
     var nodes = [this];
--- a/src/analysis/analysis-model.js
+++ b/src/analysis/analysis-model.js
@@ -1,4 +1,5 @@
 var _ = require('underscore');
+var Backbone = require('backbone');
 var Model = require('../core/model');
 var util = require('../core/util');
 
@@ -173,7 +174,6 @@ var AnalysisModel = Model.extend({
   /**
    * Return an Array with the complete node list for this analysis.
    */
-  // TODO: use backbone collection
   getNodes: function () {
     // Add current node to the list
     var nodes = [this];
@@ -184,6 +184,13 @@ var AnalysisModel = Model.extend({
       }, this);
     }
     return nodes;
+  },
+
+  /**
+   * Return a Collection with the complete node list for this analysis.
+   */
+  getNodesCollection: function () {
+    return new Backbone.Collection(this.getNodes());
   },
 
   /**
--- a/src/analysis/analysis-service.js
+++ b/src/analysis/analysis-service.js
@@ -142,8 +142,8 @@ AnalysisService.prototype._getAnalysisAttributesFromAnalysisDefinition = functio
  * Return the analysis node with the provided id
  */
 AnalysisService.prototype.findNodeById = function (id) {
-  var analyses = AnalysisService.getUniqueAnalysesNodes(this._layersCollection, this._dataviewsCollection);
-  return _.find(analyses, function (analysisNode) {
+  var analysisNodes = AnalysisService.getUniqueAnalysisNodes(this._layersCollection, this._dataviewsCollection);
+  return _.find(analysisNodes, function (analysisNode) {
     return analysisNode.get('id') === id;
   });
 };
@@ -156,11 +156,11 @@ AnalysisService.prototype.findNodeById = function (id) {
  * We have the following analyses:  (a0->a1->a2), (b0->a2)
  * This method will give us: (a0->a1->a2), (a1->a2), (a2), (b0->a2)
  */
-AnalysisService.getUniqueAnalysesNodes = function (layersCollection, dataviewsCollection) {
+AnalysisService.getUniqueAnalysisNodes = function (layersCollection, dataviewsCollection) {
   var uniqueAnalyses = {};
-  var analyses = AnalysisService.getAnalysisList(layersCollection, dataviewsCollection);
-  _.each(analyses, function (analysis) {
-    _.each(analysis.getNodes(), function (analysisNode) {
+  var analysisList = AnalysisService.getAnalysisList(layersCollection, dataviewsCollection);
+  _.each(analysisList, function (analysis) {
+    analysis.getNodesCollection().each(function (analysisNode) {
       uniqueAnalyses[analysisNode.get('id')] = analysisNode;
     });
   });
--- a/src/analysis/analysis-service.js
+++ b/src/analysis/analysis-service.js
@@ -24,7 +24,7 @@ var AnalysisService = function (opts) {
  * may have one or more "source" params pointing to another node. If a node had been created
  * already, this method updates the attributes of the existing node. New nodes are added to
  * the collection of analyses that has been injected.
- * 
+ *
  * TODO: document what's the analysis definition
  */
 AnalysisService.prototype.analyse = function (analysisDefinition) {
@@ -80,10 +80,6 @@ AnalysisService.prototype._onAnalysisRemoved = function (analysis) {
   analysis.unbind('destroy', this._onAnalysisRemoved);
 };
 
-AnalysisService.prototype.findNodeById = function (id) {
-  return this._analysisCollection.get(id);
-};
-
 AnalysisService.prototype._addAnalysisToCollection = function (analysis) {
   return this._analysisCollection.add(analysis);
 };
@@ -93,8 +89,18 @@ AnalysisService.prototype._removeAnalsyisFromIndex = function (analysis) {
 };
 
 /**
+ * Return the analysis node with the provided id
+ */
+AnalysisService.findNodeById = function (id, layersCollection, dataviewsCollection) {
+  var analyses = AnalysisService.getUniqueAnalysesNodes(layersCollection, dataviewsCollection);
+  return _.find(analyses, function (analysisNode) {
+    return analysisNode.get('id') === id;
+  });
+};
+
+/**
  * Return a list with all the analyses contained in the given collections.
- * 
+ *
  * @example
  * We have the following analyses:  (a0->a1->a2), (b0->a2)
  * This method will give us: (a0->a1->a2), (b0->a2)
@@ -108,7 +114,7 @@ AnalysisService.getAnalysisList = function (layersCollection, dataviewsCollectio
 /**
  * Return all the analysis nodes without duplicates.
  * The analyses are obtained from the layers and dataviews collections.
- * 
+ *
  * @example
  * We have the following analyses:  (a0->a1->a2), (b0->a2)
  * This method will give us: (a0->a1->a1), (a1->a2), (a2), (b0->a2)
--- a/src/analysis/analysis-service.js
+++ b/src/analysis/analysis-service.js
@@ -19,12 +19,12 @@ var AnalysisService = function (opts) {
   this._apiKey = opts.apiKey;
   this._authToken = opts.authToken;
   this._camshaftReference = opts.camshaftReference || camshaftReference; // For testing purposes
-  this._layersCollection = opts.layersCollection || [];
-  this._dataviewsCollection = opts.dataviewsCollection || [];
+  this._layersCollection = opts.layersCollection;
+  this._dataviewsCollection = opts.dataviewsCollection;
 };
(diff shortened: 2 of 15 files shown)


src/analysis/analysis-service.js
```diff
@@ -80,10 +80,6 @@ AnalysisService.prototype._onAnalysisRemoved = function (analysis) {
   analysis.unbind('destroy', this._onAnalysisRemoved);
 };
 
-AnalysisService.prototype.findNodeById = function (id) {
```
alonsogarciapablo: Why not just change the signature of this method and use `AnalysisService.getUniqueAnalysesNodes` internally? I'm a bit concerned about defining this method to the instance of the `AnalysisService` at runtime [here](https://github.com/CartoDB/cartodb.js/pull/1815/files#diff-f99ed960498d5ea8717b3e74d54bfb01R147).

IagoLast: this is a temporal fix, all this logic will be moved to the engine.

The engine will have both (layers and dataviews list) allowing the rest of our code to be state-less.

So the engine will call `AnalysisService.findByID(id, layersCollection, dataviewsCollection)``
in the meantime we expose this functionality through visModel to allow backward compatibility

src/vis/vis.js
```diff
@@ -144,6 +144,10 @@ var VisModel = Backbone.Model.extend({
       vis: this
     });
 
+    this.analysis.findNodeById = function (nodeId) {
```
alonsogarciapablo: Now that I understand what this is here... And., since this method is currently "public" (for BUILDER and deep-insights.js), do we want to add a very basic test for it?

src/vis/vis.js
```diff
@@ -137,17 +137,21 @@ var VisModel = Backbone.Model.extend({
...
       vis: this
     });
 
-    var layersFactory = new LayersFactory({
-      visModel: this,
-      windshaftSettings: windshaftSettings
-    });
+    // Public namespace exposing public methods.
+    this.analysis = {
+      createAnalysis: this.analysisService.createAnalysis,
```
alonsogarciapablo: Are we changing the public API then?

src/analysis/analysis-service.js
```diff
@@ -51,8 +51,26 @@ AnalysisService.prototype.analyse = function (analysisDefinition) {
   return analysis;
 };
```
IagoLast: Missing doc

src/vis/vis.js
```diff
@@ -137,17 +136,21 @@ var VisModel = Backbone.Model.extend({
...
       analysisCollection: this._analysisCollection,
       vis: this
     });
 
-    var layersFactory = new LayersFactory({
-      visModel: this,
-      windshaftSettings: windshaftSettings
-    });
+    // Public namespace exposing public methods.
+    this.analysis = {
```
alonsogarciapablo: Should we add a couple of tests for this namespace?

IagoLast: IMO not yet, this functionality is covered by another tests.

At the end of the project we can revisit the public api and write exclusive tests for it

src/vis/vis.js
Jesus89: Yes. We are going to remove the analysis method. The question is:
* analysis.createAnalysis() vs
* analysis.create()

IagoLast: OUCH! you are right, we might create an alias for **backwards compatibility***.

> *since im using this expression very much im gonna write **BC**

Jesus89: I think we don't need backwards compat. We can fix the Builder and DI to use the new method

IagoLast: @Jesus89  Remember this methods will be exposed through the engine

#### option1 no namespace

```javascript
// Currently we need a json definition to create an analysis;
const analysisDefinition = `{
  type: 'source',
  id: 'a0',
  query: 'SELECT * from carto.lugares',
}`;
// Use engine to create the analysis
const analysis = engine.createAnalysis(analysisDefinition);
// The analysis will be the source of our layer
const layer1 = new carto.Layer('l1', analysis, style);
```


#### option2 scoped namespace

```javascript
// Currently we need a json definition to create an analysis;
const analysisDefinition = `{
  type: 'source',
  id: 'a0',
  query: 'SELECT * from carto.lugares',
}`;
// Use engine to create the analysis
const analysis = engine.analysis.create(analysisDefinition);
// The analysis will be the source of our layer
const layer1 = new carto.Layer('l1', analysis, style);
```


#### option3  (final version)

```javascript
// The analysis can be instantiated from its constructor
const analysis = new carto.analysis.Source('a1', 'SELECT * from carto.lugares');
// The analysis will be the source of our layer
const layer1 = new carto.Layer('l1', analysis, style);
```

alonsogarciapablo: We could also leave the "public" method as `analyse` and postpone this decision. There'd be no integration work and we can get back to it once we start exposing analysis management...

src/vis/vis.js
```diff
@@ -427,6 +408,59 @@ var VisModel = Backbone.Model.extend({
...
     return this.get('loading');
+  },
+
+  /**
+   * "Load" existing analyses from the viz.json. This will generate
+   * the analyses graphs and index analysis nodes in the
+   * collection of analysis
+   */
+  _createAnalysisNodes: function (analysesDefinition) {
+    var analysisNodes = {};
```
IagoLast: rename this to `analyses` for being consistent with older code

src/vis/vis.js
```diff
@@ -427,6 +408,59 @@ var VisModel = Backbone.Model.extend({
 
   isLoading: function () {
     return this.get('loading');
+  },
+
+  /**
```
IagoLast:
```javascript 
/**
 * Returns a collection with all the unique analysis nodes defined in the analysis section under the vizjson

File: src/vis/vis.js [hunk 8]
```diff
@@ -427,6 +408,59 @@ var VisModel = Backbone.Model.extend({
...
+    var analysisRoots = [];
+    if (analysesDefinition) {
+      _.each(analysesDefinition, function (analysisDefinition) {
+        analysisRoots.push(this.analysisService.createAnalysis(analysisDefinition));
+      }, this);
+
+      _.each(analysisRoots, function (analysisRoot) {
+        _.each(analysisRoot.getNodes(), function (analysisNode) {
+          analysisNodes[analysisNode.get('id')] = analysisNode;
+          // TODO: remove
```
IagoLast: Complete this todo: why remove?

src/analysis/analysis-model.js
```diff
@@ -173,6 +173,7 @@ var AnalysisModel = Model.extend({
   /**
    * Return an Array with the complete node list for this analysis.
    */
+  // TODO: use backbone collection
   getNodes: function () {
```
Jesus89: Can we return a use backbone collection in this method?

src/vis/vis.js
Jesus89: But the `create` method is different. I prefer to do some integration than use the same method that does different thing, because if we expose `analyse` as the new `create` the behavior is not the same because it doesn't check if the analysis node already exists

src/vis/vis.js
Jesus89: Remove the analysisCollection add

src/analysis/analysis-service.js
Jesus89: Done

src/vis/vis.js
Jesus89: Use analysisNode**s** instead

src/analysis/analysis-service.js
```diff
@@ -19,21 +19,48 @@ var AnalysisService = function (opts) {
   this._apiKey = opts.apiKey;
   this._authToken = opts.authToken;
   this._camshaftReference = opts.camshaftReference || camshaftReference; // For testing purposes
-  this._layersCollection = opts.layersCollection;
-  this._dataviewsCollection = opts.dataviewsCollection;
+  this._layersCollection = opts.layersCollection || [];
```
IagoLast: IMO creating a new collection is a bad idea that could hide errors.

Anyway, it should be initialized with an empty collection not an array
```javascript
 this._layersCollection = opts.layersCollection || new Backbone.Collection([]);
```

src/analysis/analysis-service.js
```diff
@@ -19,21 +19,48 @@ var AnalysisService = function (opts) {
...
 };
 
 /**
- * Create an analysis node
- *
- * Recursively generates a graph of analyses and returns the "origin" node. Each node
- * may have one or more "source" params pointing to another node. If a node had been created
- * already, this method updates the attributes of the existing node. New nodes are added to
- * the collection of analyses that has been injected.
+  * Recursively generates a graph of analyses and returns the "origin" node. Each node
```
IagoLast: If we agree on the following naming for the nodes: root, inner and leaf
The documentation should use it

> returns the **"root"** node

src/analysis/analysis-service.js
Jesus89: Ups this mustn't be there. I prefer without || [] instead of requiring backbone in the service

IagoLast:
LGTM 🏳️

src/vis/vis.js
```diff
@@ -135,16 +135,21 @@ var VisModel = Backbone.Model.extend({
...
       apiKey: this.get('apiKey'),
       authToken: this.get('authToken'),
       layersCollection: this._layersCollection,
       dataviewsCollection: this._dataviewsCollection
     });
 
-    // Public namespace exposing public methods.
-    this.analysis = this.analysisService;
+    // Public wrapper exposing public methods.
+    this.analysis = {
```
alonsogarciapablo: Why not expose the instance of the `AnalysisService` as we're doing before?

Jesus89: To expose only the public methods and nothing more in the wrapper/namespace. But it works pointing to the analysisService instance also.

src/analysis/analysis-service.js
```diff
@@ -19,21 +19,48 @@ var AnalysisService = function (opts) {
...
+  * may have one or more "source" params pointing to another node. If a node had been created
+  * already, this method updates the attributes of the existing node. New nodes are added to
+  * the collection of analyses that has been injected.
+  */
+AnalysisService.prototype.analyse = function (analysisDefinition) {
```
alonsogarciapablo: Could we use `createAnalysis` and `updateAnalysis` here too?

```
AnalysisService.prototype.analyse = function (analysisDefinition) {
  var analysis = this.findNodeById(analysisDefinition.id);
  if (analysis) {
    this.updateAnalysis(analysisDefinition);
  } else {
    analysis = this.createAnalysis(analysisDefinition);
  }
  return analysis;
}
```

src/analysis/analysis-service.js
```diff
@@ -80,72 +93,44 @@ AnalysisService.prototype.createSourceAnalysisForLayer = function (layerId, laye
...
   var analysis = this.findNodeById(analysisDefinition.id);
 
-  if (analysis) {
-    var analysisAttrs = this._createAnalysisAttributesFromAnalysisDefinition(analysisDefinition);
-    analysis.set(analysisAttrs);
+  if (!analysis) {
+    throw new Error('The analysis ' + analysisDefinition.id + ' does not exist');
   }
 
+  var analysisAttrs = this._getAnalysisAttributesFromAnalysisDefinition(analysisDefinition, this.updateAnalysis.bind(this));
```
alonsogarciapablo: I don't understand why need to pass `this.updateAnalysis.bind(this)` (and `this.createAnalysis.bind(this)`) to `_getAnalysisAttributesFromAnalysisDefinition`... Couldn't we just invoke `this._analyse(...)` or `this.createAnalysis` inside of `_getAnalysisAttributesFromAnalysisDefinition`?

Jesus89: This is to use the same method `_getAnalysisAttributesFromAnalysisDefinition` instead of writing the code three times. It can be done in different ways

src/analysis/analysis-service.js
Jesus89: I don't think so, because the analyse method must be applied recursively to the sourceNodes. Your approach executes the analyse method only in the first iteration

alonsogarciapablo: `this.createAnalysis` creates analysis nodes recursively... Doesn't it?

Jesus89: Yes but it applies recursive `createAnalysis`,  no recursisve `analyse`

Issue #1796, opened 2017-10-02
Integrate remove-analysis-collection


alonsogarciapablo:

IagoLast:
We decided to keep using the `engine` internally:
```javascript 
// namespaced under `analysis` 
vis.analysis.findNodeById = function(id){
  return engine.findNodeById(id);
}
```


engine also delegates the call:
```javascript 
engine.findNodeById = function(id){
  return AnalysisService.findNodeById(id, this.layersCollection, this.dataviewsCollection);
}
```

Jesus89:
Integrates #1790, #1763, #1787, #1793, #1788, #1789

Related PRs:
- CARTO.js: https://github.com/CartoDB/cartodb.js/pull/1821
- deep-insights: https://github.com/CartoDB/deep-insights.js/pull/606
- CartoDB: https://github.com/CartoDB/cartodb/pull/12947

Jesus89:
In production! :rocket:

Issue #1789, opened 2017-09-28
Remove analysisCollection from AnalysisService


IagoLast:
Analysis models will now be referenced directly by some layer models and dataview models (implemented in `sources-as-references`). That's the reason why we're getting rid of the internal `AnalysisCollection` and it's the reason why [`AnalysisFactory#analyse`](https://github.com/CartoDB/cartodb.js/blob/v4/src/analysis/analysis-factory.js#L27) shouldn't need to add nodes to AnalysisCollection.

For now, we'll keep this method as the public (as in "API" public) way of generating an analysis tree from an analysis definition. We'll rethink the API to manage analysis in the near future.

Jesus89:
Done in https://github.com/CartoDB/cartodb.js/issues/1790

PR #1821, opened 2017-10-13
Remove analysis collection

Jesus89:
Related to: #1796

**EPIC**: #1790, #1763, #1787, #1793, #1788, #1789


src/analysis/analysis-service.js
```diff
@@ -0,0 +1,151 @@
...
+
+    this._analysisNodes[analysisDefinition.id] = analysis;
+    analysis.bind('destroy', this._onAnalysisRemoved, this);
+  }
+
+  return analysis;
+};
+
+/**
+ * This function applies recursively the function analyse to the sourceNodes in the analysisDefinition
```
IagoLast: The description is wrong.

> This function is used to iterate over the analysis graph. It uses the camshaft reference to extract those parameters which are analysis nodes. And call `analyse` on them.

Maybe we could add a comment remaining this wont be needed if we split `parameters` and `inputs` in the analysis model

examples/analysis_test.html
```diff
@@ -323,7 +323,7 @@
         .done(function(vis, layers) {
           var tableName = 'arboles'
 
-          var analysis = vis.analysis.analyse({
+          var analysis = vis.analysis.createAnalysis({
```
IagoLast: this method doesn`t exist anymore

src/geo/map/cartodb-layer.js
```diff
@@ -125,11 +134,21 @@ var CartoDBLayer = LayerModelBase.extend({
   },
 
   update: function (attrs) {
```
IagoLast: This method got a little tricky

src/vis/vis.js
```diff
@@ -136,18 +134,18 @@ var VisModel = Backbone.Model.extend({
...
+      authToken: this.get('authToken')
     });
 
-    var layersFactory = new LayersFactory({
-      visModel: this,
-      windshaftSettings: windshaftSettings
-    });
+    // Public wrapper exposing public methods.
+    this.analysis = {
+      analyse: this._analysisService.analyse.bind(this._analysisService),
```
IagoLast: is the `.bind` required?

src/windshaft/map-serializer/anonymous-map-serializer/analysis-serializer.js
```diff
@@ -1,55 +1,35 @@
...
  */
-function _generateUniqueAnalysisList (analysesList) {
-  var analysesIds = {};
-  return _.reduce(analysesList, function (list, analysis) {
-    if (!analysesIds[analysis.get('id')] && !_isAnalysisPartOfOtherAnalyses(analysis, analysesList)) {
-      analysesIds[analysis.get('id')] = true; // keep a set of already added analysis.
+function _generateUniqueAnalysisList (analysisList) {
+  var analysisIds = {};
+  return _.reduce(analysisList, function (list, analysis) {
+    if (!analysisIds[analysis.get('id')] && !_isAnalysisPartOfOtherAnalyses(analysis, analysisList)) {
```
IagoLast: we can use `analysis.isSourceOfAnyModel` instead `_isAnalysisPartOfOtherAnalyses` 

In fact, `_isAnalysisPartOfOtherAnalyses` could be removed

test/spec/analysis/analysis-model.spec.js
```diff
@@ -101,15 +118,101 @@ describe('src/analysis/analysis-model.js', function () {
...
+        return analysisModel;
+      };
+
+      var createAnalysisModelWithStatusWithReferences = function (visModel) {
+        var analysisModel = createFakeAnalysis({ id: 'a0', status: 'foo' }, visModel);
+        analysisModel.markAsSourceOf(new Backbone.Model());
+        return analysisModel;
+      };
+
+      var testCases = [
```
IagoLast: this hacks usually give problems

src/geo/map/cartodb-layer.js
Jesus89: I think we can get rid of the Deprecated condition since it is implemented [in the builder](https://github.com/CartoDB/cartodb/blob/249b01951fa29d2a70d0a4efbdcc52eda9de1b81/lib/assets/core/javascripts/cartodb3/deep-insights-integration/layers-integration.js#L263-L272)

src/vis/vis.js
Jesus89: Yes I think so because of the context this

src/geo/map/cartodb-layer.js
alonsogarciapablo: I'll open a PR to simplify this.

alonsogarciapablo: https://github.com/CartoDB/cartodb.js/pull/1823/files

test/spec/analysis/analysis-model.spec.js
alonsogarciapablo: There shouldn't be any problem with these tests that don't have any `this` involved IMO. The question is... are they easy to understand?

src/windshaft/map-serializer/anonymous-map-serializer/analysis-serializer.js
alonsogarciapablo: I don't think that would work @IagoLast. `analysis.isSourceOfAnyModel` checks if the analysis is referenced by a `layer` or `dataview`. `_isAnalysisPartOfOtherAnalyses` is checking if the analysis is part of another analysis.

Jesus89: I just change the analysisList name in this function, so may be we can discuss changing this method in other PR

PR #1823, opened 2017-10-13
Source cannot be updated via [CartoDB|Torque].prototype.update

alonsogarciapablo:
[CartoDB|Torque].prototype.update will raise an error when given a `source` attr.

cc: @IagoLast @Jesus89

### DIFF
--- a/src/geo/map/cartodb-layer.js
+++ b/src/geo/map/cartodb-layer.js
@@ -136,7 +136,7 @@ var CartoDBLayer = LayerModelBase.extend({
     if (attrs.source) {
       throw new Error('Use ".setSource" to update a layer\'s source instead of the update method');
     }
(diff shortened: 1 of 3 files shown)


alonsogarciapablo:
Perhaps we should be making sure only certain attributes are "updateable" as we do with `DataviewModel`s [here](https://github.com/CartoDB/cartodb.js/blob/v4/src/dataviews/dataview-model-base.js#L270).

IagoLast:
I need more context but as far as I know we should hide backbone methods like 'update' and expose only `setsource`

alonsogarciapablo:
`update` is our way of hiding Backbone's `Backbone.Model.prototype.set`. We should move towards more granular setters but It'll require some time.

PR #1826, opened 2017-10-17
1818 create engine class

IagoLast:
https://github.com/CartoDB/cartodb.js/issues/1818
---

- [x] Create a engine class.
- [x] Allow to add layers.
- [x] Serialize the status 
- [x] Allow to add dataviews.
- [x] Handle server responses


src/engine.js
```diff
@@ -0,0 +1,187 @@
...
+ */
+function Engine (params) {
+  if (!params) throw new Error('new Engine() called with no paramters');
+  this._apiKey = params.apiKey;
+  this._authToken = params.authToken;
+  this._username = params.username;
+  this._serverUrl = params.serverUrl;
+  this._templateName = params.templateName;
+  this._isNamedMap = params.templateName !== undefined;
+  this._stat_tag = params.statTag;
```
alonsogarciapablo: `this._stat_tag` -> `this._statTag`?

src/engine.js
```diff
@@ -0,0 +1,187 @@
...
+ * @param {boolean} params.templateName - While we dont remove named maps we must explicitly say when the map is named. Defaults to false.
+ * @param {boolean} params.statTag - Token used to get map view statistics.
+ * @class
+ */
+function Engine (params) {
+  if (!params) throw new Error('new Engine() called with no paramters');
+  this._apiKey = params.apiKey;
+  this._authToken = params.authToken;
+  this._username = params.username;
+  this._serverUrl = params.serverUrl;
```
alonsogarciapablo: Some of these instance variables (eg: `this._serverUrl`) are only being used in this context (when we declare `windshaftSettings`). I wouldn't personally use instance variables if they are not used elsewhere.

src/engine.js
```diff
@@ -0,0 +1,187 @@
...
+    return params;
+  }
+  if (this.auth_token) {
+    params.auth_token = this._authToken;
+    return params;
+  }
+};
+
+/**
+ * @api
```
alonsogarciapablo: I'd keep `Engine` "private" (as in not part of the Public API) for now.

test/spec/engine.spec.js
```diff
@@ -116,6 +116,18 @@ describe('Engine', function () {
...
+    it('Should use the sourceID parameter', function () {
+      // Error server response
+      spyOn($, 'ajax').and.callFake(function (params) { params.success(FAKE_RESPONSE); });
+      // Spy on modelupdater to ensure thats called with fakesourceId
+      var spy = spyOn(engine._modelUpdater, 'updateModels');
+      // Attach the error event to a spy.
+      engine.on(Engine.Events.RELOAD_SUCCESS, spy);
+
+      engine.reload('fakeSourceId');
+      expect(spy).toHaveBeenCalledWith(jasmine.anything(), 'fakeSourceId', undefined);
```
alonsogarciapablo: To make the test a bit more readable, we could call the spy something like `updateModelsSpy`.

src/engine.js
```diff
@@ -107,31 +104,17 @@ Engine.prototype.off = function off (event, callback) {
...
  * 
  * Once the response has arrived trigger a 'reload-succes' or 'reload-error' event.
+ * 
+ * @param {string} sourceId - The sourceId triggering the reload event. This is usefull to prevent uneeded requests and save data.
+ * @param {string} forceFetch - ????
  */
-Engine.prototype.reload = function reload () {
-  var params = this._getParams();
+Engine.prototype.reload = function reload (sourceId, forceFetch) {
+  var params = this._getParams(sourceId, forceFetch);
```
alonsogarciapablo: Do we need to pass `sourceId` and `forceFetch` to this method? It looks like we're not using them.

test/spec/engine.spec.js
```diff
@@ -145,8 +145,22 @@ describe('Engine', function () {
       engine.reload();
     });
 
-    it('Should update the cartolayerGroup metadata according to the server response', function (done) {
-      pending('Test not implemented');
+    fit('Should update the cartolayerGroup metadata according to the server response', function (done) {
+      // Successfull server response
```
alonsogarciapablo: Would the following work? It's very similar but we're getting rid of the `done` which might makes tests a bit slower. 

```
      engine.addLayer(layer);

      engine.reload();

      // Successfull server response
      spyOn($, 'ajax').and.callFake(function (params) { params.success(FAKE_RESPONSE); });

      var urls = engine._layerGroupModel.attributes.urls;
      // Ensure the modelUpdater has updated the cartoLayerGroup urls
      expect(urls.attributes[0]).toEqual('http://example.com/api/v1/map/2edba0a73a790c4afb83222183782123:1508164637676/0/attributes');
      expect(urls.grids[0]).toEqual(['http://example.com/api/v1/map/2edba0a73a790c4afb83222183782123:1508164637676/0/{z}/{x}/{y}.grid.json']);
      expect(urls.image).toEqual('http://example.com/api/v1/map/static/center/2edba0a73a790c4afb83222183782123:1508164637676/{z}/{lat}/{lng}/{width}/{height}.{format}');
      expect(urls.tiles).toEqual('http://example.com/api/v1/map/2edba0a73a790c4afb83222183782123:1508164637676/{layerIndexes}/{z}/{x}/{y}.{format}');
```

test/spec/engine.spec.js
```diff
@@ -0,0 +1,186 @@
...
+var Backbone = require('backbone');
+var Engine = require('../../src/engine');
+var FakeFactory = require('../helpers/fakeFactory');
+var FAKE_RESPONSE = require('./windshaft/response.mock');
+var CartoDBLayer = require('../../src/geo/map/cartodb-layer');
+
+describe('Engine', function () {
+  var fakeVis = new Backbone.Model();
+  describe('Constructor', function () {
+    it('Should throw a descriptive error when called with no parameters', function () {
```
alonsogarciapablo: We usually use `should` instead of `Should` so that tests read better.

alonsogarciapablo:
I tried running `npm run docs` to checkout the docs and it's showing the following warnings:

```
WARNING: The @public tag does not permit a value; the value will be ignored. File: engine.js, line: 120
WARNING: The @public tag does not permit a value; the value will be ignored. File: engine.js, line: 128
WARNING: The @private tag does not permit a value; the value will be ignored. File: engine.js, line: 132
WARNING: The @private tag does not permit a value; the value will be ignored. File: engine.js, line: 137
WARNING: The @private tag does not permit a value; the value will be ignored. File: engine.js, line: 143
WARNING: The @private tag does not permit a value; the value will be ignored. File: engine.js, line: 148
WARNING: The @private tag does not permit a value; the value will be ignored. File: engine.js, line: 183
WARNING: The @private tag does not permit a value; the value will be ignored. File: engine.js, line: 187
```

I did some investigation and fixed a similar error in https://github.com/CartoDB/cartodb.js/commit/362e2cf490b7279e8cd8c3abfb99601249c53f97.

IagoLast:
Good catch @alonsogarciapablo fixed the jsdoc warnings in 139e964

src/windshaft/response.js
```diff
@@ -0,0 +1,149 @@
...
+ * @constructor
+ * @param {object} windshaftSettings - Object containing the request options.
+ * @param {string} serverResponse - The json string representing a windshaft response to a map instantiation.
+ */
+function Response (windshaftSettings, serverResponse) {
+  this._windshaftSettings = windshaftSettings;
+  this._layerGroupId = serverResponse.layergroupid;
+  this._layers = serverResponse.metadata.layers;
+  this._dataviews = serverResponse.metadata.dataviews;
+  this._analyses = serverResponse.metadata.dataviews;
```
alonsogarciapablo: Something is not right here (`serverResponse.metadata.dataviews`) ☝️

PR #1836, opened 2017-10-20
Reduce d3 footprint [leapfrog]

ivanmalagon:
Currently we're requiring the whole d3 package to use only a few helper functions. This was made because in d3 v3 you couldn't require isolated modules. This is now possible so my solution is to require only the needed packages.

This way we go from

screen shot 2017-10-20 at 16 53 43

to

screen shot 2017-10-20 at 17 00 56
screen shot 2017-10-20 at 17 01 30

**Reduction**

From 146.2K to 24.9K!!!
From 52.5K to 8.7K (gzipped)!!

83% of size reduction. ❗️❗️ 👏 👏 

**Tests**

I made tests with the previous versions of d3 array functions to get values and check that it remained the same when switching to single packages. I deleted them because they are meaningless outside this refactor.

I added tests to `formatter.formatDate`. They didn't exist before and it's useful to keep them.


matallo:
👏

PR #1839, opened 2017-10-23
Fix analysis-model when there are optional sources

ivanmalagon:
https://github.com/CartoDB/camshaft/issues/334
----

This PR fixs a bug related to SQL functions with only one source.

### DIFF
--- a/src/analysis/analysis-model.js
+++ b/src/analysis/analysis-model.js
@@ -180,7 +180,10 @@ var AnalysisModel = Model.extend({
     // Recursively iterate through the inputs ( source nodes have no inputs )
     if (this.get('type') !== 'source') {
       _.forEach(this._getSourceNames(), function (sourceName) {
-        nodes = nodes.concat(this.get(sourceName).getNodes());
+        var source = this.get(sourceName);
+        if (source) {
+          nodes = nodes.concat(source.getNodes());
+        }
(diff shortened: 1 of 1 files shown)


src/analysis/analysis-model.js
```diff
@@ -180,7 +180,10 @@ var AnalysisModel = Model.extend({
     // Recursively iterate through the inputs ( source nodes have no inputs )
     if (this.get('type') !== 'source') {
       _.forEach(this._getSourceNames(), function (sourceName) {
-        nodes = nodes.concat(this.get(sourceName).getNodes());
+        var source = this.get(sourceName);
+        if (source) {
```
dgaubert: What happens if a source is mandatory and it is NOT provided? Is there a way to check whether a source is optional by looking into camshaft-reference?

ivanmalagon: It'll work as usual: passing that responsability to Maps API.

We added checks to look for optional sources (back in the day when we were developing SQL function) and @alonsogarciapablo decided to get rid of those checks at `cartodb.js` It'd be the same scenario as having an analysis with all sources mandatory and not sending the source in `mapConfig`: windshaft will return an error.

ramiroaznar:
:banderísima:

PR #1850, opened 2017-10-31
1847 api skeleton

IagoLast:
#1847
----


src/api/v4/source/dataset.js
```diff
@@ -0,0 +1,24 @@
+var AnalysisModel = require('../../../analysis/analysis-model');
+var CamshaftReference = require('../../../analysis/camshaft-reference');
+
+function Dataset (id, dataset) {
+  this._id = id;
+  this._dataset = dataset;
+}
+
+Dataset.prototype.$setEngine = function (engine) {
```
alonsogarciapablo: @ivanmalagon @Jesus89 we're following this convention for methods that are public for internal use, but not public for API users.

PR #1858, opened 2017-10-31
Formula dataview

ivanmalagon:
https://github.com/CartoDB/cartodb.js/issues/1851
----


src/api/v4/client.js
```diff
@@ -1,35 +1,34 @@
 var Events = require('./events');
 var Engine = require('../../engine');
 var LeafletCartoLayerGroupView = require('../../geo/leaflet/leaflet-cartodb-layer-group-view');
+var VERSION = require('../../../package.json').version;
```
IagoLast: This will include the whole `package.json` inside the bundle for a single `number`.

We can define `VERSION` using [webpack define plugin](https://webpack.js.org/plugins/define-plugin/) as described [here](https://github.com/webpack/webpack/issues/237#issuecomment-40398916).

src/api/v4/dataview/base.js
```diff
@@ -0,0 +1,227 @@
...
+
+Base.prototype.getData = function () {
+  throw new Error('getData must be implemented by the particular dataview.');
+};
+
+// syncOnBoundingChanges
+// isSyncedOnBoundingChanges
+
+// Protected methods
+
```
IagoLast: Can [@protected](http://usejsdoc.org/tags-protected.html) help here?

src/api/v4/dataview/base.js
```diff
@@ -0,0 +1,227 @@
...
+Base.prototype._initialize = function (source, column, options) {
+  options = this._defaultOptions(options);
+
+  this._checkSource(source);
+  this._checkColumn(column);
+  this._checkOptions(options);
+
+  this._source = source;
```
IagoLast: Sime inline comments explaining those fields would be great for newcomers in the dataviews world like me ^^

src/api/v4/dataview/base.js
```diff
@@ -0,0 +1,227 @@
...
+Base.prototype.getData = function () {
+  throw new Error('getData must be implemented by the particular dataview.');
+};
+
+// syncOnBoundingChanges
+// isSyncedOnBoundingChanges
+
+// Protected methods
+
+Base.prototype._initialize = function (source, column, options) {
```
IagoLast: Does it makes sense to you to add comments in the protected methods?

test/spec/api/v4/dataview/formula.spec.js
```diff
@@ -0,0 +1,207 @@
+var Backbone = require('backbone');
+var _ = require('underscore');
+var carto = require('../../../../../src/api/v4/index');
+
+function createFakeInternalModel () {
+  var internalModel = {
```
IagoLast: I like to call this kind of objects`internalModelMock` || `fakeInternalModel`, what do you think?

src/api/v4/dataview/base.js
```diff
@@ -0,0 +1,227 @@
...
+  if (this._internalModel) {
+    this._internalModel.set('enabled', enabled);
+  }
+  return this;
+};
+
+Base.prototype._listenToInternalModelEvents = function () {
+  if (this._internalModel) {
+    this.listenTo(this._internalModel, 'change:data', this._onDataChanged);
+    this.listenTo(this._internalModel, 'change:column', this._onColumnChanged);
```
IagoLast: In which ways can be the column changed?

By the user through the public class and ... ?

src/api/v4/constants.js
```diff
@@ -0,0 +1,45 @@
+var _ = require('underscore');
```
IagoLast: should this file be under the  `dataviews` folder?

src/api/v4/client.js
Jesus89: Ok, Let's use that

src/api/v4/dataview/base.js
Jesus89: Only with dataviewInstance.setColumn('...'), but we are adding events for every property that can be set

src/api/v4/constants.js
Jesus89: In this file we have carto.operation Enum and carto.dataview.status Enum. May be we can split it in two files

test/spec/api/v4/dataview/formula.spec.js
Jesus89: internalModelMock

src/api/v4/dataview/base.js
Jesus89: I will leave all the non-API documentation after internal release

src/api/v4/dataview/base.js
Jesus89: Same as above

src/api/v4/dataview/base.js
Jesus89: Added doc in the functions' header

src/api/v4/dataview/formula.js
```diff
@@ -0,0 +1,117 @@
...
+      result: this._internalModel.get('data'),
+      nulls: this._internalModel.get('nulls'),
+      type: 'formula'
+    };
+  }
+  return null;
+};
+
+Formula.prototype._defaultOptions = function (options) {
+  options = options || {};
```
alonsogarciapablo: [`_.defaults`](http://underscorejs.org/#defaults) might help here:

```
_.defaults(options, {
  operation: constants.operation.COUNT
});
```

We could actually do something like:

```
// base.js

Base.prototype.DEFAULTS = {};

Base.prototype._initialize = function () {
  ...
  this._options = _.defaults(options, this.DEFAULTS);
});

};

// formula.js

Formula.prototype.DEFAULTS = {
  operation: constants.operation.COUNT
}
```

Just an 💡

src/api/v4/dataview/base.js
```diff
@@ -0,0 +1,234 @@
...
+    this.listenTo(this._internalModel, 'loading', this._onStatusLoading);
+    this.listenTo(this._internalModel, 'loaded', this._onStatusLoaded);
+    this.listenTo(this._internalModel, 'error', this._onStatusError);
+    this._listenToInstanceModelEvents();
```
alonsogarciapablo: I personally find `_listenToInternalModelEvents` and `_listenToInstanceModelEvents` a bit confusing by looking at their names... `_listenToInternalModelSharedEvents` and `_listenToInternalModelSpecificEvents` (or something even longer... 😛 )

src/windshaft/filters/bounding-box.js
```diff
@@ -1,8 +1,29 @@
+var _ = require('underscore');
 var Model = require('../../core/model');
+var BOUNDING_BOX_FILTER_WAIT = 300;
 
 module.exports = Model.extend({
-  initialize: function (bounds) {
-    this.setBounds(bounds);
+  initialize: function (map) {
```
alonsogarciapablo: :+1:

src/dataviews/dataview-model-base.js
```diff
@@ -336,6 +339,24 @@ module.exports = Model.extend({
     if (!(source instanceof AnalysisModel)) {
       throw new Error('Source must be an instance of AnalysisModel');
     }
+  },
+
+  _checkBBoxFilter: function () {
+    if (this.syncsOnBoundingBoxChanges() && !this._bboxFilter) {
+      throw new Error('Cannot sync on bounding box changes. There is no bounding box filter.');
```
alonsogarciapablo: This will always be present in the context of BUILDER right? I guess that's not the case when using this through the API.

test/spec/api/v4/dataview/formula.spec.js
```diff
@@ -0,0 +1,207 @@
...
+    it('if operation is valid, it assigns it to property, returns this and nothing else if there is no internaModel', function () {
+      var returnedObject = dataview.setOperation(carto.operation.AVG);
+
+      expect(dataview.getOperation()).toEqual(carto.operation.AVG);
+      expect(returnedObject).toBe(dataview);
+    });
+
+    it('sets operation in internal model if exists', function () {
+      var internalModelMock = createInternalModelMock();
+      dataview._internalModel = internalModelMock;
```
alonsogarciapablo: Perhaps `$getInternalModel` is a bit more "stable" than `_internalModel`. We could `spyOn(dataview, '$getInternalModel').and.returnValue(internalModelMock)`. This will raise an error if `Base.prototype.$getInternalModel` is ever deleted.

alonsogarciapablo: Another 💡 .Perhaps we could have `Base.prototype.$setInternalModel` and use it both internally and in our tests.

src/windshaft/filters/bounding-box.js
```diff
@@ -1,8 +1,29 @@
...
+  _initBinds: function () {
+    this.listenTo(this._map, 'change:view_bounds_ne change:center change:zoom', _.debounce(this._boundsChanged, BOUNDING_BOX_FILTER_WAIT));
+  },
+
+  _stopBinds: function () {
+    this.stopListening(this._map, 'change:view_bounds_ne change:center change:zoom');
+  },
+
+  _boundsChanged: function () {
+    this.setBounds(this._map.getViewBounds());
```
IagoLast: we should tests the integration between the map and the boundFilter to prevent bugs if the `map.getViewBounds()` is changed.

src/api/v4/constants.js
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * Constants module for dataviews
+ */
+
+/**
+ * Enum for operation values.
+ *
+ * @enum {string} carto.operation
```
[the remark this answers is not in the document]
Jesus89: We have decided to put the operations in the top level

src/api/v4/dataview/formula.js
Jesus89: Looks good to me

src/dataviews/dataview-model-base.js
ivanmalagon: Yes. `dataviews-factory` is the responsible for creating dataviews and it's providing the bboxFilter from the very beginning.

I opted to throw an Error in case the users want the dataview to be synced but there are no bboxFilter, since they won't get what they expect.

ivanmalagon:
@alonsogarciapablo I don't feel like releasing only the bboxFilter changes. Let's do it after the integration phase of all dataview changes.

test/spec/api/v4/dataview/formula.spec.js
ivanmalagon: In dataviews there is no need of adding `$setInternalModel`. I would avoid it, allowing creating the `_internalModel` only through the `_createInternalModel` function. I would avoid exposing more functions and giving the impressions that `$setInternalModel` is a common operation in dataviews. So far, there's only a use case that is using it in specs (that are under control).

`$getInternalModel` is not used in specs. I mean, we don't read the internal model directly in specs.

src/api/v4/dataview/base.js
```diff
@@ -0,0 +1,234 @@
...
+ *
+ * @return {boolean}
+ * @api
+ */
+Base.prototype.hasError = function () {
+  return this._status === status.ERROR;
+};
+
+/**
+ * Enable the dataview
```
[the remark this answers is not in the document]
ivanmalagon: Done.

src/api/v4/dataview/base.js
ivanmalagon: Done

PR #1860, opened 2017-10-31
fix regression for static legends

rjimenezda:
This PR fixes a regression introduced recently that breaks static legends (custom, torque, custom-choropleth...etc)

Static legends never go into 'loading' state, this was prevented before with a base class that overrode the handler for vis:reload. Because there were no tests, it went unnoticed with the recent refactor.

I added a small spec just to be safe in the future 🤞

### DIFF
--- a/test/spec/geo/map/legends/static-legend-model.spec.js
+++ b/test/spec/geo/map/legends/static-legend-model.spec.js
@@ -3,7 +3,7 @@ var MockFactory = require('../../../../helpers/mockFactory');
 var Engine = require('../../../../../src/engine');
 var MyLegendModel = StaticLegendModelBase.extend({ TYPE: 'type' });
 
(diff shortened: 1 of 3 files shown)


test/spec/geo/map/legends/static-legend-model.spec.js
```diff
@@ -0,0 +1,20 @@
+var StaticLegendModelBase = require('../../../../../src/geo/map/legends/static-legend-model-base');
+var MockFactory = require('../../../../helpers/mockFactory');
+var Engine = require('../../../../../src/engine');
+var MyLegendModel = StaticLegendModelBase.extend({ TYPE: 'type' });
+
+fdescribe('src/geo/map/legends/static-legend-model-base', function () {
```
alonsogarciapablo: `fdescribe`

PR #1866, opened 2017-11-06
ShowLimitsErrors param [DNM]

matallo:
- add Tiles overlay, and extend Limits from it
- don't show overlay if param is not passed

tests pending


matallo:
done in https://github.com/CartoDB/cartodb.js/pull/1875 closing then

PR #1870, opened 2017-11-06
Public histogram dataview on API v4

ivanmalagon:
https://github.com/CartoDB/cartodb.js/issues/1853
----

Ready for code review


src/api/v4/dataview/category.js
```diff
@@ -10,7 +10,7 @@ var CategoryFilter = require('../../../windshaft/filters/category');
  * @param {carto.source.Base} source - The source where the dataview will fetch the data.
  * @param {string} column - The column name to get the data.
  * @param {object} options
- * @param {number} options.limit - The maximum number of categories in the response.
+ * @param {number} [options.limit=6] - The maximum number of categories in the response.
```
IagoLast: is this parameter optional?

src/api/v4/dataview/histogram.js
```diff
@@ -0,0 +1,186 @@
...
+ * 
+ * @return {number} Current number of bins
+ * @api 
+ */
+Histogram.prototype.getBins = function () {
+  return this._bins;
+};
+
+/**
+ * Return the distribution type of the current data according to Galtung’s AJUS System http://druedin.com/2012/12/08/galtungs-ajus-system/
```
alonsogarciapablo: Isn't there a more "official" website to point to (eg: an entry in the wikipedia)?

src/api/v4/dataview/histogram.js
```diff
@@ -0,0 +1,186 @@
...
+     * @property {number} freq
+     * @property {number} normalized
+     * @api
+     */
+    /**
+     * @typedef {object} HistogramData
+     * @property {number} nulls
+     * @property {number} totalAmount
+     * @property {BinItem[]} result
+     * @property {string} type - Constant 'histogram'
```
alonsogarciapablo: Do we need a type?

src/api/v4/dataview/histogram.js
```diff
@@ -0,0 +1,186 @@
...
+  if (!_.isFinite(bins) || bins < 1) {
+    throw new TypeError('Bins must be a positive value.');
+  }
+};
+
+Histogram.prototype._parseData = function (data, nulls, totalAmount) {
+  if (data) {
+    /**
+     * @typedef {object} BinItem
+     * @property {number} index
```
alonsogarciapablo: We should explain these properties a bit better.

src/api/v4/dataview/histogram.js
```diff
@@ -0,0 +1,186 @@
...
+     * @property {number} min - Only appears if freq > 0
+     * @property {number} max - Only appears if freq > 0
+     * @property {number} avg - Only appears if freq > 0
+     * @property {number} freq
+     * @property {number} normalized
+     * @api
+     */
+    /**
+     * @typedef {object} HistogramData
+     * @property {number} nulls
```
alonsogarciapablo: We should explain these properties a bit better (including descriptions).

src/api/v4/dataview/histogram.js
```diff
@@ -0,0 +1,186 @@
...
+     * @property {number} totalAmount
+     * @property {BinItem[]} result
```
alonsogarciapablo: Why not call it `bins`?

src/api/v4/dataview/histogram.js
```diff
@@ -0,0 +1,186 @@
...
+};
+
+/**
+ * Return the totals data (not affected by filters)
+ *
+ * @return {HistogramData}
+ * @api
+ */
+
+Histogram.prototype.getTotalsData = function () {
```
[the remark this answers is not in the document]
ivanmalagon: Totals are the whole data not affected by filters. The grey bars in our current widgets. They serve as comparison of the current histogram view against the total one.

In our first version of public API the only difference will be one being affected by view bounding box and the other one showing totals.

src/api/v4/dataview/histogram.js
ivanmalagon: You're right. I'll link to Wikipedia.

src/api/v4/dataview/histogram.js
```diff
@@ -0,0 +1,186 @@
...
+  }
+  return null;
+};
+
+Histogram.prototype._checkOptions = function (options) {
+  if (_.isUndefined(options)) {
+    throw new TypeError('Options object to create a histogram dataview is required.');
+  }
+  this._validateBins(options.bins);
+  options.bins = Math.floor(options.bins);
```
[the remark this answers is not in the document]
ivanmalagon: 🤔 

Yeah, sure.

src/api/v4/dataview/histogram.js
```diff
@@ -0,0 +1,186 @@
...
+};
+
+Histogram.prototype._listenToInternalModelSpecificEvents = function () {
+  this.listenTo(this._internalModel, 'change:bins', this._onBinsChanged);
```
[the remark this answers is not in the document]
ivanmalagon: Just being paranoid. We're providing events to notify in the wrapper objects that the dataviews properties have changed. Usually, it'll be the API user the one who'll change it. 
This code is to prevent something changing the internal model and having the wrapped object and the internal one not synchronised.

test/spec/api/v4/dataview/histogram.spec.js
```diff
@@ -0,0 +1,367 @@
...
+
+    it('options set to the provided value', function () {
+      var dataview = new carto.dataview.Histogram(source, 'population', {
+        bins: 808
+      });
+
+      expect(dataview._bins).toEqual(808);
+    });
+
+    it('throw error if no correct operation is provided', function () {
```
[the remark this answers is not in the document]
ivanmalagon: :+1:

src/api/v4/dataview/histogram.js
ivanmalagon: Not really. Erasing it.

src/api/v4/dataview/histogram.js
ivanmalagon: Yes. I noticed that when reading the generated docs. I'll put it off to a ticket where we'll check documentation as a whole.

src/api/v4/dataview/histogram.js
ivanmalagon: Same.

src/api/v4/dataview/histogram.js
ivanmalagon: Because in the other dataviews where calling the main data `result`. Just to be consistent.

src/api/v4/dataview/histogram.js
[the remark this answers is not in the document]
IagoLast: what about `getUnfiltered` data ?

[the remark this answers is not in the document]
ivanmalagon: We renamed it while doing the time-series milestone because we understood it better this way. Perhaps it was because in our widgets the option to show it is `show totals` not `show unfiltered data`.

src/api/v4/dataview/histogram.js
alonsogarciapablo: IMO `result` makes sense for formula, but `categories` and `bins` make more sense to me. We can revisit this later when we review the API as a whole.

src/api/v4/dataview/histogram.js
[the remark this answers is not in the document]
alonsogarciapablo: I would personally remove this for now because:

1. The fact that BUILDER needed this doesn't mean every developer out there will need it. In case they do, let's find the best way to do it (we don't have to propagate the not so good decissions that we made internally like having dataviews with many responsibilities).

2. Wouldn't someone be able to accomplish this exact thing by using another unfiltered dataview?

PR #1871, opened 2017-11-06
Add category dataview

Jesus89:
Related to: https://github.com/CartoDB/cartodb.js/issues/1852

[WIP] Add maxCategories attribute

### DIFF
--- a/examples/v4/category-dataview.html
+++ b/examples/v4/category-dataview.html
@@ -5,7 +5,7 @@
   <meta charset="UTF-8">
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <meta http-equiv="X-UA-Compatible" content="ie=edge">
-  <title>Basic Carto.js example</title>
+  <title>Category dataview example</title>
   <link rel="stylesheet" href="https://unpkg.com/leaflet@1.2.0/dist/leaflet.css" />
   <script src="https://unpkg.com/leaflet@1.2.0/dist/leaflet.js"></script>
   <!--Include carto styles  -->
@@ -42,9 +42,9 @@
 
       // Create the Client
       var client = new carto.Client({
-          apiKey: '84fdbd587e4a942510270a48e843b4c1baa11e18',
-          serverUrl: 'https://{user}.carto.com:443',
-          username: 'cartojs-test'
+        apiKey: '84fdbd587e4a942510270a48e843b4c1baa11e18',
+        serverUrl: 'https://{user}.carto.com:443',
+        username: 'cartojs-test'
       });
 
       // Create a source
@@ -54,8 +54,8 @@
       // Create a style
       var style = new carto.style.CartoCSS(`
         #layer {
-            marker-width: 10;
-            marker-fill: yellow;
+          marker-width: 10;
+          marker-fill: yellow;
         }`
       );
 
--- a/examples/v4/category-dataview.html
+++ b/examples/v4/category-dataview.html
@@ -70,6 +70,7 @@
 
       // Add a Category dataview
       var categoryDataview = new carto.dataview.Category(source, 'adm0name', {
+        maxCategories: 10,
         operation: carto.operation.AVG,
         operationColumn: 'pop_max'
       });
@@ -92,6 +93,10 @@
 
       categoryDataview.on('statusChanged', function (newStatus, error) {
         console.log('Status changed', newStatus, error);
+      });
+
+      categoryDataview.on('maxCategoriesChanged', function (newMaxCategories) {
+        console.log('Max categories changed', newMaxCategories);
       });
 
       categoryDataview.on('operationChanged', function (newOperation) {
--- a/examples/v4/category-dataview.html
+++ b/examples/v4/category-dataview.html
@@ -41,14 +41,24 @@
       L.tileLayer('https://cartodb-basemaps-{s}.global.ssl.fastly.net/rastertiles/voyager_nolabels/{z}/{x}/{y}.png').addTo(map);
 
       // Create the Client
+      // var client = new carto.Client({
+      //   apiKey: '84fdbd587e4a942510270a48e843b4c1baa11e18',
+      //   serverUrl: 'https://{user}.carto.com:443',
+      //   username: 'cartojs-test'
+      // });
+
       var client = new carto.Client({
-        apiKey: '84fdbd587e4a942510270a48e843b4c1baa11e18',
-        serverUrl: 'https://{user}.carto.com:443',
-        username: 'cartojs-test'
+        apiKey: '4bf183ec5867bd5f9eb59d4d33693fa4fb4283bf',
+        serverUrl: 'http://{user}.localhost.lan:8181',
+        username: 'cdb'
       });
 
+      client.on('error', function(error) {
+        console.log('ERROR', error);
+      })
+
       // Create a source
-      var source = new carto.source.Dataset('ne_10m_populated_places_simple');
+      var source = new carto.source.Dataset('twitter_chiquito_calzada_chiquigif');
       window.source = source;
 
       // Create a style
@@ -69,10 +79,10 @@
       client.getLeafletLayerView().addTo(map);
 
       // Add a Category dataview
-      var categoryDataview = new carto.dataview.Category(source, 'adm0name', {
-        maxCategories: 10,
-        operation: carto.operation.AVG,
-        operationColumn: 'pop_max'
+      var categoryDataview = new carto.dataview.Category(source, 'actor_preferredusername', {
+        maxCategories: 20,
+        operation: carto.operation.SUM,
+        operationColumn: 'retweetcount'
       });
       window.categoryDataview = categoryDataview;
 
--- a/examples/v4/category-dataview.html
+++ b/examples/v4/category-dataview.html
@@ -80,7 +80,7 @@
 
       // Add a Category dataview
       var categoryDataview = new carto.dataview.Category(source, 'actor_preferredusername', {
-        maxCategories: 20,
+        limit: 20,
         operation: carto.operation.SUM,
         operationColumn: 'retweetcount'
       });
@@ -105,8 +105,8 @@
         console.log('Status changed', newStatus, error);
       });
 
-      categoryDataview.on('maxCategoriesChanged', function (newMaxCategories) {
-        console.log('Max categories changed', newMaxCategories);
+      categoryDataview.on('limitChanged', function (newMaxCategories) {
+        console.log('Limit changed', newMaxCategories);
       });
 
       categoryDataview.on('operationChanged', function (newOperation) {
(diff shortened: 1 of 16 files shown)


src/api/v4/dataview/category.js
```diff
@@ -0,0 +1,185 @@
+var _ = require('underscore');
+var Base = require('./base');
+var constants = require('../constants');
+var CategoryDataviewModel = require('../../../dataviews/category-dataview-model');
+var CategoryFilter = require('../../../windshaft/filters/category');
+
+/**
+ * Category dataview object
+ *
+ * @param {carto.source.Base} source - The source where the datavew will fetch the data.
```
ivanmalagon: It lacks an I

src/api/v4/dataview/category.js
```diff
@@ -0,0 +1,185 @@
...
+ */
+Category.prototype.setOperationColumn = function (operationColumn) {
+  this._checkOperationColumn(operationColumn);
+  this._options.operationColumn = operationColumn;
+  if (this._internalModel) {
+    this._internalModel.set('aggregation_column', operationColumn);
+  }
+  return this;
+};
+
```
ivanmalagon: We've talked about grouping `setOperation` and `setOperationColumn` for providing the user the option of setting them at once.

src/api/v4/dataview/category.js
```diff
@@ -0,0 +1,185 @@
...
+  operationColumn: 'column'
+};
+
+Category.prototype._listenToInternalModelSpecificEvents = function () {
+  this.listenTo(this._internalModel, 'change:aggregation', this._onOperationChanged);
+  this.listenTo(this._internalModel, 'change:aggregation_column', this._onOperationColumnChanged);
+};
+
+Category.prototype._onOperationChanged = function () {
+  this.trigger('operationChanged', this._options.operation);
```
ivanmalagon: If the change is not provoked by us, `this._internalModel.get('aggregation')` and `this._options.operation` will not be synchronised.
I think we should assign the local function the value that has changed from the internal model just to be sure.

src/api/v4/dataview/category.js
```diff
@@ -0,0 +1,185 @@
...
+};
+
+Category.prototype._onOperationColumnChanged = function () {
+  this.trigger('operationColumnChanged', this._options.operationColumn);
```
ivanmalagon: Same as above.

src/api/v4/dataview/category.js
Jesus89: Same with Formula category

src/api/v4/dataview/category.js
Jesus89: I go with `setOperation(operation)`, `setOperation(operation, column)` and `setOperationColumn(column)`

Jesus89: I talk to solutions and I think it is better to go with individual set/get methods for all the dataviews and then wait for the feedback of the internal release to add new group functions.

examples/v4/category-dataview.html
```diff
@@ -0,0 +1,116 @@
...
+
+      // Add a Category dataview
+      var categoryDataview = new carto.dataview.Category(source, 'adm0name', {
+        maxCategories: 10,
+        operation: carto.operation.AVG,
+        operationColumn: 'pop_max'
+      });
+      window.categoryDataview = categoryDataview;
+
+      categoryDataview.on('dataChanged', function (newData) {
```
IagoLast: should we continue adding `dummy` comments to make the example even easier to understand?

For example answering those questions:
what does those events mean?
when are they fired?

src/api/v4/dataview/category.js
```diff
@@ -0,0 +1,240 @@
...
+  if (this._internalModel) {
+    this._internalModel.set('max_categories', maxCategories);
+  }
+  return this;
+};
+
+/**
+ * Return the current dataview maxCategories
+ *
+ * @return {string} Current dataview maxCategories
```
alonsogarciapablo: `{string}` or `{number}`? Same applies to `Category.prototype.setMaxCategories`.

src/api/v4/dataview/category.js
```diff
@@ -0,0 +1,240 @@
...
+/**
+ * Set the dataview operation
+ *
+ * @param  {carto.operation} operation
+ * @return {carto.dataview.Category} this
+ * @api
+ */
+Category.prototype.setOperation = function (operation) {
+  this._checkOperation(operation);
+  this._options.operation = operation;
```
alonsogarciapablo: We store "options" as private instance variables in layers, sources, and styles. We should unify both approaches! Why did you go with `this._options`?

src/api/v4/dataview/category.js
```diff
@@ -0,0 +1,240 @@
...
+};
+
+Category.prototype.DEFAULTS = {
+  maxCategories: 5,
+  operation: constants.operation.COUNT,
+  operationColumn: 'column'
+};
+
+Category.prototype._listenToInternalModelSpecificEvents = function () {
+  this.listenTo(this._internalModel, 'change:max_categories', this._onMaxCategoriesChanged);
```
alonsogarciapablo: When would these attributes change?

src/api/v4/dataview/category.js
```diff
@@ -0,0 +1,241 @@
...
+  }
+};
+
+Category.prototype._checkOperation = function (operation) {
+  if (_.isUndefined(operation) || !constants.isValidOperation(operation)) {
+    throw new TypeError('Operation for category dataview is not valid. Use carto.operation');
+  }
+};
+
+Category.prototype._checkOperationColumn = function (operationColumn) {
```
alonsogarciapablo: Do all operations support `operationColumn`? If that's not the case... should we throw an error?

src/api/v4/dataview/category.js
```diff
@@ -0,0 +1,241 @@
...
+};
+
+Category.prototype._checkMaxCategories = function (maxCategories) {
+  if (_.isUndefined(maxCategories)) {
+    throw new TypeError('Max categories for category dataview is required.');
+  }
+  if (!_.isNumber(maxCategories)) {
+    throw new TypeError('Max categories for category dataview must be a number.');
+  }
+  if (maxCategories <= 0) {
```
alonsogarciapablo: Do we also want to set a limit so that developers cannot retrieve hundreds of categories?

test/spec/api/v4/dataview/category.spec.js
```diff
@@ -0,0 +1,290 @@
...
+  var engine = {
+    name: 'Engine mock',
+    reload: function () {}
+  };
+  spyOn(engine, 'reload');
+
+  return engine;
+}
+
+describe('category dataview public v4 API', function () {
```
alonsogarciapablo: We normally use the file path here. eg: `describe('src/api/v4/dataview/category', ...)`.

src/api/v4/dataview/category.js
Jesus89: :+1:

test/spec/api/v4/dataview/category.spec.js
alonsogarciapablo: (it makes filtering specs easier)

src/api/v4/dataview/category.js
Jesus89: The upper limit must be in the server.

test/spec/api/v4/dataview/category.spec.js
Jesus89: Ok

src/api/v4/dataview/category.js
```diff
@@ -0,0 +1,241 @@
...
+     */
+    /**
+     * @typedef {object} CategoryData
+     * @property {number} count
+     * @property {number} max
+     * @property {number} min
+     * @property {number} nulls
+     * @property {string} operation
+     * @property {CategoryItem[]} result
+     * @property {string} type - Constant 'category'
```
alonsogarciapablo: Do we need/want to expose this? I'd say we don't... In case we do, should we also expose a constant?

src/api/v4/dataview/category.js
ivanmalagon: I think it's because we set the variable `this._options` in Base. But we can then assign those options to variable instances.

So, OK, let's go with separated variable instances.

src/dataviews/category-dataview-model.js
```diff
@@ -17,14 +17,16 @@ module.exports = DataviewModelBase.extend({
     {
       type: 'category',
       filterEnabled: false,
+      categories: 6,
```
alonsogarciapablo: Perhaps `limit` is a more descriptive word for this?

src/api/v4/dataview/category.js
```diff
@@ -0,0 +1,241 @@
...
+var CategoryDataviewModel = require('../../../dataviews/category-dataview-model');
+var CategoryFilter = require('../../../windshaft/filters/category');
+
+/**
+ * Category dataview object
+ *
+ * @param {carto.source.Base} source - The source where the dataview will fetch the data.
+ * @param {string} column - The column name to get the data.
+ * @param {object} options
+ * @param {number} options.maxCategories - The maximum number of categories in the response.
```
alonsogarciapablo: `options.limit` might also work here.

src/dataviews/category-dataview-model.js
Jesus89: Categories is the word used in the server url

src/api/v4/dataview/category.js
ivanmalagon: All operations can have an operationColumn.

Depending on the column type, it can throw an error, but it's Maps API the one that throws it.

test/spec/api/v4/dataview/category.spec.js
ivanmalagon: Ooops! I did that also 😞

src/api/v4/dataview/category.js
ivanmalagon: Do you mean the `type` constants or the whole object documentation?

src/api/v4/dataview/category.js
ivanmalagon: :+1:

Jesus89: limit is other option for maxCategories, yes

src/api/v4/dataview/category.js
Jesus89: Nop :)

ivanmalagon:
🇯🇲

src/dataviews/category-dataview-model.js
alonsogarciapablo: That doesn't mean we have to use the same word 😄

Jesus89: Yes. We use now `limit` for the public API. But this file is for the internal model.

alonsogarciapablo: I still find `limit` a better name even for the internal model `dataview.get('categories')` is a bit generic, don't you think?

Issue #1872, opened 2017-11-07
Show map overlay for torque limit errors


rubenmoya:
Show map overlay when torque requests throw a `429` error.

![screen shot 2017-08-31 at 11 10 35](https://user-images.githubusercontent.com/905225/29915755-0b58e372-8e3d-11e7-93d3-4f4ce5fe58b7.png)

rubenmoya:
Closed via #1873

PR #1873, opened 2017-11-07
Add limits overlay to torque layers

rubenmoya:
Related to: https://github.com/CartoDB/cartodb.js/issues/1872
This PR depends on https://github.com/CartoDB/torque/pull/288

### DIFF
--- a/src/api/create-vis.js
+++ b/src/api/create-vis.js
@@ -12,6 +12,7 @@ var DEFAULT_OPTIONS = {
   tooltip: true, // TODO: it seems that this is no longer used
   logo: true,
   show_empty_infowindow_fields: false,
+  showLimitErrors: false,
   interactiveFeatures: false
 };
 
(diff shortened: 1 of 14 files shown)


rubenmoya:
Tests pending

matallo:
7f5

src/api/create-vis.js
```diff
@@ -12,6 +12,7 @@ var DEFAULT_OPTIONS = {
   tooltip: true, // TODO: it seems that this is no longer used
   logo: true,
   show_empty_infowindow_fields: false,
+  showLimitErrors: true,
```
matallo: by default this should be `false`

rubenmoya: yep, my bad, I was using it for testing purposes, let me fix it!

src/geo/leaflet/leaflet-torque-layer-view.js
```diff
@@ -5,7 +5,7 @@ var LeafletLayerView = require('./leaflet-layer-view');
 var TorqueLayerViewBase = require('../torque-layer-view-base');
 var util = require('cdb.core.util');
 
-var LeafletTorqueLayer = function (layerModel, leafletMap, mapModel) {
+var LeafletTorqueLayer = function (layerModel, leafletMap, mapModel, showLimitErrors) {
```
matallo: do we really need this argument here? in the end we're making use of `this.showLimitErrors`

rubenmoya: Yes, because in the next line, we have to pass `showLimitErrors` to `LeafletLayerView`

```
LeafletLayerView.apply(this, arguments);
```

PR #1877, opened 2017-11-09
Create HistogramData Object

IagoLast:
This PR moves the HistogramData to its custom object so the `histogram-dataview` file is more readable.

Also adds a small code-refactoring issues.

### DIFF
--- a/src/api/v4/dataview/histogram/histogram-data.js
+++ b/src/api/v4/dataview/histogram/histogram-data.js
@@ -1,0 +1,45 @@
+var _ = require('underscore');
+
+/**
+ * @description
+ * #Object containing histogram data.
+ * 
+ * @typedef {object} HistogramData
+ * @property {number} nulls - The number of items with null value.
+ * @property {number} totalAmount - The number of elements returned.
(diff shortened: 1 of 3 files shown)


src/api/v4/dataview/histogram/histogram-data.js
```diff
@@ -0,0 +1,45 @@
...
+ * #Object containing histogram data.
+ * 
+ * @typedef {object} HistogramData
+ * @property {number} nulls - The number of items with null value.
+ * @property {number} totalAmount - The number of elements returned.
+ * @property {BinItem[]} result - Array containing the {@link BinItem|data bins} for the histogram. .
+ * @property {string} type - String with value: **histogram**
+ * @api
+ */
+function HistogramData (data, nulls, totalAmount) {
```
ivanmalagon: Two things:
1. `HistogramData` is not an object but a function that parses data from one format to another. I'd call the function `parseHistogramData` and document the object `histogramData` as returned but this function.
2. In the same spirit as above, `HistogramData` seems to be used as a constructor by its name but it's not.

Great PR. It improves reading hands down.

Issue #1869, opened 2017-11-06
Error handling


alonsogarciapablo:
- [x] Argument errors (eg: wrong or missing params for API method)
  - [x] Layers
  - [x] Dataviews
  - [x] Sources (Dataset / SQL)
  - [x] Style
- [x] Define a `CartoError` object (only has a message property for now).
- [x] `carto.Client` triggers `error` (with `CartoError`) for:
  - [x] Global Errors (instantiation errors)

IagoLast:
To expose windshaft errors through the client we must change how the engine works.

IagoLast:
### Error list

- [ ] Client initialized with no api key

alonsogarciapablo:
- Methods that trigger an engine reload return promises (Client, Layer, Source, Dataviews). Promises are rejected with a `CartoError`
- Individual objects (Layer, Source, Dataviews) trigger an `error` message with a `CartoError` as argument.
- `carto.Client` triggers `error` event with `CartoError`.
- `CartoError` only has a message property for now.

Issue #20, opened 2012-12-27
Map is fullscreen even when setting the size of the div


tjwebb:
I cannot get the map to _not_ be fullscreen!

Example here:
http://dev.nbtsolutions.net/~tjwebb/gwi/map.html

I'm doing a simple embed like so: `cartodb.createVis('map', 'http://nbt.cartodb.com/api/v1/viz/15055/viz.json');`

and I'm setting the size of `map` like so:

```
#map {
  width: 600px;
  height: 500px;
}
```

Ideas?

PR #1882, opened 2017-11-10
1869 error handling

IagoLast:
#1869
----

- Client throws cartoErrors wrapped from engine errors.
- `.addLayer` and `.addDataview` return a promise, if those promises are rejected the parameter is a CartoError
- Add an error example with a code snippet and description.
- Update carto.events doc, use a custom namespace with constants instead enum

[https://gyazo.com/cb4c6c96d4c9f31c7e1dc87e6f2ff5dd](https://gyazo.com/cb4c6c96d4c9f31c7e1dc87e6f2ff5dd)

### DIFF
--- a/examples/v4/category-dataview.html
+++ b/examples/v4/category-dataview.html
@@ -35,15 +35,10 @@
       L.tileLayer('https://cartodb-basemaps-{s}.global.ssl.fastly.net/rastertiles/voyager_nolabels/{z}/{x}/{y}.png').addTo(map);
 
       // Create the Client
-      // var client = new carto.Client({
-      //   apiKey: '84fdbd587e4a942510270a48e843b4c1baa11e18',
-      //   serverUrl: 'https://{user}.carto.com:443',
-      //   username: 'cartojs-test'
-      // });
 
       var client = new carto.Client({
         apiKey: '4bf183ec5867bd5f9eb59d4d33693fa4fb4283bf',
-        serverUrl: 'http://{user}.localhost.lan:8181',
+        serverUrl: 'http://cdb.localhost.lan:8181',
         username: 'cdb'
       });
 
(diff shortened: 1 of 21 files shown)


src/api/v4/error.js
```diff
@@ -0,0 +1,59 @@
+/**
+ * Build a cartoError from a generic error
+ * @constructor
+ * 
+ * @return {CartoError} A well formed object representing the error.
+ */
+function CartoError (error) {
+  if (!error) {
```
alonsogarciapablo: I think we can get rid of this if (we're also assigning 'unexpected error' to `this.message` in line #20)

src/api/v4/error.js
```diff
@@ -0,0 +1,59 @@
...
+  return error && error.errors_with_context;
+}
+
+module.exports = CartoError;
+
+/**
+ * 
+ * Represents an error in the carto library.
+ * 
+ * Some actions like adding a layer to a map will trigger a **reload cycle**
```
alonsogarciapablo: "reload cycle" is not very clear to me. I'd try to find something more meaningful for developers.

IagoLast: umm, any suggestion?

I would try to explain how carto.js works (including the reload-cycle) to developers in order to make it easier to understand

examples/v4/category-dataview.html
```diff
@@ -35,15 +35,10 @@
...
-      // var client = new carto.Client({
-      //   apiKey: '84fdbd587e4a942510270a48e843b4c1baa11e18',
-      //   serverUrl: 'https://{user}.carto.com:443',
-      //   username: 'cartojs-test'
-      // });
 
       var client = new carto.Client({
         apiKey: '4bf183ec5867bd5f9eb59d4d33693fa4fb4283bf',
-        serverUrl: 'http://{user}.localhost.lan:8181',
+        serverUrl: 'http://cdb.localhost.lan:8181',
```
IagoLast: use carto-js-test

PR #1886, opened 2017-11-10
v4 Time series dataview

ivanmalagon:
https://github.com/CartoDB/cartodb.js/issues/1854
----

Ready!


examples/v4/timeseries-dataview.html
```diff
@@ -0,0 +1,158 @@
...
+      var map = L.map('map').setView([40.4169, -3.7035], 11);
+      L.tileLayer('https://cartodb-basemaps-{s}.global.ssl.fastly.net/rastertiles/voyager_nolabels/{z}/{x}/{y}.png').addTo(map);
+
+      // Create the Client
+      var client = new carto.Client({
+        apiKey: '84fdbd587e4a942510270a48e843b4c1baa11e18',
+        username: 'cartojs-test'
+      });
+
+      client.on(carto.events.SUCCESS, function () {
```
alonsogarciapablo: I would remove this from this example.

src/api/v4/dataview/time-series/index.js
```diff
@@ -0,0 +1,177 @@
...
+ * @return {carto.dataview.TimeSeries} this
+ * @api
+ */
+TimeSeries.prototype.setOffset = function (offset) {
+  this._validateOffset(offset);
+  var prevOffset = this._offset;
+  this._offset = offset;
+  if (this._internalModel) {
+    this._internalModel.set('offset', hoursToSeconds(offset));
+  } else if (prevOffset !== offset) {
```
alonsogarciapablo: Interesting! Not sure we're doing this elsewhere but its a good thing! cc: @IagoLast @Jesus89

src/api/v4/dataview/time-series/index.js
```diff
@@ -0,0 +1,177 @@
...
+};
+
+/**
+ * Set the local timezone flag. If enabled, the time offset is overriden by the use local timezone
+ * 
+ * @param {boolean} localTimezone
+ * @return {carto.dataview.TimeSeries} this
+ * @api
+ */
+TimeSeries.prototype.setLocalTimezone = function (localTimezone) {
```
alonsogarciapablo: Boolean params are always a smell to me... I would split this into more semantic methods (`enableLocalTimezone` | `disableLocalTimeZone` or something like `useLocalTimezone` and `useServerTimezone`...).

alonsogarciapablo: Also, how likely do you think this will be changed on runtime?

src/api/v4/dataview/time-series/index.js
```diff
@@ -0,0 +1,177 @@
...
+  this._internalModel = new HistogramDataviewModel({
+    source: this._source.$getInternalModel(),
+    column: this._column,
+    aggregation: this._aggregation,
+    offset: hoursToSeconds(this._offset),
+    localTimezone: this._localTimezone,
+    sync_on_data_change: true,
+    sync_on_bbox_change: !!this._boundingBoxFilter,
+    enabled: this._enabled,
+    column_type: 'date'
```
alonsogarciapablo: Would be nice to have a constant for Dataview column types instead of using 'date'...

examples/v4/timeseries-dataview.html
```diff
@@ -0,0 +1,158 @@
...
+      timeSeries.on('statusChanged', function (newStatus, error) {
+        console.log('Status changed', newStatus, error);
+        if (newStatus === carto.dataview.status.LOADING) {
+          console.log('loading');
+        }
+        if (newStatus === carto.dataview.status.LOADED) {
+          console.log('loaded');
+        }
+        if (newStatus === carto.dataview.status.ERROR) {
+          console.log('error: ', error);
```
alonsogarciapablo: Not sure why but example is throwing some errors:

screen shot 2017-11-15 at 11 59 37

alonsogarciapablo: I think it has something to do with requests from the dataview that are being cancelled when panning the map.

src/api/v4/dataview/histogram/parse-histogram-data.js
```diff
@@ -28,7 +32,7 @@ function parseHistogramData (data, nulls, totalAmount) {
    * @api
    */
   return {
-    result: _createResult(data, maxFreq),
+    result: _createResult(compactData, maxFreq),
```
Jesus89: Rename to `bins`

src/api/v4/dataview/time-series/parse-time-series-data.js
```diff
@@ -0,0 +1,74 @@
...
+   * 
+   * @typedef {object} TimeSeriesData
+   * @property {number} nulls - The number of items with null value.
+   * @property {number} totalAmount - The number of elements returned.
+   * @property {number} offset - The time offset in hours. Needed to format UTC timestamps into the proper timezone format.
+   * @property {TimeSeriesBinItem[]} result - Array containing the {@link TimeSeriesBinItem|data bins} for the time series.
+   * @api
+   */
+  return {
+    result: _createResult(compactData, maxFreq),
```
Jesus89: Rename to `bins`

examples/v4/timeseries-dataview.html
ivanmalagon: That shouldn't happen since `aborted` requests should not throw erros. I'll take a look.

src/api/v4/dataview/time-series/index.js
ivanmalagon: This is changed on runtime in embed views currently... so.. yes, it's probable. I'll change the name.

src/api/v4/dataview/time-series/index.js
ivanmalagon: Well, this is internal to be able to use `histogram` dataview as if it was two separate entities. Not sure if it's worthy in this case.

src/api/v4/dataview/histogram/parse-histogram-data.js
ivanmalagon: Better if we rename the dataview parameter to `numberOfBins`

src/api/v4/dataview/time-series/parse-time-series-data.js
ivanmalagon: Same.

src/api/v4/dataview/histogram/parse-histogram-data.js
ivanmalagon: Forget about it. That should change in `histogram` dataview, not here.

examples/v4/timeseries-dataview.html
ivanmalagon: Got it. We're subscribed to `error` events in the internal model. Backbone always trigger an `error` event when an Ajax request fails.
In the issue `error handling 2` we've fixed this so remember to get back to test this example once it's merged.

PR #1888, opened 2017-11-13
Use cdn_url in the server response wrapper

IagoLast:
Related to: https://github.com/CartoDB/cartodb-platform/issues/3709
---

The mock server response used in the unit tests was wrong so we couldn't catch this error.

This PR:

- Fixes the unit tests using a mocked server response where cdn_url is not inside the metadata object.
- Fixes the ResponseWrapper to use  `serverResponse.cdn_url` inside `serverResponse.metadata. cdn_url`


alonsogarciapablo:
🚀 🚀 🚀

Issue #1891, opened 2017-11-13
Error handling 2


alonsogarciapablo:
Api objects should trigger events (inherit from backbone.Events) and the api user can suscribe to them allowing error handling.

- [x] CartoCss throws `CartoError` when invalid.
- [x] Source throws `CartoError` when invalid.
- [x] Dataview throws `CartoError` when invalid.

PR #1893, opened 2017-11-13
1891 error handling 2

IagoLast:
https://github.com/CartoDB/cartodb.js/issues/1891
---


src/api/v4/error.js
```diff
@@ -13,13 +13,23 @@ function CartoError (error) {
...
+
   this.message = 'unexpected error';
 }
 
 function _isWindshaftError (error) {
   return error && error.errors_with_context;
 }
 
+function _handleAjaxResponse (error) {
+  var dataviewError = JSON.parse(error.responseText);
```
alonsogarciapablo: I would make this a bit more generic (not mentioning dataviews). Also, we are assuming `error.responseText` is a JSON. Perhaps we can add a try...catch around `JSON.parse` and make sure `response.errors` is there and is an array, otherwise return a generic error.

src/api/v4/layer/layer.js
```diff
@@ -275,6 +283,13 @@ function _checkSource (source) {
     throw new TypeError('The given object is not a valid source. See "carto.source.Base"');
   }
 }
+
+/**
+ * Return true when a windshaft error is because a styling error.
+ */
+function _isStyleError (windshaftError) {
+  return windshaftError.message && windshaftError.message.indexOf('style') >= 0;
```
alonsogarciapablo: If they style error messages start with `style` we could check that `windshaftError.message.indexOf('style') === 0` to be more accurate.

src/api/v4/source/dataset.js
```diff
@@ -33,14 +33,18 @@ Dataset.prototype = Object.create(Base.prototype);
...
+  var internalModel = new AnalysisModel({
     id: this.getId(),
     type: 'source',
     query: 'SELECT * from ' + this._dataset
   }, {
     camshaftReference: CamshaftReference,
     engine: engine
   });
+
+  internalModel.on('change:error', this._triggerError, this);
```
alonsogarciapablo: Why not move this up to the base class so that we dont have to repeate it in `src/api/v4/source/sql.js`?

PR #1898, opened 2017-11-15
Automatically deploy dist to dist branch

Jesus89:
Every commit in `public-api` (it could be changed) generates automatically the distribution files and @Cartofante commits the changes to the `dist` branch.

Dist branch: https://github.com/CartoDB/cartodb.js/tree/dist.
CDN: https://cdn.rawgit.com/CartoDB/cartodb.js/dist/public/carto.js


scripts/deploy-to-dist.sh
```diff
@@ -0,0 +1,49 @@
...
+  exit 0
+fi
+
+echo "Starting deployment"
+echo "Target: dist branch"
+
+DIST_DIR="dist"
+TMP_DIST_DIR="tmp_dist"
+PUBLIC_DIR="public"
+INTERNAL_DIR="internal"
```
alonsogarciapablo: There is no need to "upload" the internal dist files to this branch IMO.

Jesus89: This can be useful for the Builder in the future

Jesus89: But I can remove it for now

Jesus89:
I think we can use this now. I would leave the versions for the public (next) release

PR #1905, opened 2017-11-17
1890 jsdoc

IagoLast:


src/api/v4/dataview/base.js
```diff
@@ -328,3 +330,23 @@ Base.prototype.$getInternalModel = function () {
 };
 
 module.exports = Base;
+
+/**
+ * Event triggered when the column in a dataview changes.
+ *
+ * Contains a single argument with the name of the changed column.
+ * 
+ * @event carto.dataview.Base.columnChanged
```
alonsogarciapablo: From what I understand, [@event](http://usejsdoc.org/tags-event.html) is something you should use to document the object associated to an event, and not the event type itself. I might be wrong, but [here is how I used it](https://github.com/CartoDB/cartodb.js/blob/public-api/src/api/v4/leaflet/layer-group.js#L73-L82).

Jesus89:
I think we need to use the @fires tag in the functions that triggers the events.

src/api/v4/dataview/base.js
IagoLast: I based my doc on [this example](http://usejsdoc.org/tags-event.html):

> A typical event is represented by an object with a defined set of properties.

So `carto.dataview.Base.columnChanged` is an event which is only a string with the column name.

IagoLast:
@Jesus89 maybe you are right, but I want to emphasise the fact that events can arrive asynchronously, so is the object the one throwing events.

A client can throw a `success` without explicitly  calling `client.reload()`

src/api/v4/dataview/base.js
alonsogarciapablo:
```
  /**
     * Snowball event.
     *
     * @event Hurl#snowball
     * @type {object}
     * @property {boolean} isPacked - Indicates whether the snowball is tightly packed.
     */
    this.emit('snowball', {
        isPacked: this._snowball.isPacked
    });
```

`@event` is NOT documenting "snowball" (name of the event, like `columnChanged`). It's documenting `{ isPacked: xxx }`. I don't think we want `carto.dataview.Base.columnChanged` as a menu item in the documentation.

alonsogarciapablo: Actually, perhaps we might en up having a single `ChangeEvent` for all `xxxChanged` events, don't you think?

PR #1923, opened 2017-11-22
Update semver version in package.json

amartinj:
Update semver version

Fixes [#159](https://github.com/CartoDB/bug-bounty/issues/159)

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
     "leaflet": "1.0.3",
     "load-grunt-tasks": "~0.6.0",
     "npm-watch": "^0.3.0",
-    "semver": "~4.1.0",
+    "semver": "~4.3.0",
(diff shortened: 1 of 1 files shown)


Jesus89:
Good moment to update semver to `5.4.1`?

amartinj:
@Jesus89 as good as perfect. I wasn't sure it would cause trouble upgrading to the last, but... I guess it won't

Jesus89:
Semver is only used here:

- https://github.com/CartoDB/cartodb.js/blob/1e770ede4d2764181104322d4d7e5ba9ce393455/grunt/tasks/prompt.js#L7
- https://github.com/CartoDB/cartodb.js/blob/1e770ede4d2764181104322d4d7e5ba9ce393455/Gruntfile.js#L11

And the API seems to be the same so :es:

PR #1928, opened 2017-11-23
13089 remove vizjson auth tokens fetch

gfiorav:
CartoDB.js part of fix for issue https://github.com/CartoDB/cartodb/issues/13089

In the current develop version of this repo, `auth_tokens` are assumed to come in the `vizjson`. This is no longer the case, so this PR modifies `image.js` to load those from options.

There's also a modification to a spec to check the new behaviour.


xavijam:
cc @alonsogarciapablo

test/spec/core/image.spec.js
```diff
@@ -512,7 +514,8 @@ describe("Image", function() {
 
     };
 
-    var image = cartodb.Image(vizjson).size(400, 300);
+    var options = { auth_tokens: ["e900fe76cc3c1eed4fc018d027d82c8b0e59b2c484d1941954f34b4818a5d660"] };
```
alonsogarciapablo: What is `e900fe76cc3c1eed4fc018d027d82c8b0e59b2c484d1941954f34b4818a5d660`?

alonsogarciapablo: Do we need this variable?

alonsogarciapablo:
I am fixing the tests once and for all!

test/spec/core/image.spec.js
gfiorav: I just moved it from the vizjson stub, it's the token that will later be expected

alonsogarciapablo:
🇪🇸

gfiorav:
@alonsogarciapablo how should we deploy this?

src/vis/image.js
```diff
@@ -85,6 +85,8 @@
 
       this.no_cdn = options.no_cdn;
 
+      this.auth_tokens = options.auth_tokens;
```
gfiorav: @alonsogarciapablo do you think we should guard against null here?

```js
this.auth_tokens = options.auth_tokens || [];
```

Maybe this will break when no auth_tokens are passed in, and later down the line some array operation is performed on null.

xavijam:
IMHO, and sorry to be late for the party, I'd follow the release process but not deploy to our CDN. But this is your call guys.

alonsogarciapablo:
I'm fine with that @xavijam suggested. Make sure you add an entry to NEWS.md to keep track of this change. Then you will just need to merge this and bump the git submodule in github.com/cartodb/cartodb. You can ask for help on this but you will basically have to get into the `lib/assets/javascripts/cdb` and checkout the latest commit. Then commit the whole thing and deploy.

xavijam:
👏

Issue #1918, opened 2017-11-21
Hovering over feature only changes cursor for top layer


javitonino:
**STR**
In this example map (http://bl.ocks.org/javitonino/663c24f88a091bd850804240f50fa6549) I have created two layers with hover interactivity. In the POIs layer (iconized), if you hover over the icon, the cursor changes to a hand. In the points (circles) layer, if you hover, the cursor doesn't change (or it flickers).

If you reorder the layers, the behaviour changes. It's not related to layer style, but rather to position. Only the top layer works correctlt.

IagoLast:
For some reason the events are fired twice.

In the buggy layer we got a `featureOver` followed by a `featureOut`, so the final cursor is the default one.

[![https://gyazo.com/fa6346922bc5ad6f5234b199b0ff14f9](https://i.gyazo.com/fa6346922bc5ad6f5234b199b0ff14f9.gif)](https://gyazo.com/fa6346922bc5ad6f5234b199b0ff14f9)


I'll keep investigating...

alonsogarciapablo:
Probably each layer fires an event and they contradict each other (one fires `featureOver` and the other one `featureOut`). @IagoLast also think that will have to implement something similar for Google Maps soon.

IagoLast:
@alonsogarciapablo wow, good catch! is this a carto.js v4 problem? or is a deeper one?

alonsogarciapablo:
I think it's basically related to how wax.js works and how we're using it. We basically have two layers and two different wax "interactions" (one per layer). That's the reason why we need to check to check if any feature of any layer is being "feauture-overed". [This is how we solve this internally](https://github.com/CartoDB/cartodb.js/blob/v4/src/vis/map-cursor-manager.js).

PR #1929, opened 2017-11-23
Update mouse interactivity for v4

IagoLast:
https://github.com/CartoDB/cartodb.js/issues/1918


src/api/v4/leaflet/layer-group.js
```diff
@@ -84,4 +92,23 @@ LayerGroup.prototype._triggerLayerFeatureEvent = function (eventName, internalEv
...
+}
+
+/** 
+ * Check if a layer is interactive
+ */
+function isInteractive (layer) {
+  if (!layer) {
+    return false;
+  }
+  return layer.hasFeatureClickColumns() || layer.hasFeatureOverColumns();
```
alonsogarciapablo: This looks like some [feature envy](https://sourcemaking.com/refactoring/smells/feature-envy)... Why not move this whole method to `carto.layer.Layer` and just do `layer.isInteractive`?

src/api/v4/leaflet/layer-group.js
```diff
@@ -84,4 +92,23 @@ LayerGroup.prototype._triggerLayerFeatureEvent = function (eventName, internalEv
   }
 };
 
+function anyLayerHovered (map) {
```
alonsogarciapablo: `map` is a bit misleading here... Perhaps we can just use a regular method and use `this._hoveredLayers` instead of passing `this._hoveredLayers` to a function.

src/api/v4/leaflet/layer-group.js
```diff
@@ -42,14 +44,20 @@ LayerGroup.prototype._onFeatureClick = function (internalEvent) {
 
 LayerGroup.prototype._onFeatureOver = function (internalEvent) {
   var layer = this._layers.findById(internalEvent.layer.id);
-  if (layer && (layer.hasFeatureClickColumns() || layer.hasFeatureOverColumns())) {
+  if (isInteractive(layer)) {
+    this._hoveredLayers[internalEvent.layerIndex] = true;
```
alonsogarciapablo: Working with indexes is a bit dangerous. Imagine a layer is removed...

src/api/v4/leaflet-layer.js
```diff
@@ -107,4 +115,23 @@ var LeafletLayer = L.TileLayer.extend({
   }
 });
 
+function anyLayerHovered (map) {
```
Jesus89: _.any ?

src/api/v4/leaflet/layer-group.js
Jesus89: Each element has always an unique id

Issue #1892, opened 2017-11-13
Error handling 3


alonsogarciapablo:
We should revisit error handling both internally and externally.

### Internally
- [ ] List all possible errors.
  - [ ] Generate a list with all the possible `windshaft`errors: (sources,  cartocss, auth, limits..)
  - [ ] Generate a list with all the possible `network` errors. (server down, connection lost)
  - [ ] Generate a list of potential common `runtime` errors. (not very common, maybe a malformed json?)
  - [ ] Generate a list of `validation` errors.
- [ ] Understand how BUILDER/deep-insights.js are consuming errors from CARTO.js
- [ ] Update the `cartoError` class and create a hierarchy with well defined messages.

### With this hierarchy.

* **Client**
  * Triggers ALL errors.
  * `.addLayer` return a promise with a cartoError when something goes wrong.
  * `.addLayers` return a promise with a cartoError when something goes wrong.
  * `.removeLayer` return a promise with a cartoError when something goes wrong.
  * `.removeLayers` return a promise with a cartoError when something goes wrong.
  * `.addDataview` return a promise with a cartoError when something goes wrong. 
  * `.addDataviews` return a promise with a cartoError when something goes wrong. 
  * `.removeDataview` return a promise with a cartoError when something goes wrong. 
* **Layer**
  * Triggers all `layer_related`errors.
  * `.setSource` return a promise with a cartoError when something goes wrong.
  * `.setStyle` return a promise with a cartoError when something goes wrong.
* **Source**
  * Triggers all `source` related errors.
  * **SQL**
    * `.setQuery` return a promise with a cartoError when something goes wrong.
* **CartoCSS**
  * Triggers all `cartocss`related erros.
* **Dataview**
    * Triggers all `source` related errors.
    * Triggers all errors from Windshaft requests.
    * Triggers all validation errors from dataview parameters.

IagoLast:
@ivanmalagon we need to complete the list with the async methods in the dataviews.

ivanmalagon:
Working on it

IagoLast:
@ivanmalagon IMO we should update the possible `dataview` errors first and then follow the list order .

ivanmalagon:
The `dataviews` error list is already updated. They capture and trigger an error when their sources have a problem, trigger their own validation errors (`bins` not being a number, f.i.) and triggers all errors that come from Windshaft. That's it.

ivanmalagon:
#### Client

[...]

PR #1932, opened 2017-11-24
Start error qualification

ivanmalagon:
https://github.com/CartoDB/cartodb.js/issues/1892
----

Made some changes to how errors are parsed and digested. We were swallowing Windshaft errors if they didn't have a global flag. We were not qualifying them properly as well.

Apart from that, I've created a POC of an error list that can ease categorizing errors and giving better feedback to the users.


examples/public/interactivity.html
```diff
@@ -31,8 +38,9 @@
 
     // Create the Client
     var client = new carto.Client({
-      apiKey: '84fdbd587e4a942510270a48e843b4c1baa11e18',
-      username: 'cartojs-test'
+      apiKey: 'f8428dda3b56a9f421dd408a86a283b0efcd1a6d',
+      username: 'ivan',
+      serverUrl: 'http://ivan.localhost.lan:8181'
```
Jesus89: Remove the serverUrl

examples/public/interactivity.html
```diff
@@ -15,6 +15,13 @@
     html, body { margin: 0; padding: 0; }
     #map { position: absolute; top: 0; left: 0; width: 100%; height: 100%; }
   </style>
+  <script>
+    window._trackJs = {
+      token: "cf7a7cd8411140689ea94e92d9fba842",
```
Jesus89: Is it secure to expose the trackJs token here?

examples/public/interactivity.html
```diff
@@ -45,6 +53,13 @@
         marker-fill: red;
       }`);
 
+    spanishCitiesStyle = new carto.style.CartoCSS(`
+      #layer {
+        tucutu: 10;
```
IagoLast: i would use `invalid-property` in order to be more explicit

src/api/v4/client.js
```diff
@@ -1,12 +1,17 @@
...
 var Backbone = require('backbone');
-var CartoError = require('./error');
+var CartoError = require('./error-handling/carto-error');
 var Engine = require('../../engine');
 var Events = require('./events');
 var LayerBase = require('./layer/base');
 var Layers = require('./layers');
 var LeafletLayer = require('./leaflet-layer');
 var VERSION = require('../../../package.json').version;
+var CartoValidationError = require('./error-handling/carto-validation-error');
```
IagoLast: deps were sorted

src/api/v4/client.js
```diff
@@ -1,12 +1,17 @@
...
+
+function getValidationError (code) {
```
IagoLast: utility functions are defined at the end at the file

src/api/v4/client.js
```diff
@@ -275,6 +280,10 @@ Client.prototype._bindEngine = function (engine) {
   }.bind(this));
 
   engine.on(Engine.Events.RELOAD_ERROR, function (err) {
+    this.trigger(Events.ERROR, new CartoError(err, { layers: this._layers }));
+  }.bind(this));
+
+  engine.on('error', function (err) {
```
IagoLast: Use `Engine.Events.ERROR` here. Btw i would change the name to "TILE_ERROR"

src/api/v4/error-handling/carto-error.js
```diff
@@ -8,25 +24,51 @@ var UNEXPECTED_ERROR = 'unexpected error';
...
+  return friendlyError;
 }
 
+// Windshaft should have been parsed already 
 function _isWindshaftError (error) {
-  return error && error.errors_with_context;
+  return error && error.origin === 'windshaft';
+}
+
+function transformWindshaftError (error, layers, analysis) {
```
IagoLast: use underscore for private functions so we know they are safe to nuke 💥

src/api/v4/error-handling/carto-error.js
```diff
@@ -1,4 +1,20 @@
+var _ = require('underscore');
+var retrieveFriendlyError = require('./error-list');
 var UNEXPECTED_ERROR = 'unexpected error';
+var GENERIC_ORIGIN = 'generic';
+
```
IagoLast: a small description would be nice here to understand this method without reading the code

src/api/v4/error-handling/error-list.js
```diff
@@ -0,0 +1,293 @@
...
+    filter: {
+      'invalid-bounds-object': {
+        messageRegex: 'invalidBoundsObject',
+        friendlyMessage: 'Bounds object is not valid. Use a carto.filter.Bounds object'
+      }
+    }
+  }
+};
+
+function buildErrorCode (error, key) {
```
IagoLast: use underscore for private functions so we know they are safe to nuke 💥

src/engine.js
```diff
@@ -139,8 +140,8 @@ Engine.prototype.reload = function (options) {
       resolve();
     };
     options.error = function (serverResponse) {
-      oldError(serverResponse);
-      reject(serverResponse);
+      var parsedError = oldError(serverResponse);
```
IagoLast: whats `oldError`  ?

examples/public/interactivity.html
alonsogarciapablo: +1 Not sure we can/want to include a trackJs token in the library...

src/api/v4/error-handling/carto-error.js
```diff
@@ -8,25 +24,51 @@ var UNEXPECTED_ERROR = 'unexpected error';
...
+}
+
+function transformWindshaftError (error, layers, analysis) {
+  var cartoError = {
+    message: error.message,
+    type: error.type,
+    origin: error.origin
+  };
+  if (error.type === 'layer' && layers) {
+    cartoError.layer = layers.findById(error.layerId);
```
alonsogarciapablo: Instead of having .layer and .source, perhaps we could find a generic name for the resource/model that failed (like Backbone does) so that it works for both cases. Also, we should document this/these property/ies.

src/vis/vis.js
```diff
@@ -120,6 +120,7 @@ var VisModel = Backbone.Model.extend({
 
     this._engine.on(Engine.Events.RELOAD_SUCCESS, this._onEngineReloadSuccess, this);
     this._engine.on(Engine.Events.RELOAD_ERROR, this._onEngineReloadError, this);
+    this._engine.on(Engine.Events.ERROR, this._onEngineError, this);
```
alonsogarciapablo: What is the difference between `Engine.Events.RELOAD_ERROR` and `Engine.Events.Error`? Its not super clear given that names are a bit similar IMO.

src/api/v4/leaflet-layer.js
```diff
@@ -88,6 +90,11 @@ var LeafletLayer = L.TileLayer.extend({
     this._triggerLayerFeatureEvent(Layer.events.FEATURE_OUT, internalEvent);
   },
 
+  _onFeatureError: function (error) {
+    var cartoError = new CartoError(error);
+    this.fire(Layer.events.FEATURE_ERROR, cartoError);
```
alonsogarciapablo: IMO this is not really a feature error... If Maps API returns some errors when calculating the interactivity it might affect multiple features/layer, not only one feature/layer as the name implies. Perhaps we can find a better name?

This class is not firing any other CARTO events... Its the layer "model" who triggers featureClicked, featureOver, etc... Shouldn't we unify this (ie: either the model or the view fires this short of events).

Are we documenting this event?

examples/public/interactivity.html
ivanmalagon: It's the way it's shown in their documentation to add the credentials.

And it's there to check if the library was correctly tracking. We can get rid of it.

src/api/v4/error-handling/carto-error.js
ivanmalagon: In this case it's not a function of a class. It's not being exposed since it only exports a function. No need to underscore them.

src/api/v4/error-handling/carto-error.js
ivanmalagon: That's an option as long as we document the different types that can be returned.
Having separated properties avoid the user to duck type what object it is.

src/api/v4/error-handling/carto-error.js
ivanmalagon: :+1:

src/api/v4/error-handling/carto-error.js
ivanmalagon: Forget about my comment. I mistook this comment for the one in the `error-list`.

src/api/v4/error-handling/error-list.js
ivanmalagon: In this case it's not a function of a class. It's not being exposed since it only exports a function. No need to underscore them.

src/api/v4/leaflet-layer.js
ivanmalagon: The problem in this case it's that the error is not attached to any particular layer. So we have three options:
- Trigger it from the layer (as it's in this PR).
- Trigger it from the first layer that has interactivity.
- Trigger it from all the layers that have interactivity.

For me, the second one is good enough and we can refactor this code to support it. The problem will affect that first layer so it's a good place to trigger the event.

@IagoLast was about to change the name to TILE_ERROR if I'm not mistaken.

src/engine.js
ivanmalagon: `oldXXX` is the name that was already there assigned to the error and success callbacks.

src/vis/vis.js
ivanmalagon: Same as above. @IagoLast suggested TILE_ERROR. Good name.

src/api/v4/leaflet-layer.js
ivanmalagon: Any thoughts on this @alonsogarciapablo @IagoLast @Jesus89 ?

My vote if triggering from the first layer with interactivty.

alonsogarciapablo: My vote -> `Trigger it from all the layers that have interactivity.`

src/api/v4/native/leaflet-layer.js
```diff
@@ -86,6 +89,15 @@ var LeafletLayer = L.TileLayer.extend({
       this._map.getContainer().style.cursor = 'auto';
     }
     triggerLayerFeatureEvent(Layer.events.FEATURE_OUT, internalEvent, this._layers);
+  },
+
+  _onFeatureError: function (error) {
+    var cartoError = new CartoError(error);
+    _.each(this._layers.toArray(), function (layer) {
+      if (layer.isInteractive()) {
+        layer.trigger(Layer.events.TILE_ERROR, cartoError);
```
alonsogarciapablo: Is this event documented?

src/api/v4/dataview/base.js
```diff
@@ -214,19 +215,19 @@ Base.prototype._initialize = function (source, column, options) {
 
 Base.prototype._checkSource = function (source) {
   if (!(source instanceof SourceBase)) {
-    throw new TypeError('Source property is required.');
+    throw this._getValidationError('sourceRequired');
```
alonsogarciapablo: Having a key + message for each and every possible "validation" error seems like a lot of work to me. You might want to kill me, but I personally think that a TypeError + some string interpolation might just be okay in this case where a user just needs a very basic message error. Something like rails does (`validate_presence_of :source`) would be nice.

alonsogarciapablo: Basically, we could build something a bit more generic that didn't force us to define (and maintain) a key and a friendly message.

IagoLast: Is not so hard to maintain, instead throwing an error with a single string every time, just create a pair key, message.
Doing this we get lot of advantages, we could give the user an url with info to solve the error, we get better error classification, less chances of different logs for the same error..

Issue #1734, opened 2017-09-05
Expose client.getGoogleMapsMapType() as part of the API


alonsogarciapablo:
It should be similar to https://github.com/CartoDB/cartodb.js/issues/1730 but the object should be a [Custom Google Maps Overlay](https://developers.google.com/maps/documentation/javascript/customoverlays#add) that can be added to an existing Google Maps map:

```js
var mapType = client.getGoogleMapsMapType();

var map = new google.maps.Map(document.getElementById('map'), {
  zoom: 10,
  center: {lat: 41.850, lng: -87.650}
});

// Insert this overlay map type as the first overlay map type at
// position 0. Note that all overlay map types appear on top of
// their parent base map.
map.overlayMapTypes.insertAt(0, mapType);
```

Jesus89:
Now you need to pass the map to the get function because of WAX:

```js
var map = new google.maps.Map(document.getElementById('map'), {
  zoom: 10,
  center: {lat: 41.850, lng: -87.650}
});

var mapType = client.getGoogleMapsMapType(map);

map.overlayMapTypes.push(mapType);
```

PR #1937, opened 2017-11-28
Add getGoogleMapsMapType function

Jesus89:
Related to: https://github.com/CartoDB/cartodb.js/issues/1734

**NOTE**

Now you need to pass the map to the get function because of WAX:

```js
var map = new google.maps.Map(document.getElementById('map'), {
  zoom: 10,
  center: {lat: 41.850, lng: -87.650}
});

var mapType = client.getGoogleMapsMapType(map);

map.overlayMapTypes.push(mapType);
```

This could be optimized in the future when WAX is removed.

### DIFF
--- a/src/api/v4/native/google-maps-map-type.js
+++ b/src/api/v4/native/google-maps-map-type.js
@@ -1,0 +1,79 @@
+/* global google */
+var _ = require('underscore');
+var Layer = require('../layer');
+var GMapsCartoDBLayerGroupView = require('../../../geo/gmaps/gmaps-cartodb-layer-group-view');
+
+/**
+ * This object is a custom Google Maps MapType to enable feature interactivity
+ * using an internal GMapsCartoDBLayerGroupView instance.
+ *
+ */
+
+function GoogleMapsMapType (layers, engine, map) {
+  this._layers = layers;
+  this._engine = engine;
+  this._map = map;
+
+  this._hoveredLayers = [];
+
+  this.tileSize = new google.maps.Size(256, 256);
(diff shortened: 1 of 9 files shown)


src/api/v4/native/google-maps-map-type.js
```diff
@@ -0,0 +1,83 @@
+/* global google */
+var _ = require('underscore');
+var Layer = require('../layer');
+var GMapsCartoDBLayerGroupView = require('../../../geo/gmaps/gmaps-cartodb-layer-group-view');
+
+/**
+ * This object is a custom Google Maps MapType to enable feature interactivity
```
IagoLast: nice comment :D

alonsogarciapablo:
`client.getGoogleMapsMapType(map);` could probably check if `google` is defined.

alonsogarciapablo:
What about the CARTO attribution? Not sure how Google Maps handles this.

Jesus89:
@iago we could extract some logic, but for example leaflet goes with `this._map.getContainer().style.cursor = 'pointer';` and gmaps with `this._map.setOptions({ draggableCursor: 'pointer' });`. I think we could extract only the `_triggerLayerFeatureEvent` function.

@alonsogarciapablo I think it is a good idea to check `goggle` (and also `leaflet`) in the `getGoogleMapsMapType` and `getLeafletLayer` functions.

We need to check how attribution is done in gmaps :+1:

Jesus89:
It is not possible to change the attribution. So we are done!

PR #1938, opened 2017-11-29
Fix Choropleth legend with one bucket

ethervoid:
If we receive just one filter we're not showing the ramp color properly
because the `linear_gradient` css property in the template need two
values so we're going to pass the same value and legend twice

### DIFF
--- a/src/windshaft-integration/legends/rule-to-choropleth-legend-adapter.js
+++ b/src/windshaft-integration/legends/rule-to-choropleth-legend-adapter.js
@@ -12,15 +12,23 @@ var isEveryBucketValid = function (rule) {
 };
 
 var generateColors = function (buckets) {
-  return _.map(buckets, function (bucket, i) {
-    var label = '';
-    if (i === 0) {
-      label = bucket.filter.start;
-    } else if (i === buckets.length - 1) {
-      label = bucket.filter.end;
-    }
-    return { value: bucket.value, label: label.toString() };
-  });
+  if (buckets.length === 1) {
+    var bucket = buckets[0];
+    var labelStart = bucket.filter.start;
+    var labelEnd = bucket.filter.end;
+    return [{ value: bucket.value, label: labelStart.toString() },
+      { value: bucket.value, label: labelEnd.toString() }];
+  } else {
+    return _.map(buckets, function (bucket, i) {
+      var label = '';
+      if (i === 0) {
+        label = bucket.filter.start;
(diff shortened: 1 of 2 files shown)


src/windshaft-integration/legends/rule-to-choropleth-legend-adapter.js
```diff
@@ -12,15 +12,23 @@ var isEveryBucketValid = function (rule) {
...
-  return _.map(buckets, function (bucket, i) {
-    var label = '';
-    if (i === 0) {
-      label = bucket.filter.start;
-    } else if (i === buckets.length - 1) {
-      label = bucket.filter.end;
-    }
-    return { value: bucket.value, label: label.toString() };
-  });
+  if (buckets.length === 1) {
```
IagoLast: could we simplify this a little bit?

```javascript
 function generateColors (buckets) {
  if (buckets.length === 1) {
    var bucket = buckets[0];
    var labelStart = bucket.filter.start;
    var labelEnd = bucket.filter.end;
    return [{ value: bucket.value, label: labelStart.toString() }, { value: bucket.value, label: labelEnd.toString() }];
  }
  return _.map(buckets, function (bucket, i) {
    var label = '';
    if (i === 0) {
      label = bucket.filter.start;
    }
    if (i === buckets.length - 1) {
      label = bucket.filter.end;
    }
    return { value: bucket.value, label: label.toString() };
  });
}
```

ethervoid: Sure, changing

ethervoid:
@IagoLast made the change, care to review again :)

src/windshaft-integration/legends/rule-to-choropleth-legend-adapter.js
```diff
@@ -12,6 +12,12 @@ var isEveryBucketValid = function (rule) {
 };
 
 var generateColors = function (buckets) {
+  if (buckets.length === 1) {
```
alonsogarciapablo: Another (crazy?) option here...

```
if (buckets.length === 1) {
  buckets = [ buckets[0], buckets[0] ] 
}
return _.map(buckets, function (bucket, i) {
...

```

ethervoid: I'd work but IMHO is less readable

alonsogarciapablo:
👏  👏  👏

PR #1940, opened 2017-11-30
Refactor error-handling

IagoLast:


src/api/v4/error-handling/carto-error-extender.js
```diff
@@ -0,0 +1,82 @@
...
+ * unknown error.
+ * @param {CartoError} cartoError 
+ */
+function _getListedError (cartoError, errorList) {
+  var errorListkeys = _.keys(errorList);
+  var key;
+  for (var i = 0; i < errorListkeys.length; i++) {
+    key = errorListkeys[i];
+    if (errorList[key].messageRegex.test(cartoError.message)) {
+      return {
```
ivanmalagon: I personally don't like exiting the function and the loop at the same time within the loop.

src/api/v4/error-handling/carto-error-extender.js
```diff
@@ -0,0 +1,82 @@
+var _ = require('underscore');
+var ERROR_LIST = require('./error-list');
+
+/**
+ * Get two parameters to enrich a CartoError.
+ * - friendlyMessage: A easy to understand error description.
+ * - errorCode: A unique errror code
```
ivanmalagon: Arrrrrrr

src/api/v4/error-handling/carto-error-extender.js
```diff
@@ -0,0 +1,82 @@
...
+  };
+}
+
+/**
+ * Replace $0 with the proper paramter in the listedError regex to build a friendly message
+ */
+function _replaceRegex (cartoError, listedError) {
+  var match = cartoError.message && cartoError.message.match(listedError.messageRegex);
+  if (match && match.length > 1) {
+    return listedError.friendlyMessage.replace('$0', match[1]);
```
ivanmalagon: The argument `cartoError` can have an undefined `.friendlyMessage` property, throwing an error in this case.

src/api/v4/error-handling/carto-error-extender.js
```diff
@@ -0,0 +1,82 @@
...
+  }
+  return listedError.friendlyMessage;
```
ivanmalagon: Same.

If there's no `.friendlyMessage` entry we should use the original message.

src/api/v4/error-handling/carto-error.js
```diff
@@ -71,6 +65,14 @@ function transformWindshaftError (error, layers, analysis) {
   return cartoError;
 }
 
+function _transformAjaxError (error) {
```
ivanmalagon: Another personal taste complain. These functions are not "private", they're auxiliary functions that are used by the function that it's exported, but they're not shown outside this module. So there's no need of marking them as protected.

src/api/v4/layer/base.js
```diff
@@ -45,3 +45,25 @@ Base.prototype.$getInternalModel = function () {
 };
 
 module.exports = Base;
+
+function generateColors (buckets) {
```
ivanmalagon: 🤔 

Added function after `module.exports`that is not used?

src/api/v4/style/cartocss.js
```diff
@@ -1,6 +1,6 @@
 var _ = require('underscore');
 var Base = require('./base');
-var CartoValidationError = require('../error-handling/carto-error');
+var CartoValidationError = require('../error-handling/carto-validation-error');
```
ivanmalagon: :+1:

src/api/v4/error-handling/carto-error.js
IagoLast: agreed, the js scopes give us real privacity, but as long as the file grows the underscore helps to know which methods are "private".

Besides personal taste, the most of the library uses underscores, if we are removing them lets do it everywhere.

src/api/v4/error-handling/carto-error-extender.js
IagoLast: Since we know each key is unique, we avoid iterating the half of the array.

src/api/v4/error-handling/carto-error-extender.js
IagoLast: fixed

src/api/v4/error-handling/carto-error-extender.js
```diff
@@ -57,6 +57,9 @@ function _getListedError (cartoError, errorList) {
  * Replace $0 with the proper paramter in the listedError regex to build a friendly message
  */
 function _replaceRegex (cartoError, listedError) {
+  if (listedError.friendlyMessage) {
```
ivanmalagon: it should be `if NOT friendlyMessage`

Issue #1733, opened 2017-09-05
Expose metadata from CartoCSS/TurboCarto styles


alonsogarciapablo:
Developers should be able to get the metadata associated to the CartoCSS/TurboCarto that was given to the Maps API. Also, since CartoCSS/TurboCarto can be changed on runtime, developers should be able to detect when metadata of a given "style" has changed.

Here's some pseudocode:

```js
// We also want the legend to be refreshed when new metadata is available
citiesStyle.on('change:metadata', function (metadata) {
  console.log(metadata);
});
```

Jesus89:
The 'metadataChanged' event is going to be added to the Layer object:

```js
layer.on('metadataChanged', function (metadata) {
  metadata.isBubble();
  metadata.isCategory();
  metadata.isChoropleth();
  console.log(metadata);
});
```

makella:
@Jesus89 

One way to distinguish buckets and categories is by data type. Below are some ideas. 

Of course, the pairs don't have to go together in all cases, so there could be a way to mix and match.

### Value and Type
`var value = styleMetadata.getValue();`
`var type = styleMetadata.getType();`

### Quantity and Quality
`var quantity = styleMetadata.getQuantity();`
`var quality = styleMetadata.getQuality();`

### Value and Class
`var value = styleMetadata.getValue();`
`var class = styleMetadata.getClass();`
  - this could be confusing because the word `class` could also be used as a synonym for `bucket` 

### Measure and Type
`var measure = styleMetadata.getMeasure();`
`var type = styleMetadata.getType();`

### Scale and Group
`var scale = styleMetadata.getScale();`
`var group = styleMetadata.getGroup();`

### Range and Theme
`var range = styleMetadata.getRange();`
`var theme = styleMetadata.getTheme();`
- I know `range` in TurboCARTO scales a property and for creating a choropleth map we don't use `range` but it is distributing values over a range of colors.

Jesus89:
Thanks @makella. We will use a getType() function to distinguish the type of the styleMetadata.

PR #1947, opened 2017-12-05
Add Style metadata to API

Jesus89:
Related to: https://github.com/CartoDB/cartodb.js/issues/1733, https://github.com/CartoDB/cartodb.js/issues/1946
---


src/api/v4/layer/layer.js
```diff
@@ -388,4 +396,14 @@ function _isStyleError (windshaftError) {
  * @api
  */
 
+/**
+ * Event triggered when the style metadata of the layer changes.
+ *
+ * Contains a list of metadata objects.
+ *
+ * @event carto.layer.Layer.metadataChanged
+ * @type {carto.layer.Layer}
```
IagoLast: [@type ](https://github.com/CartoDB/cartodb.js/pull/1947/files#diff-8e22edeb2556633c9879fc8ae88dd8edR304) `{ styles: metadataList } ` ?

src/api/v4/layer/metadata/base.js
```diff
@@ -0,0 +1,66 @@
...
+  this._property = rule.getProperty();
+}
+
+/**
+ * Return true is the metadata type is categories
+ *
+ * @return {boolean}
+ * @api
+ */
+Base.prototype.isCategories = function () {
```
IagoLast: With this we are coupling the base class to its children.

IMO we should add an empty `.getType()` method in the base class and make each child override it returning an string.

`buckets.getType(); // 'buckets`
`categories.getType(); // 'categories`

This way we preserve the original functionality without forcing the base class to know its children, plus we get the ability to do switch sentences:

```javascript
switch(metadata.getType()) {
  case 'buckets':
    return handlebuckets(metadata);
  case 'categories':
    return handleCategories(metadata):
...
}
```

src/api/v4/layer/metadata/buckets.js
```diff
@@ -0,0 +1,79 @@
+var Base = require('./base');
+
+/**
+ * Metadata type gradient
```
IagoLast: TODO: add more doc

src/api/v4/layer/metadata/categories.js
```diff
@@ -0,0 +1,56 @@
+var Base = require('./base');
+
+/**
+ * Metadata type categories
+ *
```
IagoLast: TODO: add more doc

src/api/v4/layer/metadata/parser.js
```diff
@@ -0,0 +1,35 @@
+var BucketsMetadata = require('./buckets');
+var CategoriesMetadata = require('./categories');
+var Rule = require('../../../../windshaft-integration/legends/rule.js');
+
+module.exports = {
```
IagoLast: minor issue but in order to be consistent:

- private methods are prefixed with underscore
- exported functions are defined in the primary scope (instead inside a module.exports object)
-  module.exports goes at the end of the file and only contains references (not definitions)

PR #1950, opened 2017-12-07
Fixes FRA, issue 29

csobier:
@inigomedina 

Fixes https://github.com/CartoDB/full-reference-api/issues/29


Jesus89:
@csobier we don't want to merge `v4` in `develop` yet.

Also, the issue https://github.com/CartoDB/full-reference-api/issues/29 is still not fixed in `v4`. Do we want to add a `setZIndex` function for the carto layers?

csobier:
@Jesus89, OK, good to know. Thanks.

Yes, if there is time to add new code before the beta, please add. If not, we can create a new issue for the public release.

Jesus89:
We can add little changes for the beta on Monday, but we need to have clear the API for this.

Now the Leaflet layers has its own setZIndex function so we need to distinguish it in the docs from the Carto layer `setZIndex(<number>)`. May be `bringToFront()` and `bringToBack()`functions are useful too.

Jesus89:
From the team, we have decided to freeze the development for the beta release and we will think about Carto layers ordering in next versions.

csobier:
I 100% agree to code freeze, so we need to include how to reorg the layers with what we have. Is the text description from this PR okay?

Jesus89:
-> https://github.com/CartoDB/full-reference-api/issues/29

PR #1955, opened 2017-12-12
Wax webpack

IagoLast:
🎉


Jesus89:
This will be done in the integration ticket

Issue #1954, opened 2017-12-12
Integration: Builder & DI


Jesus89:
Builder PR: https://github.com/CartoDB/cartodb/pull/13211
Deep-insights.js PR: https://github.com/CartoDB/deep-insights.js/pull/635
Carto.js PR: https://github.com/CartoDB/cartodb.js/pull/1956

* Using zera instead of wax
* Using webpack in Carto.js

Jesus89:
:rocket:

PR #1956, opened 2017-12-12
Wax webpack

IagoLast:
Related to: https://github.com/CartoDB/cartodb.js/issues/1954

🎉


webpack.config.js
```diff
@@ -1,6 +1,12 @@
 const path = require('path');
 const webpack = require('webpack');
 
+var version = require('./package.json').version;
```
IagoLast: minor issue:
- can we use string literals for this?
- can we extract this to a `/scripts/banner.js` and use it from both webpack configs?

alonsogarciapablo:
👏 👏 👏 👏 👏

elenatorro:
🎉🎉🎉🎉🎉

Issue #1736, opened 2017-09-05
Replace browserify by webpack and fix bundling issues


alonsogarciapablo:
- [x] Migrate from browserify to webpack to be consistent with other projects (eg: http://github.com/cartodb/cartodb)
- [x] Fix bundling issues (eg: bundle depends on window variables like wax). Make sure other projects can require this library easily.
- [x] Add Promise polyfill
- [x] Enable npm publish
- [x] Update the deploy system
- [x] Publish to our cdn

Jesus89:
- [ ] Enable npm publish
- [ ] Update the deploy system

PR #1957, opened 2017-12-12
Update deploy system

Jesus89:
https://github.com/CartoDB/cartodb.js/issues/1736
---


scripts/generate-package-json.js
```diff
@@ -0,0 +1,8 @@
+var fs = require('fs');
```
IagoLast: Since is a little weird IMO we should add a comment explaining what this file does and why

src/geo/gmaps/gmaps-cartodb-layer-group-view.js
```diff
@@ -1,7 +1,7 @@
 /* global Image, google */
 var _ = require('underscore');
 var GMapsLayerView = require('./gmaps-layer-view');
-var zera = require('../../../vendor/zera');
+var zera = require('iagolast-zera');
```
IagoLast: have you tried linking builder, di.js and carto ? is builder compilation working?

Jesus89: The idea is to merge this working PR and then update `wax-webpack` branch in the integration with Builder & DI.

webpack.config.js
```diff
@@ -1,14 +1,19 @@
...
     path: path.resolve(__dirname, 'dist/public'),
-    filename: 'carto.bundle.js',
+    filename: 'carto.js',
     library: 'carto',
     libraryTarget: 'umd'
   },
-  // Tell webpack not to include Leaflet in the bundle. Leaflet should will be included manually
+  plugins: [
+    // Include only the lastest camshaft-reference
+    new webpack.IgnorePlugin(/^\.\/((?!0\.59\.4).)*\/reference\.json$/)
```
IagoLast: how this works?

webpack.min.config.js
```diff
@@ -0,0 +1,22 @@
...
+module.exports = {
+  entry: './src/api/v4/index.js',
+  output: {
+    path: path.resolve(__dirname, 'dist/public'),
+    filename: 'carto.min.js',
+    library: 'carto',
+    libraryTarget: 'umd'
+  },
+  plugins: [
+    new UglifyJsPlugin(),
```
IagoLast: I think the plugins are executed in the specified order so first we need to ignore the camshaft references and then uglify the whole bundle

webpack.config.js
Jesus89: This regex filter the `./0.59.3/reference.json` and previous versions, leaving only the latest version, that is the 0.59.4

webpack.min.config.js
Jesus89: I think it doesn't matter, because the resulting carto.min.js has the same size. I'm going to change the order anyway.

webpack.config.js
IagoLast: so if someone updates the required camshaft reference should update this file too right?

Jesus89: Yes, and also the browserify file. This will be nices when we get rid of camshaft-reference.

PR #1964, opened 2017-12-13
Improve docs

IagoLast:


src/api/v4/layer/layer.js
```diff
@@ -25,14 +36,18 @@ var metadataParser = require('./metadata/parser');
...
  * @example
- * // with options
+ * // Create a layer indicating which columns in the data will throw events on mouse over
  * new carto.layer.Layer(citiesSource, citiesStyle, {
- *   featureClickColumns: [ 'name', 'population' ],
  *   featureOverColumns: [ 'name' ]
  * });
+ * @example
+ * // Listen to the event thrown when the mouse is over a feature
+ * layer.on(carto.layer.events.FEATURE_OVER, featureEvent => {
```
alonsogarciapablo: Please use 'featureOver' here.

IagoLast: you mean raw string? if so we got to change the examples too

src/api/v4/client.js
```diff
@@ -15,8 +15,12 @@ function getValidationError (code) {
 /**
  * This is the entry point for a CARTO.js application.
  *
- * A CARTO client allows managing layers and dataviews. It also takes care
- * of the communication between a CARTO.js application and the services in CARTO.
+ * A CARTO client allows managing layers and dataviews. Once you add a layer or a dataview, the client 
+ * handles the communication against the server in order to process those data.
```
alonsogarciapablo: "...the client handles the communication against the server in order to process those data"... 🤔

src/api/v4/client.js
```diff
@@ -15,8 +15,12 @@ function getValidationError (code) {
...
+ * 
+ * This means the client is an asynchronous object: When any change occurs, it will send a request to the server
+ * and act accordingly by updating the views or throwing an error when appropriate. This whole process is known as the `reload cycle`.
```
alonsogarciapablo: I would try to avoid talking about the "reload cycle". It's a bit too internal for me. We can mention that some methods are async because the client needs to sync / communicate with the server.

src/api/v4/layer/layer.js
alonsogarciapablo: Yes, we're moving towards that in https://github.com/CartoDB/cartodb.js/pull/1963.

src/api/v4/layer/layer.js
```diff
@@ -9,11 +9,22 @@ var metadataParser = require('./metadata/parser');
 
 /**
  * Represent a layer Object.
+ * 
+ * A layer is the primary way to visualize geospatial data. 
+ * 
+ * To create a layer a {@link carto.source.Base|source} and {@link carto.style.Base|styles}
+ * are required:
+ * 
+ * - The {@link carto.source.Base|source} is used to know which **what** data will be displayed in the Layer.
```
alonsogarciapablo: There is an extra `which` here I think.

src/api/v4/layer/layer.js
```diff
@@ -25,14 +36,18 @@ var metadataParser = require('./metadata/parser');
  * @fires carto.layer.styleChanged
  * @fires carto.layer.MetadataEvent
  * @example
- * // no options
+ * // Create a layer with no options
  * new carto.layer.Layer(citiesSource, citiesStyle);
  * @example
- * // with options
+ * // Create a layer indicating which columns in the data will throw events on mouse over
```
alonsogarciapablo: Create a layer indicating what columns will be included in the `featureOver` event.

Issue #1974, opened 2017-12-15
Use leaflet from window.L instead commonjs


IagoLast:
Needed to support CJS apps ⚠️ 
--- 

This reduces the bundle size allowing users to load Leaflet externally.

- When using umd just include leaflet as a script 
- When using common.js: Do `require('leaflet')` inside your bundle

Jesus89:
Closed by #1978

PR #1978, opened 2017-12-18
Use window leaflet

Jesus89:
Related to: https://github.com/CartoDB/cartodb.js/issues/1974

This PR fixes:
- Leaflet was required before Carto.js in the browser
- Leaflet was required using Carto.js in node

### DIFF
--- a/src/cartodb.js
+++ b/src/cartodb.js
@@ -1,6 +1,5 @@
-var isLeafletAlreadyLoaded = !!window.L;
+window.L = require('leaflet');
 
-var L = require('leaflet');
 require('mousewheel'); // registers itself to $.event; TODO what's this required for? still relevant for supported browsers?
 require('mwheelIntent'); // registers itself to $.event; TODO what's this required for? still relevant for supported browsers?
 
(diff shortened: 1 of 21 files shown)


Jesus89:
I think we can assume that it comes from window, as the `google` lib.

With the `/* global X */` tags in the beginning of the file we can see that it comes from the global scope.

IagoLast:
Lets vote!

@alonsogarciapablo @ivanmalagon

src/cartodb.js
```diff
@@ -1,6 +1,5 @@
-var isLeafletAlreadyLoaded = !!window.L;
+window.L = require('leaflet');
```
alonsogarciapablo: Shouldn't we remove this too?

alonsogarciapablo:
I'm fine with using `L` directly. We all know where it comes from.

src/cartodb.js
Jesus89: This is the only require used. It is used by the internal bundle if we remove this line the Builder do not include Leaflet in the window context.

Issue #1976, opened 2017-12-18
Prevent fetch grid.json files non-interactive layers


IagoLast:
Currently the `grid.json` tiles containing the interactivity information for a layer are being requested whether the layer is interactive or not.

For some reason (bug?) get a different behaviour that form embeds and builder.

- Builder:
https://cartocdn-ashbu_a.global.ssl.fastly.net/iago-carto/api/v1/map/0acefb8ae9657da337e1b8e37b79128b:1510838928849/1/7/60/48.grid.json?api_key=552dc6b016acb007acf0b9091719be8d0a37847b

- Embed:
https://cartocdn-ashbu_a.global.ssl.fastly.net/iago-carto/api/v1/map/iago-carto@f2153e54@c70f5b8a89a9d7f657e9f29bd8724126:1510838928849/1/7/62/48.grid.json

Jesus89:
:rocket:

ivanmalagon:
:clap: :clap: :clap: :clap: :clap: 

Now TrackJS in back on track (pun intended)

PR #1979, opened 2017-12-18
Prevent fetching grid.json tiles in embeds with no interactive layers.

IagoLast:
#1976
--- 

- Check if a layer is interactive or the map has the interactivity forced
- Update cartodb-layer-group tests
  - Add leaflet tests
  - Add google tests


src/geo/cartodb-layer-group-view-base.js
```diff
@@ -20,12 +21,20 @@ CartoDBLayerGroupViewBase.prototype = {
...
       }
     }, this);
   },
 
+  _isIteractionEnabledForLayer: function (layerModel) {
+    if (!layerModel.isVisible()) {
+      return false;
+    }
+    // Builder needs interactivity when the layer has no popup/infowindows to allow feature edition.
+    return layerModel.isInteractive() || this._mapModel.isFeatureInteractivityEnabled();
```
alonsogarciapablo: The second condition has little to do with the layer as the name of this method suggests so I would extract it from this method. It might also make sense to include the `isVisible` part inside of `isInteractive`? Perhaps not...

IagoLast: yep, the name is not the best... maybe `_shouldEnableInteraction(layer)`  ?

src/geo/cartodb-layer-group-view-base.js
```diff
@@ -57,6 +66,7 @@ CartoDBLayerGroupViewBase.prototype = {
         .on('off', function (o) {
           if (self._interactionDisabled) return;
           o = o || {};
+          // TODO: zera has an .on('error', () => { }) callback that should be used here
```
alonsogarciapablo: `... that should be used here` -> Why don't we use it then?

test/spec/geo/gmaps/gmaps-cartodb-layer-group.spec.js
```diff
@@ -1,40 +1,297 @@
...
+      layerGroupView = new LeafletCartoDBLayerGroupView(layerGroupModelMock, {
+        nativeMap: googleMap,
+        mapModel: mapModelMock
+      });
+
+      expect(layerGroupView.options.tiles[0]).toEqual('http://0.ashbu.cartocdn.com/documentation/api/v1/map/9876543210/1,2/{z}/{x}/{y}.png');
+    });
+  });
+
+  xdescribe("when there aren't tile URLS", function () {
```
alonsogarciapablo: xdescribe

test/spec/geo/gmaps/gmaps-cartodb-layer-group.spec.js
```diff
@@ -1,40 +1,297 @@
...
+        // expect(event.position).toEqual({ x: 264, y: 309 });
+        // expect(event.latlng).toEqual([42.293564192170095, -8.173828125000002]);
+        done();
+      });
+
+      layerGroupView.interaction[0]._eventEmitter.dispatchEvent('featureout', eventData);
+    });
+
+    it('should trigger a "featureError" event when interactivity triggers a featureOut event', function (done) {
+      pending('Not implemented yet');
```
alonsogarciapablo: pending

test/spec/geo/gmaps/gmaps-cartodb-layer-group.spec.js
```diff
@@ -1,40 +1,297 @@
...
+        mapModel: mapModelMock
+      });
+
+      expect(layerGroupView.options.tiles[0]).toEqual('http://0.ashbu.cartocdn.com/documentation/api/v1/map/9876543210/1,2/{z}/{x}/{y}.png');
+    });
+  });
+
+  xdescribe("when there aren't tile URLS", function () {
+    it('should fetch empty tiles', function () {
+      pending('how to compare the empty gif against the tile url?');
```
alonsogarciapablo: pending

test/spec/geo/gmaps/gmaps-cartodb-layer-group.spec.js
```diff
@@ -1,40 +1,297 @@
...
+        nativeMap: googleMap,
+        mapModel: mapModelMock
+      });
+
+      expect(layerGroupView.interaction.length).toEqual(2);
+    });
+
+    it('should enable interaction when there are interactive layers ', function () {
+      mapModelMock.isFeatureInteractivityEnabled.and.returnValue(false);
+      spyOn(cartoDbLayer0, 'isInteractive').and.returnValue(true);
```
alonsogarciapablo: Perhaps we can move this spy to the `beforeEach` so that we don't have to repeat it in other `it` blocks.

test/spec/geo/shared-tests-for-cartodb-layer-group-views.js
```diff
@@ -1,389 +0,0 @@
-var _ = require('underscore');
```
alonsogarciapablo: Why are we removing this and duplicating the tests instead?

alonsogarciapablo: This was/is kind of convenient IMO.

src/geo/cartodb-layer-group-view-base.js
IagoLast: Since the current issue is generating multiple errors in the track.js and we want to fix it asap IMO we should address the zera thing in a different issue.

test/spec/geo/shared-tests-for-cartodb-layer-group-views.js
IagoLast: Jasmine logs/workflow are easier to read this way (imo)

src/geo/cartodb-layer-group-view-base.js
IagoLast: https://github.com/CartoDB/zera/issues/5

Issue #1970, opened 2017-12-15
Add fetch & Promise Polyfill to V3


IagoLast:
`zera` uses fetch to get the grid.json files from the server. Since fetch wont work in old browsers we need to add a polyfill.

We [added a polyfill for the v4 version](https://github.com/CartoDB/cartodb.js/blob/v4/src/api/v4/index.js#L23) but we forgot to add the polyfill in the v3 (used by builder).

At the moment @ivanmalagon fixed this situation adding the polyfill in builder but the bug is still here.

Jesus89:
I think it would be nice to add the polyfill in `carto-zera`. Then this package will be fully-compatible with old browsers.

IagoLast:
@ivan should we remove the polyfill from builder?

ivan:
Wrong @ivan

PR #1980, opened 2017-12-18
Add polyfills

IagoLast:
#1970
---

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,3 +1,8 @@
+// Add polyfill for `fetch`
+require('whatwg-fetch');
+// Add polyfill for `Promise`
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -1,3 +1,8 @@
+// Add polyfill for `fetch`
+require('whatwg-fetch');
```
alonsogarciapablo: Shouldn't this be also added to `package.json`?

Jesus89: This is already in the package.json because we use it for the public bundle.

alonsogarciapablo: :+1:

Issue #1908, opened 2017-11-20
Get rid of moment-timezone


ivanmalagon:
As of now, we're requiring `moment-timezone` only to get the user timezone offset. It brings 418KB of weight:

![image](https://user-images.githubusercontent.com/1078228/33018877-6be70ed0-cdf8-11e7-89c6-6675a51a66cb.png)

We can use the native way of doing it https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTimezoneOffset

Take into account that Windshaft doesn't support offset that are not complete hours, so we should add some logic to give a right offset.

Jesus89:
Removed also `moment` dependency.

Jesus89:
Closed by https://github.com/CartoDB/cartodb.js/pull/1981

PR #1981, opened 2017-12-19
Replace moment

Jesus89:
Related to: https://github.com/CartoDB/cartodb.js/issues/1908

- Replace moment-timezone calculations
- Replace moment calculations
- Remove moment and moment-timezone deps

The new calculations return exactly the same values that the ones from moment.js

Final bundle size:
- **carto.js**: 1.4 MB to **913 kB**
- **carto.min.js**: 758 kB to **315 kB** :tada:


src/dataviews/helpers/histogram-helper.js
```diff
@@ -35,27 +29,22 @@ function trimBuckets (buckets, filledBuckets, totalBuckets) {
...
 helper.fillTimestampBuckets = function (buckets, start, aggregation, numberOfBins, from, totalBuckets) {
-  var startDate = moment.unix(start).utc();
   var filledBuckets = []; // To catch empty buckets
   var definedBucket = false;
 
   for (var i = 0; i < numberOfBins; i++) {
     definedBucket = buckets[i] !== undefined;
     filledBuckets.push(definedBucket);
 
+    var bucketStart = add(start, i, AGGREGATION_DATA[aggregation]);
```
Algunenano: Would it be possible to use `buckets[i+1].start -1` for the end value instead (handling the last item)? PostgreSQL already handles a lot of stuff like changing timezones (winter/summer or [political reasons](https://www.timeanddate.com/news/time/venezuela-change-timezone.html)) so for example a day bucket could have 23.5 hours.

Algunenano: Nevermind, since not all buckets are returned this is not possible. We would need to start returning start and end from windshaft first.

src/dataviews/helpers/histogram-helper.js
```diff
@@ -83,47 +72,46 @@ helper.hasChangedSomeOf = function (list, changed) {
...
+/**
+ * Add a `number` of aggregations to the provided timestamp
+ *
+ * @param {number} timestamp - Starting timestamp
+ * @param {number} number - Number of aggregations to add
+ * @param {object} aggregation
+ * @param {string} aggregation.unit - unit of the aggregation
+ * @param {number} aggregation.factor - number of aggretagion units
+ */
+function add (timestamp, number, aggregation) {
```
IagoLast: do we have tests for this funcion?

Jesus89: No unit tets, but I have tested it, comparing the output results with the moment.js output for different timestamps and different aggregations.

``` js
function test () {
  var timestamp = 12341324;
  var startDate = moment.unix(timestamp).utc();

  console.log('Start test', new Date(timestamp * 1000));

  var agg = 'quarter';

  for (var i = 0; i < 100000; i++) {
    var momentResult = startDate.clone().add(i, MOMENT_AGGREGATIONS[agg]).unix();
    var localResult = addTimestamp(timestamp, i, MOMENT_AGGREGATIONS_DATA[agg]);

    if (localResult !== momentResult) {
      console.log(new Date(localResult * 1000), new Date(momentResult * 1000));
    }
  }

  console.log('End test');
}

IagoLast: I think having unit tests for "complex" functions like this is always a good idea

alonsogarciapablo: +1 for adding some unit tests

Jesus89: There you go: https://github.com/CartoDB/cartodb.js/commit/1d2879d13f7353ab85f7abec9420fb13b3a5e3e8

PR #1983, opened 2017-12-19
JSDoc fixes

alonsogarciapablo:


src/api/v4/client.js
```diff
@@ -81,11 +81,9 @@ Client.prototype.addLayer = function (layer, opts) {
 };
 
 /**
- * Add multiple layers to the client. Note the hierarchal order of layers.
- * Use this method instead multiple calls to **addLayer** to improve performance.
+ * Add multiple layers to the client at once.
  *
- * @param {carto.layer.Base[]} - An array with the layers to be added. Note that ([A, B]) displays B as the first layer. Alternatively, client.addLayer(A); client.addLayer(B);
- * @param {object} opts
+ * @param {carto.layer.Base[]} - An array with the layers to be added. Note that ([A, B]) displays B as the first layer.
```
Jesus89: `Note that ([A, B]) displays B as the **top** layer.`

alonsogarciapablo: Good catch! 🔝

Issue #1942, opened 2017-12-01
Promisify layer interactivity


ivanmalagon:
The methods `.setFeatureClickColumns` and `.setFeatureOverColumns` should be returning and handling engine promises, just like their sibling functions `.setStyle` and `.setSource`.

If not, an error in those functions runs another logic, exposing a Windshaft error and returning an uncaught promise.

PR #1986, opened 2017-12-19
Return Promises in setFeatureColumns functions

Jesus89:
Related to: https://github.com/CartoDB/cartodb.js/issues/1942
---


src/api/v4/layer/layer.js
```diff
@@ -217,12 +233,27 @@ Layer.prototype.getFeatureClickColumns = function (columns) {
...
+    this._featureOverColumns = columns;
+    return Promise.resolve();
+  }
+  // Update the internalModel and return a promise
+  this._internalModel.tooltip.fields.set(_getInteractivityFields(columns).fields, { silent: true });
+  return this._engine.reload()
+    .then(function () {
+      this._featureOverColumns = columns;
+    }.bind(this))
+    .catch(function (err) {
```
IagoLast: Boy scout rule here:  I think this 3 lines could be extracted into a function to reduce LOC

```javascript
function _rejectAndTriggerError(err){
  var error = new CartoError(err);
  this.trigger(EVENTS.ERROR, error);
  return Promise.reject(error);
}
```

And use:

```javascript
.catch(_rejectAndTriggerError.bind(this));
```

ivanmalagon: :+1: That pattern is used in every promise rejection in the file so we can refactor the four usages.

PR #1996, opened 2018-01-02
[WIP] Tilejson support

rochoa:
This is my proposal about how to start exposing access to vector tiles through the new CARTO.js library.

I'm sure about how unsure I'm about these changes, so all the feedback is welcome.

## Update
This is not only about vector tiles, this is also about how we enable libraries not supported *natively* within CARTO.js—i.e., how do I use CARTO.js in combination with OpenLayers?

### DIFF
--- a/src/api/v4/client.js
+++ b/src/api/v4/client.js
@@ -343,7 +343,7 @@ Client.prototype.getTilejson = function (format, layerFilters) {
   return (format === 'vector') ? vectorTilejson(layergroup, layerFilters) : rasterTilejson(layergroup, layerFilters);
 };
 
-function tileUrls(layergroup, format, layerFilters) {
+function tileUrls (layergroup, format, layerFilters) {
   layerFilters = layerFilters || 'mapnik';
   layerFilters = Array.isArray(layerFilters) ? layerFilters : [layerFilters];
   var rasterTilesUrl = layergroup.tiles
@@ -352,23 +352,23 @@ function tileUrls(layergroup, format, layerFilters) {
   if (layergroup.subdomains.length === 0) {
     return [rasterTilesUrl];
   }
-  return layergroup.subdomains.map(function(subdomain) {
+  return layergroup.subdomains.map(function (subdomain) {
     return rasterTilesUrl.replace('{s}', subdomain);
   });
 }
 
-function rasterTilejson(layergroup, layerFilters) {
+function rasterTilejson (layergroup, layerFilters) {
   return {
-    tilejson: "2.2.0",
+    tilejson: '2.2.0',
     tiles: tileUrls(layergroup, 'png', layerFilters),
-    grids: layergroup.grids,
+    grids: layergroup.grids
   };
 }
 
-function vectorTilejson(layergroup, layerFilters) {
+function vectorTilejson (layergroup, layerFilters) {
   return {
-    tilejson: "2.2.0",
-    tiles: tileUrls(layergroup, 'mvt', layerFilters),
+    tilejson: '2.2.0',
+    tiles: tileUrls(layergroup, 'mvt', layerFilters)
   };
 }
 
--- a/src/api/v4/client.js
+++ b/src/api/v4/client.js
@@ -326,6 +326,53 @@ Client.prototype.getGoogleMapsMapType = function (map) {
 };
 
 /**
+ * Return a {@link https://github.com/mapbox/tilejson-spec|TileJSON} for the map tileset.
+ *
+ * @example
+ * // Get the tilejson from the client
+ * const tilejson = client.getTilejson();
+ *
+ * @param {string} - The tileset format: vector or raster.
+ *
+ * @returns A {@link https://github.com/mapbox/tilejson-spec|L.TileJSON} for the raster or vector tileset.
+ *
+ * @api
+ */
+Client.prototype.getTilejson = function (format, layerFilters) {
+  var layergroup = this._engine._cartoLayerGroup.get('urls');
+  return (format === 'vector') ? vectorTilejson(layergroup, layerFilters) : rasterTilejson(layergroup, layerFilters);
+};
+
+function tileUrls(layergroup, format, layerFilters) {
+  layerFilters = layerFilters || 'mapnik';
+  layerFilters = Array.isArray(layerFilters) ? layerFilters : [layerFilters];
+  var rasterTilesUrl = layergroup.tiles
+    .replace('{format}', format)
+    .replace('{layerIndexes}', layerFilters.join(','));
+  if (layergroup.subdomains.length === 0) {
+    return [rasterTilesUrl];
+  }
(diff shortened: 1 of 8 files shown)


IagoLast:
Im gonna start with the style issues and once fixed we can start the real review 🕶 

First of all run `yarn lint:fix`/ `npm run lint:fix`  and update the PR  with the auto-solved issues

src/api/v4/client.js
```diff
@@ -325,6 +325,53 @@ Client.prototype.getGoogleMapsMapType = function (map) {
...
+ * // Get the tilejson from the client
+ * const tilejson = client.getTilejson();
+ *
+ * @param {string} - The tileset format: vector or raster.
+ *
+ * @returns A {@link https://github.com/mapbox/tilejson-spec|L.TileJSON} for the raster or vector tileset.
+ *
+ * @api
+ */
+Client.prototype.getTilejson = function (format, layerFilters) {
```
rochoa: @IagoLast, I think your comment is gone after I fixed the linting errors.

> Do we really want to expose this function? I think is not consistent with our current raster api where we directly expose a raster layer.
> `client.getLeafletLayer().addTo(map) // raster`

I'm willing to consider any alternative that enables using any mapping library that we don't support *natively* in CARTO.js.

IagoLast: Are we going to allow "clustering options" in raster layers?

IagoLast:
are we going to allow "clustering options" in raster layers?

IagoLast:
I see a design issue here, we are degenerating `carto.layer.Layer` and doing some weird stuff (empty styles) to create something that in my opinion is different enough to be a different kind of object (a vector layer).

So in terms of public api, I would add a  new `carto.layer.Vector` this kind of layer doesn't require a style so we could [get rid of this weird checks](https://github.com/CartoDB/cartodb.js/pull/1996/files#diff-8e22edeb2556633c9879fc8ae88dd8edR59) 

instead we could have

```javascript
function VectorLayer(id, source, options) {
  // ...
  this._id = id;
  this.aggregation = options.aggregation;
  // ... 
}


VectorLayer.prototype._createInternalModel = function (engine) {
  var internalModel = new CartoDBLayer({
    id: this._id, 
    source: this._source.$getInternalModel(),
    cartocss: undefined, // we dont need Empty style
    cartocss_version: CartoCSS.DEFAULT_CARTOCSS_VERSION, // not sure if we use this for anything
    aggregation: this._aggregation,
    visible: this._visible,
    infowindow: _getInteractivityFields(this._featureClickColumns),
    tooltip: _getInteractivityFields(this._featureOverColumns)
  }, { engine: engine });
```

This way we:

- Remove the [empty style](https://github.com/CartoDB/cartodb.js/pull/1996/files#diff-49a49fe9fa9308f78b83386b786aebe4)
- Have two well defined kinds of layer, each one with its custom parameters and behaviours



Maybe a different approach could be applied for what we call `internal models` (objects that are not exposed to the final users) in this case `mapnik-layer` and `torque-layer`.

should we have a internal `vector-layer` ?

src/windshaft/map-serializer/anonymous-map-serializer/layers-serializer.js
```diff
@@ -77,6 +77,15 @@ function optionsForMapnikLayer (layerModel) {
     options.aggregation = aggregation;
   }
 
+  var minzoom = layerModel.get('minzoom');
+  if (Number.isFinite(minzoom)) {
+    options.minzoom = minzoom;
+  }
+  var maxzoom = layerModel.get('maxzoom');
+  if (Number.isFinite(maxzoom)) {
```
IagoLast: this wont work in IE :trollface:

rochoa: I know, but you are here to help me with the right approach. 😄

rochoa:
I agree this design is not perfect. What if I want to have my layer to support a fallback mechanism to render in raster if the client doesn't support vector tiles/rendering?

After your comment, wouldn't be better to keep this Layer as simple as possible — not holding the style — and having a `carto.layer.Raster` layer that actually has the style?

BTW, the aggregation is a valid option for raster tiles.

IagoLast:
> What if I want to have my layer to support a fallback mechanism to render in raster if the client doesn't support vector tiles/rendering?

We can throw an error and let the user implement a fallback. IMO is not trivial to implement this fallback ourselves: (for example, what would be the default style?)

> After your comment, wouldn't be better to keep this Layer as simple as possible — not holding the style — and having a carto.layer.Raster layer that actually has the style?

`carto.layer.Layer` actually means `carto.layer.Raster` maybe we should rename it and create a `carto.layer.Base` holding the common behaviour to be consistent with the remaining objects.

rochoa:
> We can throw an error and let the user implement a fallback. IMO is not trivial to implement this fallback ourselves: (for example, what would be the default style?)

If I define the layer with CartoCSS I can use that layer as a raster layer but also as a vector layer, so I might want to do some kind of fallback mechanism, see https://github.com/CartoDB/cartodb.js/commit/c40dd96fe381d48a018a14ee0243aa4da7fe4119.

IagoLast:
> If I define the layer with CartoCSS I can use that layer as a raster layer but also as a vector layer, so I might want to do some kind of fallback mechanism, see c40dd96

yep, but if rasterLayer and vectorLayer both have cartocss, and options as parameters in the constructor I dont see the need for creating a new "class" since they will work pretty much the same right?

alonsogarciapablo:
cc: @inigomedina

ivanmalagon:
Great work, @rochoa. Unfortunately, we can't accept this PR as it is but it'll serve as a guide. The main issue is that it doesn't reflect the approach that we follow adding features to the API.

Out of your PR, I identified all the changes that it includes and created several issues to tackle them. As I see it, we have two groups of tasks:
- Aggregation ones: this are used by raster as well and not related per se to vector tiles
    - Allow setting ID in layers #2000 : helpful also for a vizjson@next
    - Make layer style optional #2001 
    - Add minzoom and maxzoom to layer options #2002 
    - Add aggregation options to layers #2005 
- Enable libraries not supported natively #2006 

Questions will rise along the way.

I tagged them with `v4.1` since these improvements are out of 4.0 release (unless @inigomedina says the contrary).

ivanmalagon:
Just for the record, the PR works like a charm. `carto.js` rendering vector tiles and windshaft aggregating data.

vt



image

alonsogarciapablo:
Wow. It looks really nice @rochoa!

inigomedina:
> I tagged them with v4.1 since these improvements are out of 4.0 release (unless @inigomedina says the contrary).

Agree. @rochoa did you expect to add it to v4.0?

rochoa:
Why? 😢 @rubenmoya @rjimenezda

rubenmoya:
We were doing some cleanup of old PRs/branches, sorry!

Issue #2000, opened 2018-01-10
Allow setting ID in layers


ivanmalagon:
We decided not to allow users to set their own IDs for layers in 4.0. The reason was that now the API is programatic, so the user already have objects with the layers so she could refer to them using those objects directly.

Two issues now forces us to allow custom IDs:
- Vector rendering: as seen in this [PR #1996](https://github.com/CartoDB/cartodb.js/pull/1996/files) the rendering library needs the ID of the layers. The developer should know in advance those IDs.
- vizjson@next: there's a growing expectation on providing something à la vizjson. That usage need to resemble the IDs set by Builder to allow developers to modify data in Builder and expect that the IDs will remain the same.

Tasks:
- Enable setting custom IDs to `carto.layer.Layer` via constructor.
- Add getter to retrieve the ID.
- Perform validations. We can't allow multiple layers with the same ID.
- Modify documentation.
- TBD: retrieve layers by ID.

PR #2009, opened 2018-01-15
2000 allow custom IDs in layers

IagoLast:
https://github.com/CartoDB/cartodb.js/issues/2000
--- 

Pending documentation update.


src/api/v4/layer/layer.js
```diff
@@ -75,8 +77,9 @@ function Layer (source, style, options) {
   this._visible = true;
   this._featureClickColumns = options.featureClickColumns || [];
   this._featureOverColumns = options.featureOverColumns || [];
-
-  Base.apply(this, arguments);
+  if (options.id) {
+    this._id = options.id;
+  }
```
ivanmalagon: With this approach, we generate an `id` always and then, in case the user provides an ID, it gets overridden.

I think it'd better to pass the `id` to `Base` as an option and, if present, assign it and avoiding generating one. `Base` exists only to manage the `id` and the `internalModel` so it seems a good place to contain all logic related to `ids`.

IagoLast: good catch

Issue #2005, opened 2018-01-10
Add aggregation options to layers


ivanmalagon:
As part of the 5X project, Windshaft now supports aggregation, both for raster and vector layers.

The aggregation format is:
```javascript
aggregation: {
  threshold: 10K, // start aggregating data when the layer has above this amount of rows
  resolution: 1, // aggregation per pixel resolution. 1 is aggregate per 1x1 pixel, 2 aggregate per 2x2 pixels, 1/2 aggregate per subpixel... This number should be powers of 2
  placement: 'point-sample', // location of the resulting aggregated point. Values: 'centroid', 'point-grid', 'point-sample'
  columns: { // Columns to return as aggregated data
    'mag': { // name to return as the aggregated column
      aggregate_function: 'sum',
      aggregated_column: 'mag' // Column to aggregate by
    }
  }              
}

```

My suggestion is to create a class to hold this options, as we do with `CartoCSS`. This way we can make better validation of the aggregation options.

Tasks:
- Create an aggregation option to pass to layers.
- Create documentation

PR #2010, opened 2018-01-15
Add windshaft aggregation

ivanmalagon:
https://github.com/CartoDB/cartodb.js/issues/2005
----

### DIFF
--- a/src/api/v4/layer/aggregation.js
+++ b/src/api/v4/layer/aggregation.js
@@ -1,20 +1,40 @@
 var _ = require('underscore');
 var CartoValidationError = require('../error-handling/carto-validation-error');
 
-// Taken from https://carto.com/docs/carto-engine/maps-api/tile-aggregation/#columns
-var VALID_OPERATIONS = {
-  avg: true,
-  sum: true,
-  min: true,
-  max: true,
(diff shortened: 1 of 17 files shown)


src/api/v4/layer/aggregation.js
```diff
@@ -0,0 +1,150 @@
+var _ = require('underscore');
+var CartoValidationError = require('../error-handling/carto-validation-error');
+
+// Taken from https://carto.com/docs/carto-engine/maps-api/tile-aggregation/#columns
+var VALID_OPERATIONS = {
```
ivanmalagon: We must provide an enum for this, just like we do with `carto.dataview.status`, `carto.dataview.status`, ...

src/api/v4/layer/aggregation.js
```diff
@@ -0,0 +1,150 @@
...
+  avg: true,
+  sum: true,
+  min: true,
+  max: true,
+  mode: true
+};
+
+// Taken from https://carto.com/docs/carto-engine/maps-api/tile-aggregation/#placement
+var VALID_PLACEMENTS = {
```
ivanmalagon: Same as above.

Issue #2002, opened 2018-01-10
Add minzoom and maxzoom to layer options


ivanmalagon:
Windshaft now supports layers with `minzoom` and `maxzoom` options.

Their meaning is that tiles won't render that layer below `minzoom` level or above `maxzoom'.

These names are taken from [TileJSON specification](https://github.com/mapbox/tilejson-spec/tree/master/2.2.0) so my suggestion is to leave them as `minzoom` and `maxzoom` instead of aggregating them in a zoom object (that it's the approach we're taking in carto.js API).

Tasks:
- Implement these options and their validation in Layer and its serializer.
- Update documentation.

PR #2015, opened 2018-01-17
2002 add layer zoom options

IagoLast:
#2002 
---

### DIFF
--- a/src/windshaft/map-serializer/anonymous-map-serializer/layers-serializer.js
+++ b/src/windshaft/map-serializer/anonymous-map-serializer/layers-serializer.js
@@ -77,6 +77,14 @@ function optionsForMapnikLayer (layerModel) {
     };
   }
 
+  if (isFinite(layerModel.get('minzoom'))) {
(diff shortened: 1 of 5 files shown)


src/windshaft/map-serializer/anonymous-map-serializer/layers-serializer.js
```diff
@@ -77,6 +77,14 @@ function optionsForMapnikLayer (layerModel) {
     };
   }
 
+  if (isFinite(layerModel.get('minzoom'))) {
```
ivanmalagon: Since the default value is `0` we are always sending `minzoom`. I don't know if it has implications in server side but just in case, I'd do that same thing you've done with `maxzoom`, leaving it `undefined`.

IagoLast: According to [this](https://github.com/CartoDB/Windshaft/blob/master/doc/MapConfig-1.7.0.md#2-file-format) `undefined` means infinite and is the default value for maxZoom.

I just copied the default values from windshaft

PR #2017, opened 2018-01-17
11341 add dblclick event

elenatorro:
Related issue: https://github.com/CartoDB/cartodb/issues/11341

This PR  adds a double click listener that removes the anchor point that has been clicked by the user. I am not sure if this is the best way to solve this, opinions needed :)

polygon

### DIFF
--- a/src/api/v4/native/trigger-layer-feature-event.js
+++ b/src/api/v4/native/trigger-layer-feature-event.js
@@ -37,7 +37,6 @@ module.exports = function (eventName, internalEvent, layers) {
      * @type {carto.layer.FeatureEvent}
      * @api
      */
-    /**
 
     /**
      * Fired when user moves the mouse over a feature.
--- a/src/api/v4/native/trigger-layer-feature-event.js
+++ b/src/api/v4/native/trigger-layer-feature-event.js
@@ -37,6 +37,14 @@ module.exports = function (eventName, internalEvent, layers) {
      * @type {carto.layer.FeatureEvent}
      * @api
      */
+    /**
+
+     * Fired when user clicks twice on a feature.
+     *
+     * @event featureDblclicked
+     * @type {carto.layer.FeatureEvent}
+     * @api
+     */
 
     /**
      * Fired when user moves the mouse over a feature.
--- a/src/api/v4/native/trigger-layer-feature-event.js
+++ b/src/api/v4/native/trigger-layer-feature-event.js
@@ -39,13 +39,6 @@ module.exports = function (eventName, internalEvent, layers) {
      */
     /**
 
-     * Fired when user clicks twice on a feature.
-     *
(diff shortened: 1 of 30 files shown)


src/api/v4/native/trigger-layer-feature-event.js
```diff
@@ -37,6 +37,7 @@ module.exports = function (eventName, internalEvent, layers) {
      * @type {carto.layer.FeatureEvent}
      * @api
      */
+    /**
```
IagoLast: !

src/geo/cartodb-layer-group-view-base.js
```diff
@@ -51,20 +51,20 @@ CartoDBLayerGroupViewBase.prototype = {
       this.interaction[layerIndexInLayerGroup] = new this.interactionClass()
         .map(this.nativeMap)
         .tilejson(tilejson)
-        .on('on', function (o) {
+        .on('on', function (event) {
```
IagoLast: To be more consistent I think this also could be renamed to `zeraEvent`

alonsogarciapablo:
Please make sure this works with Google Maps too! Thanks!

IagoLast:
> Please make sure this works with Google Maps too! Thanks!

Just checked, works like a charm ;)

elenatorro:
I added a couple of tests with two main use cases. I have been talking with @IagoLast and @ivanmalagon about the examples we are providing.

This feature works in the examples when editing an existing polygon, but if you create a new one you would need to bind manually the `dblclick` event. I think that for solving this issue, these changes are enough. However, I would like to discuss with you if you agree or if we should make more changes :)

test/spec/geo/geometry-views/shared-tests-for-path-views.js
```diff
@@ -98,6 +98,31 @@ module.exports = function (Path, MapView, PathView) {
...
+            40
+          ]
+        });
+
+        this.geometry.setCoordinates([
+          [ -10, 10 ], [ 10, 20 ], [ 30, 40 ]
+        ]);
+
+        this.geometry.addPoint(pointToDelete);
+
```
IagoLast: Just a minor thing:

I would add a pre-spec to clarify the test

```javascript
expect(this.geometry.getCoordinates()).toEqual([[ -10, 10 ], [ 10, 20 ], [ 30, 40 ], [-40, 40]); // <---
pointToDelete.trigger('dblclick');
expect(this.geometry.getCoordinates()).toEqual([[ -10, 10 ], [ 10, 20 ], [ 30, 40 ]);
```

alonsogarciapablo: 🙇

elenatorro: 👍

PR #2024, opened 2018-01-30
Override scrollwheel if it's boolean

rubenmoya:
When migrating the dashboard to builder (which uses cartodb.js v4 now) we found a problem with the `user feed` view, where you could scroll the map with the trackpad/mouse wheel. I've found out it was because of this:

```
vizjson.options.scrollwheel = options.scrollwheel || vizjson.options.scrollwheel;
```

In the `vizjson` the option is true, and why are trying to override it with `false`, which is impossible using an `||`.


Jesus89:
:+1:

Issue #2026, opened 2018-01-31
Remove Tangram's vector rendering support from v4 private API


rochoa:
Remove all code and all dependencies regarding Tangram vector rendering in v4 branch. As this is not part of the public API, we don't need to bump the major version.

PR #2028, opened 2018-01-31
Remove tangram support

dv343:
https://github.com/CartoDB/cartodb.js/issues/2026


IagoLast:
@Jesus89 for some reason the bundle size is the same, any ideas?

Jesus89:
The bundle size must be the same because it does not contain the tangram lib

IagoLast:
but we included [tangram.cartodb](https://github.com/CartoDB/cartodb.js/pull/2028/files#diff-33d355f56e59638d50713e9262b8c82aL4) which was around 50K

Jesus89:
But this file is not included in the public bundle

dv343:
Tested on staging, seems good to me, can you CR it @Jesus89 @IagoLast ?

Jesus89:
It seems to work without issues :sparkles:

alonsogarciapablo:
:fire::fire::fire:

Issue #2003, opened 2018-01-10
Check parameters in carto.filter.boundingboxLeaflet


IagoLast:
Currently when calling 

```javascript
const bboxFilter = carto.filter.boundingboxLeaflet(undefined);
```

We get ' cannot read property `on` of `undefined`.

I think we should check the parameters and trigger a better error when the parameter is not a leafletMap.

PR #2030, opened 2018-01-31
2003 bounding box error

IagoLast:
https://github.com/CartoDB/cartodb.js/issues/2003
---


src/api/v4/filter/bounding-box-leaflet.js
```diff
@@ -53,4 +56,12 @@ BoundingBoxLeaflet.prototype.$getInternalModel = function () {
   return this._internalModel;
 };
 
+// Helper to check if an element is a leafletmap object
+function _isLeafletMap (element) {
```
Jesus89: I would use `instance of L.Map` to skip also Bool, Number, String, Array and wrong Objects.

test/spec/api/v4/filter/bounding-box-leaflet.spec.js
```diff
@@ -0,0 +1,11 @@
+var carto = require('../../../../../src/api/v4/index');
+
+describe('api/v4/filter/bounding-box-leaflet', function () {
+  describe('constructor', function () {
+    it('should throw a descriptive error when initialized with invalid parameters', function () {
+      expect(function () {
```
Jesus89: Add more tests including bool, number, string, etc as map

IagoLast: I dont think is a good idea, the essence of the test is "throw an error when parameter is not a map?"

All those cases are equivalent

src/api/v4/filter/bounding-box-leaflet.js
```diff
@@ -53,4 +57,9 @@ BoundingBoxLeaflet.prototype.$getInternalModel = function () {
...
+  return element instanceof L.map;
```
Jesus89: Be careful! You need to use `L.Map` instead of L.map (that is the function)

test/spec/api/v4/filter/bounding-box-leaflet.spec.js
Jesus89: That's true.

src/api/v4/filter/bounding-box-gmaps.js
```diff
@@ -53,4 +59,9 @@ BoundingBoxGoogleMaps.prototype.$getInternalModel = function () {
   return this._internalModel;
 };
 
+// Helper to check if an element is a leafletmap object
+function _isGoogleMap (element) {
+  return element instanceof google.maps.Map;
```
ivanmalagon: We're relying on `google` to be present but this check will throw an error if object `google` is not present.

In this case we should trigger an error asking the user to get google properly added to the document.

src/api/v4/filter/bounding-box-leaflet.js
ivanmalagon: Same as with `google` above.

src/api/v4/filter/bounding-box-gmaps.js
```diff
@@ -53,4 +58,12 @@ BoundingBoxGoogleMaps.prototype.$getInternalModel = function () {
   return this._internalModel;
 };
 
+// Helper to check if an element is a leafletmap object
+function _isGoogleMap (element) {
+  if (!google) {
+    throw Error('Google maps is not defined');
```
ivanmalagon: What about a CartoError?

src/api/v4/filter/bounding-box-leaflet.js
```diff
@@ -53,4 +57,9 @@ BoundingBoxLeaflet.prototype.$getInternalModel = function () {
   return this._internalModel;
 };
 
+// Helper to check if an element is a leafletmap object
+function _isLeafletMap (element) {
+  return element instanceof L.Map;
```
ivanmalagon: Same check as with google here. We're relying in a global object that can be not present.

src/api/v4/filter/bounding-box-gmaps.js
IagoLast: IMO this is a "runtime exception" not a cartoError

Jesus89: I would use here the `_isGoogleMapsLoaded` function from the client.js file (may be moving it to a utils.js file)

src/api/v4/filter/bounding-box-leaflet.js
```diff
@@ -53,4 +57,12 @@ BoundingBoxLeaflet.prototype.$getInternalModel = function () {
   return this._internalModel;
 };
 
+// Helper to check if an element is a leafletmap object
+function _isLeafletMap (element) {
+  if (!L) {
```
Jesus89: Same here with the `_isLeafletLoaded` function

src/api/v4/filter/bounding-box-leaflet.js
```diff
@@ -24,6 +25,9 @@ var BoundingBoxFilterModel = require('../../../windshaft/filters/bounding-box');
  * dataview.addFilter(bboxFilter);
  */
 function BoundingBoxLeaflet (map) {
+  if (!_isLeafletMap(map)) {
+    throw new Error('Bounding box needs a leaflet map but got: ' + map);
```
Jesus89: ~leaflet~ Leaflet

src/api/v4/filter/bounding-box-gmaps.js
```diff
@@ -24,6 +26,9 @@ var BoundingBoxFilterModel = require('../../../windshaft/filters/bounding-box');
  * dataview.addFilter(bboxFilter);
  */
 function BoundingBoxGoogleMaps (map) {
+  if (!_isGoogleMap(map)) {
+    throw new Error('Bounding box needs a google map but got: ' + map);
```
Jesus89: ~google map~ Google Maps map

src/api/v4/native/utils.js
```diff
@@ -0,0 +1,26 @@
+
+function isLeafletLoaded () {
```
IagoLast: could move this to `core/util` ?

Jesus89: It could be, but it is in the API scope so I think a good place is in the `api` directory

ivanmalagon: I'm for that too. We are not exposing these functions in the public API so moving it to the internals so we can use it throughout the whole codebase.

Jesus89: Not a big deal so go for it :)

examples/public/filters/bounding-box-leaflet.html
```diff
@@ -128,6 +128,7 @@ <h2>Population</h2>
 
       client.addDataview(formulaDataview);
 
+      L = undefined;
```
Jesus89: ?

Issue #1982, opened 2017-12-19
Define API for Carto layers order


Jesus89:
``` js
//         BOTTOM v     v TOP
client.addLayers([A, B, C]);

// 0 < 1 < 2

A.setOrder(0); // A, B, C
A.setOrder(1); // B, A, C
A.setOrder(2); // B, C, A
A.setOrder(-1); // throw error: out of range
A.setOrder(3); // throw error: out of range

A.bringToFront(); // same as A.setOrder(<layers length>);
A.brintToBack(); // same as A.setOrder(0);
```

cc @csobier @alonsogarciapablo @IagoLast

IagoLast:
LGTM!

alonsogarciapablo:
What happens to the order of `B` and `C` when `A.setOrder(0)` is executed?

Why do you think `A.bringToFront()` and `A.bringToBack` will be useful?

Perhaps we can see how other libraries solve this problem. 

Also, something like `client.addLayer(myLayer, { at: 1 }` might be useful.

Jesus89:
> What happens to the order of B and C when A.setOrder(0) is executed?

Since the order is the index, B and C will have order 1 and 2 respectively.

> Why do you think A.bringToFront() and A.bringToBack will be useful?

I think these functions are more explicit and you don't need to remember the number of layers (for the `bringToFront` function).

> Also, something like client.addLayer(myLayer, { at: 1 }) might be useful.

May be we could rename this function to `client.addLayerAt(1, myLayer)` (or`client.addLayerAt(myLayer, 1)`)

Jesus89:
Hi @alonsogarciapablo @inigomedina @ivanmalagon

Are you OK with this API functions?

IagoLast:
- Since this functions require a new map instantiation I assume they are going to be async right? otherwise we could have concurrence issues.
- Are you sure about the layer `reordering`? how leaflet works?

Jesus89:
Layer ordering is a required feature. It is an async task so we need to provide also Promises. Leaflet has  zIndex and bringTo functions for the Leaflet LayerGroup.

IagoLast:
I mean `reordering`

Whats the expected behaviour? how leaflet (or any reference lib) work?
Example reordering layers?
``` javascript
// [l1, l2, l3] 
l1.setOrder(2);
// [l2, l3, l1]  --> l2 and l3 were reordered!
```

Example without reordering layers ?
``` javascript
// [l1, l2, l3] 
[...]

PR #2031, opened 2018-01-31
Define layers order

Jesus89:
Related to: https://github.com/CartoDB/cartodb.js/issues/1982

- [x] Add `moveLayer(layer, index)` function to Engine
- [x] Add `moveLayer(layer, index)` function to Client
- [x] Add `setOrder(index)` function to Layer
- [x] Add `bringToBack()` function to Layer
- [x] Add `bringToFront()` function to Layer
- [x] Add tests
- [x] Add examples
- [x] Little refactor for internal `triggerLayerFeatureEvent` function

### DIFF
--- a/src/api/v4/client.js
+++ b/src/api/v4/client.js
@@ -165,7 +165,7 @@ Client.prototype.removeLayers = function (layers) {
  * // Move layer order
  * client.moveLayer(layer1, 0)
  * .then(() => {
- *  console.log('Layers removed');
+ *  console.log('Layer moved');
  * })
  * .catch(cartoError => {
  *  console.error(cartoError.message);
--- a/src/api/v4/client.js
+++ b/src/api/v4/client.js
@@ -346,8 +346,6 @@ Client.prototype.getLeafletLayer = function () {
  * @api
  */
 Client.prototype.getGoogleMapsMapType = function (map) {
-  // NOTE: the map is required here because of wax.g.connector
-
   // Check if Google Maps is loaded
   _isGoogleMapsLoaded();
   if (!this._gmapsMapType) {
--- a/src/api/v4/client.js
+++ b/src/api/v4/client.js
@@ -159,6 +159,34 @@ Client.prototype.removeLayers = function (layers) {
 };
 
 /**
+ * Move layer order.
+ *
+ * @example
+ * // Move layer order
+ * client.moveLayer(layer1, 0)
+ * .then(() => {
+ *  console.log('Layers removed');
+ * })
+ * .catch(cartoError => {
+ *  console.error(cartoError.message);
+ * });
+ *
+ *
+ * @param {carto.layer.Base} - The layer to be moved
+ * @param {number} toIndex - Final index for the layer
+ *
+ * @fires error
+ * @fires success
+ *
+ * @returns {Promise} A promise that will be fulfilled when the layer is moved
+ * @api
+ */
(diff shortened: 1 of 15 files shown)


Jesus89:
@IagoLast @ivanmalagon @alonsogarciapablo @inigomedina

src/api/v4/client.js
```diff
@@ -158,6 +158,39 @@ Client.prototype.removeLayers = function (layers) {
...
 };
 
+/**
+ * Move layer order.
+ *
+ * @example
+ * // Move layer order
+ * client.moveLayer(layer1, 0)
+ * .then(() => {
+ *  console.log('Layers removed');
```
IagoLast: wrong log

src/api/v4/client.js
```diff
@@ -158,6 +158,39 @@ Client.prototype.removeLayers = function (layers) {
...
+ * })
+ * .catch(cartoError => {
+ *  console.error(cartoError.message);
```
IagoLast: does this method actually returns a cartoerror?

src/api/v4/layer/layer.js
```diff
@@ -376,6 +423,13 @@ Layer.prototype._createInternalModel = function (engine) {
 
 // Internal functions.
 
+Layer.prototype.$setClient = function (client) {
```
IagoLast: What is a client?

src/api/v4/client.js
Jesus89: :roll_eyes:

src/api/v4/client.js
Jesus89: Yes

src/api/v4/layer/layer.js
```diff
@@ -376,6 +423,13 @@ Layer.prototype._createInternalModel = function (engine) {
...
+  if (this._client) {
```
ivanmalagon: Shouldn't we check if `client` is the same instance as we do with `engine`? Just in case.

src/api/v4/native/leaflet-layer.js
```diff
@@ -68,7 +68,8 @@ var LeafletLayer = L.TileLayer.extend({
   },
 
   _onFeatureClick: function (internalEvent) {
-    triggerLayerFeatureEvent(Layer.events.FEATURE_CLICKED, internalEvent, this._layers);
+    var layer = this._layers.findById(internalEvent.layer.id);
+    triggerLayerFeatureEvent(Layer.events.FEATURE_CLICKED, internalEvent, layer);
```
ivanmalagon: We're mixing in this PR stuff related to interactivity along with the layer order, right? Correct me if I'm wrong.

ivanmalagon: If we're adding more stuff apart from layer ordering, please write that down in the issue.

src/api/v4/layer/layer.js
Jesus89: It should be always the same client for the engine. But just is case

ivanmalagon: Ok. If this comes from `engine` and the check is done at that level we're good to go then.

IagoLast:
how this affects interactivity?

Jesus89:
Interactivity is not affected by this PR.

ivanmalagon:
Passed acceptance.

🇯🇲

PR #2037, opened 2018-02-01
Set Leaflet zoomAnimationThreshold to 1000 (internal only)

dv343:
Set Leaflet `zoomAnimationThreshold` to 1000 (in the internal API) to avoid unnecessary gray background flashes.


Jesus89:
In order to be aligned with Builder, it would be nice to use leaflet 1.3.1 in Carto.js too (this is only required for the tests).

dv343:
> In order to be aligned with Builder, it would be nice to use leaflet 1.3.1 in Carto.js too (this is only required for the tests).

Done

Issue #2051, opened 2018-02-12
Use yarn


ivanmalagon:
We're moving all frontend projects from using `npm` to `yarn` (>=0.27.5 <1.0.0)

Task:
- Use yarn. ;)

Notes:
- Research if using `yarn.lock` is advisable here (since this is a library).

PR #2053, opened 2018-02-12
Use yarn

Jesus89:
Related to: https://github.com/CartoDB/cartodb.js/issues/2051

- [x] Use yarn@0.27.5 (this is the version used in our CI)
- [x] Add `yarn.lock` file
- [x] Update README commands
- [x] Update Travis configuration
- [x] Add `grunt-cli` as a devDependency
- [x] Use `grunt-eslint / eslint` dep instead of `eslint` dep
- [x] Fix `no-trailing-spaces`  lint errors


.eslintrc.json
```diff
@@ -8,6 +8,7 @@
   "globals": {
   },
   "rules": {
-    "no-useless-escape": "off"
+    "no-useless-escape": "off",
+    "no-trailing-spaces": "off"
```
ivanmalagon: What introduced this need?

.travis.yml
```diff
@@ -1,15 +1,12 @@
 sudo: false
-cache: false
+cache: yarn
 language: node_js
 node_js:
   - 6.9.2
-install:
-  - npm install
-before_script:
```
ivanmalagon: Why has `before_script` section disappeared?

Jesus89: This section is not used anymore. The `secrets.json` file is not used here and I have added `grunt-cli` to the devDependencies.

.eslintrc.json
Jesus89: It seems that this is failing in travis: https://travis-ci.org/CartoDB/cartodb.js/builds/340449842, but not in my local machine :S

PR #2068, opened 2018-03-07
Use empty string fallback instead of .toString method

rubenmoya:
Related to: https://github.com/CartoDB/support/issues/1368

Prevents getting `Uncaught Error: Cannot read property 'toString' of null` when the API returns `null` names in the response.

### DIFF
--- a/src/dataviews/category-dataview/search-model.js
+++ b/src/dataviews/category-dataview/search-model.js
@@ -81,13 +81,13 @@ module.exports = Model.extend({
 
   _parseData: function (categories) {
     var newData = [];
-    _.each(categories, function (d) {
-      if (!d.agg) {
+    _.each(categories, function (data) {
+      if (!data.agg) {
         newData.push({
           selected: false,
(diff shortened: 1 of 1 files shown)


src/dataviews/category-dataview/search-model.js
```diff
@@ -81,13 +81,13 @@ module.exports = Model.extend({
...
-    _.each(categories, function (d) {
-      if (!d.agg) {
+    _.each(categories, function (data) {
+      if (!data.agg) {
         newData.push({
           selected: false,
-          name: (d.category || d.name).toString(),
-          agg: d.agg,
-          value: d.value
+          name: data.category || data.name || '',
```
ivanmalagon: Perhaps it doesn't happen but...

The original code implies that `category` or `name` can be something apart from a string, a number for instance. Then, it converts it to string.

In your change, if `category` or `name` comes as numbers, they'll be passed along.

`(data.category || data.name || '').toString()` would do the trick, right?

PR #2069, opened 2018-03-13
Re-render legends when layers order is changed

rubenmoya:
Related to https://github.com/CartoDB/support/issues/1379

This PR adds a custom `layerMoved` event to refresh legends when `moveCartoDBLayer` is used.

### DIFF
--- a/src/geo/ui/legends/legends-view.js
+++ b/src/geo/ui/legends/legends-view.js
@@ -21,10 +21,8 @@ var LegendsView = View.extend({
   },
 
   _initBinds: function () {
-    this._layersCollection.on('add remove', this._onLayerAddedOrRemoved, this);
-    this.add_related_model(this._layersCollection);
-    this.settingsModel.on('change', this._onSettingsModelChanged, this);
-    this.add_related_model(this.settingsModel);
+    this.listenTo(this._layersCollection, 'add remove layerMoved', this._onLegendsChanged);
+    this.listenTo(this.settingsModel, 'change', this._onSettingsModelChanged);
   },
 
   render: function () {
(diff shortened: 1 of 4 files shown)


src/geo/ui/legends/legends-view.js
```diff
@@ -21,10 +21,8 @@ var LegendsView = View.extend({
   },
 
   _initBinds: function () {
-    this._layersCollection.on('add remove', this._onLayerAddedOrRemoved, this);
-    this.add_related_model(this._layersCollection);
-    this.settingsModel.on('change', this._onSettingsModelChanged, this);
-    this.add_related_model(this.settingsModel);
+    this.listenTo(this._layersCollection, 'add remove layerMoved', this._onLegendsChanged);
+    this.listenTo(this.settingsModel, 'change', this._onSettingsModelChanged);
```
jesusbotella: Don't we have to make `settingsModel` required? If it is not present, it might fail.

src/geo/ui/legends/legends-view.js
```diff
@@ -109,7 +107,7 @@ var LegendsView = View.extend({
     this.$(this._container()).append(layerLegendsView.render().$el);
   },
 
-  _onLayerAddedOrRemoved: function (layerModel) {
+  _onLegendsChanged: function (layerModel) {
```
jesusbotella: I think I'd put a name related with layers, like `onLayersChanged`, because this is the layer events' callback.

rubenmoya: wow, that was a fail 🤦‍♂️

PR #2072, opened 2018-03-19
Get layer aggregation from attributes instead of options

elenatorro:
[WIP]

Project: https://github.com/CartoDB/product/issues/30
Related Issue: https://github.com/CartoDB/cartodb/pull/13718

### DIFF
--- a/src/geo/map/cartodb-layer.js
+++ b/src/geo/map/cartodb-layer.js
@@ -42,7 +42,7 @@ var CartoDBLayer = LayerModelBase.extend({
     this.infowindow.fields.bind('reset add remove', this._reload, this);
     this.tooltip.fields.bind('reset add remove', this._reload, this);
 
-    this.aggregation = options.aggregation;
+    this.aggregation = attrs.aggregation;
 
     LayerModelBase.prototype.initialize.apply(this, arguments);
   },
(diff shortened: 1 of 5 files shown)


ivanmalagon:
Why are we doing this? I get that this in WIP but take into account that modifying this would break the v4 public API.

elenatorro:
Yes, I am aware of that. I think there is no way to send `options` instead of `attrs` from builder to cartodb.js, and that is why one of my suggestions was to add the default options directly on cartodb.js. However, since we are discussing to make it [transparent to cartodb.js](https://github.com/CartoDB/cartodb/pull/13718#issuecomment-374868382), let me see if it is possible to do it from builder by sending the aggregation through the `options` parameter.

src/geo/map/cartodb-layer.js
```diff
@@ -5,9 +5,16 @@ var InfowindowTemplate = require('./infowindow-template');
 var TooltipTemplate = require('./tooltip-template');
 var Legends = require('./legends/legends');
 var AnalysisModel = require('../../analysis/analysis-model');
+var Aggregation = require('../../api/v4/layer/aggregation');
```
ivanmalagon: Not sure about requiring objects of the public API (`api/v4`) from the internal library code.

In this case it's required only to get the placement constant but it's not needed to trigger a default aggregation.

Issue #2071, opened 2018-03-19
dataviews return forbidden error when using auth api beta with carto.js v4 beta


oriolbx:
When creating an API key with `maps` permission for a private dataset and using this api key for a CARTO.js v4 beta, the map is rendered correctly.

However, when adding a Formula dataview, CARTO.js v4 is returning me the next error message:

![screen shot 2018-03-19 at 09 54 49](https://user-images.githubusercontent.com/11177693/37586120-a7775da4-2b5b-11e8-81f6-c1d7d773fe55.png)

![screen shot 2018-03-19 at 09 55 14](https://user-images.githubusercontent.com/11177693/37586131-aa3d46ca-2b5b-11e8-8add-bc7e99d8e28f.png)

If I try to add SQL permissions to the dataset, I get the same error.

I'm attaching the code of the map here:

[auth_api_cartojsv4.html.zip](https://github.com/CartoDB/cartodb.js/files/1824298/auth_api_cartojsv4.html.zip)

ivanmalagon:
Tested against my account. It happens the same.

The error is returned by the dataview URL

Request to
```
http://0.ashbu.cartocdn.com/hacheka/api/v1/map/a58e0f800a28170c476dd1bed127469a:0/dataview/formula-c4?
```

Response
```
{
  "errors": [
    "permission denied for relation airbnb_listings"
  ],
  "errors_with_context": [
    {
      "type": "unknown",
      "message": "permission denied for relation airbnb_listings"
    }
  ]
}
```

dgaubert:
I'm missing something, where is the `api_key` in that request?

ivanmalagon:
It's normal that you don't see the api_key... because it's not there.

![que yo la vea](https://user-images.githubusercontent.com/1078228/37708990-075e4fee-2d09-11e8-8343-010e86e0b1c4.gif)

ivanmalagon:
CARTO.js already published to npm and our CDN.

`v4.0.0-beta.37`

#### CDN

[https://cartodb-libs.global.ssl.fastly.net/carto.js/v4.0.0-beta/carto.min.js](https://cartodb-libs.global.ssl.fastly.net/carto.js/v4.0.0-beta/carto.min.js)

#### NPM

[https://www.npmjs.com/package/carto.js](https://www.npmjs.com/package/carto.js)

jsanz:
[...]

PR #2073, opened 2018-03-21
Propagate API keys to dataviews in public API

ivanmalagon:
Related to https://github.com/CartoDB/cartodb.js/issues/2071
----

This PR propagates the credentials set in `carto.client` to dataviews. It fixes a bug, where the dataviews didn't had the API key and then they build URLs with no keys.

### Acceptance

*Public API*
- Create all type of widgets against a private dataset.
- Use an API key [1] with SQL select and MAPS permission for that dataset.
- Check that the widgets return data. No 403 must be received.
- Create all type of widgets against a public dataset.
- Use the `default_public` key.
- Check that widgets work.

*Builder*
- [x] Generally speaking, check that widgets behave.
- [x] In Builder, create several type of widgets. They must be created with no problem and filters should work.
- [x] In Category widgets, perform a search.
- [x] Publish the map
- [x] Interact with the widgets. They should work as always.
- [x] To test Auth tokens
    - [x] In a organization, share a private map with all type of widgets with a colleague.
    - [x] Seeing the map with the colleague profile, check in the URLs that the widgets are requested using auth tokens.
    - [x] Widgets should work as always.

### DIFF
--- a/test/spec/api/v4/dataview/category.spec.js
+++ b/test/spec/api/v4/dataview/category.spec.js
@@ -1,6 +1,7 @@
 var Backbone = require('backbone');
 var _ = require('underscore');
 var carto = require('../../../../../src/api/v4/index');
+var createEngine = require('../../../fixtures/engine.fixture.js');
 
 function createInternalModelMock () {
   var internalModelMock = {
@@ -35,16 +36,6 @@ function createInternalModelMock () {
 
 function createSourceMock () {
   return new carto.source.Dataset('foo');
-}
-
-function createEngineMock () {
-  var engine = {
-    name: 'Engine mock',
-    reload: function () {}
-  };
-  spyOn(engine, 'reload');
-
-  return engine;
 }
 
 describe('api/v4/dataview/category', function () {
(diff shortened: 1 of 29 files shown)


test/spec/api/v4/dataview/category.spec.js
```diff
@@ -169,7 +160,7 @@ describe('api/v4/dataview/category', function () {
       dataview.on('limitChanged', limitChangedSpy);
 
       expect(limitChangedSpy).not.toHaveBeenCalled();
-      dataview.$setEngine(createEngineMock());
+      dataview.$setEngine(createEngine());
```
Jesus89: Why not using MockFactory.createEngine? This is used more than 20 times. IMO we should use only one of these (mockFactory OR engine.fixture)

ivanmalagon: I was aiming the other way round, using the fixture for everything instead of getting a fat MockFactory.

I'll open a separated issue to unify this but I didn't want to clutter the PR even more.

Jesus89: Agree :+1:

ivanmalagon:
Mock unification being done at https://github.com/CartoDB/cartodb.js/pull/2076

PR #2074, opened 2018-03-22
Move template2x decision to leaflet rendering

javitonino:
https://github.com/CartoDB/cartodb/issues/13746

### DIFF
--- a/src/geo/leaflet/leaflet-tiled-layer-view.js
+++ b/src/geo/leaflet/leaflet-tiled-layer-view.js
@@ -14,6 +14,14 @@ var generateLeafletLayerOptions = function (layerModel) {
   };
 };
 
+var generateLeafletUrlTemplate = function (layerModel) {
+  if (window.devicePixelRatio > 1 && layerModel.get('urlTemplate2x')) {
+    return layerModel.get('urlTemplate2x');
+  } else {
+    return layerModel.get('urlTemplate')
+  }
+}
+
 var LeafletTiledLayerView = function (layerModel, opts) {
   LeafletLayerView.apply(this, arguments);
 
@@ -37,7 +45,7 @@ LeafletTiledLayerView.prototype = _.extend(
   LeafletLayerView.prototype,
   {
     _createLeafletLayer: function () {
-      return new L.TileLayer(this.model.get('urlTemplate'), generateLeafletLayerOptions(this.model));
+      return new L.TileLayer(generateLeafletUrlTemplate(this.model), generateLeafletLayerOptions(this.model));
     },
 
     _onAdd: function () {
@@ -50,7 +58,7 @@ LeafletTiledLayerView.prototype = _.extend(
 
     _modelUpdated: function () {
       L.Util.setOptions(this.leafletLayer, generateLeafletLayerOptions(this.model));
-      this.leafletLayer.setUrl(this.model.get('urlTemplate'));
+      this.leafletLayer.setUrl(generateLeafletUrlTemplate(this.model));
     }
   }
 );
--- a/src/vis/layers-factory.js
+++ b/src/vis/layers-factory.js
@@ -55,13 +55,9 @@ var LAYER_CONSTRUCTORS = {
   tiled: function (attrs, options) {
     checkProperties(attrs, ['urlTemplate']);
 
-    var layerUrl = LayersFactory.isHdpi() && attrs.urlTemplate2x
-      ? attrs.urlTemplate2x
-      : attrs.urlTemplate;
-
     attrs.urlTemplate = LayersFactory.isHttps()
-      ? transformToHTTPS(layerUrl)
-      : transformToHTTP(layerUrl);
+      ? transformToHTTPS(attrs.urlTemplate)
+      : transformToHTTP(attrs.urlTemplate);
 
     return new TileLayer(attrs, { engine: options.engine });
   },
@@ -152,10 +148,6 @@ LayersFactory.prototype.createLayer = function (type, attrs) {
 
 LayersFactory.isHttps = function () {
   return (window && window.location.protocol && window.location.protocol === 'https:') || false;
-};
-
-LayersFactory.isHdpi = function () {
-  return window.devicePixelRatio > 1;
 };
 
 module.exports = LayersFactory;


test/spec/geo/leaflet/leaflet-tiled-layer-view.spec.js
```diff
@@ -2,70 +2,132 @@ var Backbone = require('backbone');
...
-      subdomains: 'a',
-      opacity: 1,
-      attribution: 'CARTO',
-      errorTileUrl: 'http://matall.in',
-      maxZoom: 10,
-      minZoom: 0,
-      tms: false
+  describe('does not support high resolution screens', function () {
+    beforeEach(function () {
+      this.layerModel = new Backbone.Model({
```
IagoLast: I don't like to add stuff to the context because of the linter won't work instead I use local scoped variables, but is just a matter of personal taste

test/spec/geo/leaflet/leaflet-tiled-layer-view.spec.js
```diff
@@ -2,70 +2,132 @@ var Backbone = require('backbone');
...
-    this.layerModel = new Backbone.Model({
-      urlTemplate: 'http://whatever.com',
-      subdomains: 'a',
-      opacity: 1,
-      attribution: 'CARTO',
-      errorTileUrl: 'http://matall.in',
-      maxZoom: 10,
-      minZoom: 0,
-      tms: false
+  describe('does not support high resolution screens', function () {
```
IagoLast: (Another personal taste thing)
I think most of the tests in cartodb.js follow the [given, when, then](https://martinfowler.com/bliki/GivenWhenThen.html) rule so they are easier to follow.


```
(given a) 
leaflet-tiled-layer-view
    when high-resolution screens are not supported
        should have options set correctly
```

test/spec/geo/leaflet/leaflet-tiled-layer-view.spec.js
```diff
@@ -2,70 +2,132 @@ var Backbone = require('backbone');
...
-      expect(this.layerView.leafletLayer.options.attribution).toBe('carto');
-      expect(this.layerView.leafletLayer.options.errorTileUrl).toBe('http://matall.in');
-      expect(this.layerView.leafletLayer.options.maxZoom).toBe(12);
-      expect(this.layerView.leafletLayer.options.minZoom).toBe(1);
-      expect(this.layerView.leafletLayer.options.opacity).toBe(1);
-      expect(this.layerView.leafletLayer.options.tms).toBeTruthy();
-    });
+    describe('is high resolution screen', function () {
+      describe('._modelUpdated', function () {
+        it('to a model without high resolution enabled', function () {
```
IagoLast: [same here](https://github.com/CartoDB/cartodb.js/pull/2074/files#r176676515).

So I´d try to transform this:
> leaflet-tiled-layer-view: supports high resolution screens is high resolution screen ._modelUpdated to a model 
without high resolution enabled

Into

> leaflet-tiled-layer-view: when the layer-model supports high-resolution screens and the screen is high resolution ._modelUpdated should pick the high-resolution option as layer URL

src/geo/leaflet/leaflet-tiled-layer-view.js
```diff
@@ -49,12 +37,34 @@ LeafletTiledLayerView.prototype = _.extend(
...
+
+    _isHdpi: function () {
+      return window && window.devicePixelRatio && window.devicePixelRatio > 1;
     }
   }
 );
 
 LeafletTiledLayerView.prototype.constructor = LeafletTiledLayerView;
 
+var _generateLeafletLayerOptions = function (layerModel) {
```
IagoLast: again just a matter of personal taste:
I prefer to create a function using [function declarations](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function) instead [function expressions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/function)

nowadays it doesn't matter but a few years ago with function expressions you assign an anonymous function to a variable so the stack traces were all `anonymous` functions

src/geo/leaflet/leaflet-tiled-layer-view.js
```diff
@@ -37,7 +25,7 @@ LeafletTiledLayerView.prototype = _.extend(
   LeafletLayerView.prototype,
   {
     _createLeafletLayer: function () {
-      return new L.TileLayer(this.model.get('urlTemplate'), generateLeafletLayerOptions(this.model));
+      return new L.TileLayer(_getUrlTemplate(this._isHdpi(), this.model), _generateLeafletLayerOptions.call(this, this.model));
```
IagoLast: Are u sure the `.call(this,` is needed?

IagoLast: `_generateLeafletLayerOptions` is a pure function, it receives a layerModel and returns a options object, it wont use the js context at all right?

src/geo/leaflet/leaflet-tiled-layer-view.js
```diff
@@ -25,7 +25,7 @@ LeafletTiledLayerView.prototype = _.extend(
   LeafletLayerView.prototype,
   {
     _createLeafletLayer: function () {
-      return new L.TileLayer(_getUrlTemplate(this._isHdpi(), this.model), _generateLeafletLayerOptions.call(this, this.model));
+      return new L.TileLayer(_getUrlTemplate.call(this), _generateLeafletLayerOptions.call(this));
```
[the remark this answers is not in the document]
elenatorro: This a way to declare private functions without having them on the object's `prototype`. I do agree with the fact that therefore we can not consider it a "pure" function. However, as we have talked, I think that if we do not use this approach anywhere else in cartodb.js, it would be better not to use it in this what.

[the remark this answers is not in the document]
elenatorro: And about what you mentioned about how to declare functions. I do agree that functions must have a name so they are easy to debug, I changed that. However, I prefer to assign the function to a variable because then, they are not hoisted, and they can not be used before they have been declared, and I do like this approach.

PR #2075, opened 2018-03-22
Added example for the url server parameter

jsanz:
I have [this example](https://codepen.io/jsanz/pen/MVmNpm) running against a current carto instance, but was hesitant to use a real one since that carto instance can disappear anytime. I think it's worth adding that example to make clear that the URL needs the `/user/username`, even the console error is meaningful enough to be easy to fix, so not really strong on this suggestion :smile:


ivanmalagon:
I tweaked your suggestion, @jsanz . Can you take a look and tell me if you find it well explained?

jsanz:
Yep, looks good to me

PR #2087, opened 2018-04-13
Debounced fetch

Jesus89:
Related to: https://github.com/CartoDB/carto.js/issues/2140

This PR optimizes the server calls using a debouncer in the `engine.reload` function.

The current issue is that you can not do more than 3 requests without having a `"You are over platform's limits"` Windshaft error and this is a problem because you can not do things like the code below without having a server error.

```js
for (let dataview of dataviews) {
  dataview.setOperation(carto.operation.SUM);
}
```

Now with this PR, server calls are optimized with a debounce of 100 ms. Therefore the previous cases are fixed.

- [x] Add unit tests for Promises/Callbacks/Events
- [x] Add stack system for Promises/Callbacks
- [x] Refactor `engine.reload` function
- [ ] Builder integration and testing

#### Promises

- Resolve:
    - Single call: one resolve without params
    - Consecutive calls: **no batch request, only resolves the last call**
- Reject:
    - Single call: one reject with CartoError `error` param
    - Consecutive calls: **no batch request, only rejects the last call**
 
```js
client._engine.reload()
.then(() => console.log('Then'))
.catch(error => console.log('Catch', error.message));
```

#### Callbacks
- Success:
    - Single call: one success callback without params (`serverResponse` doesn't reach the callback)
    - Consecutive calls: **no batch request, only calls the last callback**
- Error:
    - Single call: one error callback with CartoError `error` param
    - Consecutive calls: **no batch request, only calls the last callback**

```js
client._engine.reload({
  success: () => console.log('Success'),
  error: error => console.log('Error', errors.message)
});
```

#### Events
- Started: no params
- Success: no params
- Error: CartoError `error` param
**no batch request**

```js
client._engine.on(Engine.Events.RELOAD_STARTED, () => console.log('Started'));
client._engine.on(Engine.Events.RELOAD_SUCCESS, () => console.log('Success'));
client._engine.on(Engine.Events.RELOAD_ERROR, error => console.log('Error', error.message));
client._engine.reload();
```

### Current state

**Adding batch server requests for consecutive `reload` calls**. Take into account the `reload` options:
- sourceId: latest
- forceFetch: OR-ed flags
- includeFilters: latest

#### Promises

- Resolve:
    - Single call: one resolve without params
    - Consecutive calls: **batch request, multi resolve without params**
- Reject:
    - Single call: one reject with CartoError `error` param
    - Consecutive calls: **batch request, multi reject with CartoError `error` param**
 
```js
client._engine.reload()
.then(() => console.log('Then 1'))
.catch(error => console.log('Catch 1', error.message));

client._engine.reload()
.then(() => console.log('Then 2'))
.catch(error => console.log('Catch 2', error.message));

client._engine.reload()
.then(() => console.log('Then 3'))
.catch(error => console.log('Catch 3', error.message));
```

#### Callbacks
- Success:
    - Single call: one success callback without params
    - Consecutive calls: **batch request, multi success callback without params**
- Error:
    - Single call: one error callback with CartoError `error` param
    - Consecutive calls: **batch request, multi error callback with CartoError `error` param**

```js
client._engine.reload({
  success: () => console.log('Success 1'),
  error: error => console.log('Error 1', error.message)
});

client._engine.reload({
  success: () => console.log('Success 2'),
  error: error => console.log('Error 2', error.message)
});

client._engine.reload({
  success: () => console.log('Success 3'),
  error: error => console.log('Error 3', error.message)
});
```

#### Events
- Started: no params
- Success: no params
- Error: CartoError `error` param
**batch request, handlers are called only once**

```js
client._engine.on(Engine.Events.RELOAD_STARTED, () => console.log('Started'));
client._engine.on(Engine.Events.RELOAD_SUCCESS, () => console.log('Success'));
client._engine.on(Engine.Events.RELOAD_ERROR, error => console.log('Error', error.message));
client._engine.reload();
client._engine.reload();
client._engine.reload();
```

NOTE: we need to be careful with this changes regarding this library and Builder.

NOTE2: there are some xit added because integration tests are coupled with unit tests and servers are not quite good today.

NOTE3: Please remove the acceptance example just before merging, when this PR is finished. It's the example from StackExchange that triggered this issue.

cc @ivanmalagon @inigomedina


src/engine.js
```diff
@@ -143,9 +149,24 @@ Engine.prototype.off = function (event, callback, context) {
  *
  */
 Engine.prototype.reload = function (options) {
+  var self = this;
```
ivanmalagon: I don't understand well this function.

1.- Why don't you use the `debounce` function from `underscore`?

2.- I don't grasp the use of Promise here since `_reload` already returns one.  Let me check it in more detail.

src/engine.js
```diff
@@ -143,33 +150,79 @@ Engine.prototype.off = function (event, callback, context) {
...
-    var errorCb = options.error;
-    options.success = function (serverResponse) {
-      successCb(serverResponse);
-      resolve();
-    };
-    options.error = function (errors) {
-      errorCb(errors);
-      var error = self._getSimpleWindshaftError(errors);
-      reject(error);
+    batchOptions = {
```
ivanmalagon: About this options:
- sourceId: ok. It should have the latest option, if any.
- forceFetch: this should be `true` if any stack call is `true`. But this condition is wrong, because if `batchOptions.forceFetch` is false and `options.forceFetch` is true, it returns false.
- includeFilters should have the last includeFilters. In this condition, it's using the last saved options if they exist.

Jesus89: Yes, I left that pending for discussion.
- sourceId: latest
- forceFetch: OR-ed flags
- includeFilters: I'm not sure about this options. By default is true, so if you want to disable it all should be false, true otherwise (?)

Jesus89:
I have added unit tests and updated the code with the new behaviour:
- sourceId: latest
- forceFetch: OR-ed flags
- includeFilters: latest

rjimenezda:
retest this please

Jesus89:
Hi @rjimenezda. Tests fail because of this https://github.com/CartoDB/carto.js/pull/2087/commits/1f398b04f843719523b41c1f3fb8d8a2d21e9743. There are bad practices there with unit tests (+e2e) so I think we should keep the comments until the server or the account for the tests is fine (or just update/fix the tests).

rjimenezda:
Reverted the specs then, all green 😂

rjimenezda:
I've tested on builder and found no issues, also for the public API version, the fix works beautifully.

Should we keep the acceptance example @ivanmalagon ?

test/spec/windshaft/error.mock.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  responseText: '{ "errors": ["Postgis Plugin: ERROR:  transform: couldnt project point (242 611 0): latitude or longitude exceeded limits."]}'
```
rubenmoya: Where is this used? I can't see it anywhere

rjimenezda: engine.spec.js:7

rubenmoya: the file was closed because it was too big for github 😂

Jesus89: Tests, tests, tests :green_heart:

ivanmalagon:
No. Remove the acceptance example. It's from the user that triggered the question on Stack Overflow.

PR #2088, opened 2018-04-16
Popups with just images don't look good on IE and Edge

VictorVelarde:
Related to https://github.com/CartoDB/cartodb/issues/13808

[WIP] This is the simplest solution: disable css clip-path when using IE or Edge, as it is not supported (https://caniuse.com/#feat=css-clip-path). 

The result is displayed below (notice the triangle, that was previously hidden):

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "cartodb.js",
-  "version": "4.0.0-beta.40",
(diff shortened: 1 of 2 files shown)


VictorVelarde:
This will be resolved at https://github.com/CartoDB/carto.js/pull/2105

PR #2111, opened 2018-05-07
Add new methods to cartoDB layer

IagoLast:
This methods help to solve https://github.com/CartoDB/support/issues/1399

### DIFF
--- a/src/geo/map/cartodb-layer.js
+++ b/src/geo/map/cartodb-layer.js
@@ -149,13 +149,13 @@ var CartoDBLayer = LayerModelBase.extend({
     LayerModelBase.prototype.remove.apply(this, arguments);
   },
 
-  $getTableName: function () {
-    if (this.attributes.source.attributes.options) {
-      return this.attributes.source.attributes.options.table_name;
+  getTableName: function () {
+    if (this.get('source').has('options')) {
+      return this.get('source').get('options').table_name;
     }
   },
 
-  $getApiKey: function () {
+  getApiKey: function () {
     return this._engine.getApiKey();
   }
 },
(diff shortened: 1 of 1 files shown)


src/geo/map/cartodb-layer.js
```diff
@@ -147,6 +147,16 @@ var CartoDBLayer = LayerModelBase.extend({
   remove: function () {
     this.getSource().unmarkAsSourceOf(this);
     LayerModelBase.prototype.remove.apply(this, arguments);
+  },
+
+  $getTableName: function () {
```
elenatorro: I would add some tests for `$getTableName` and `$getApiKey`. 

In addition, but this is a **personal** opinion, I would return `null` if there are not `this.attributes.source.attributes.options`. If the result of this function is assigned to a value, this value would be `undefined`, and I consider that is more correct to set it to `null` instead of `undefined`.

src/geo/map/cartodb-layer.js
```diff
@@ -147,6 +147,16 @@ var CartoDBLayer = LayerModelBase.extend({
...
+    if (this.attributes.source.attributes.options) {
```
ivanmalagon: Two things:

- Instead of accessing `this.attributes` use the `get` method (general tip for Backbone models). In this case, you can use the existing `getSource()`.
- This is an internal class and we are not following the `$` convention that we use in the public API.

src/geo/map/cartodb-layer.js
```diff
@@ -147,6 +147,16 @@ var CartoDBLayer = LayerModelBase.extend({
   remove: function () {
     this.getSource().unmarkAsSourceOf(this);
     LayerModelBase.prototype.remove.apply(this, arguments);
+  },
+
+  getTableName: function () {
+    if (this.get('source').has('options')) {
```
ivanmalagon: :+1:

PR #2116, opened 2018-05-11
Fix remove layers

ivanmalagon:
This PR fixes `removeLayers`.

We were removing elements from the same array that we were traversing. The side effect of that is that the internal array was not completely emptied so next time the user instantiate the map, it has layers that were supposed to be removed.

This PR adds also missing tests. Neither `client.removeLayers` nor `layers.remove` had any.

### DIFF
--- a/src/api/v4/client.js
+++ b/src/api/v4/client.js
@@ -163,7 +163,13 @@ Client.prototype.removeLayer = function (layer) {
  * @api
  */
 Client.prototype.removeLayers = function (layers) {
-  layers.forEach(this._removeLayer, this);
+  var clone = layers.slice(0);
+  var length = layers.length;
+  var i = 0;
+  for (i = 0; i < length; i++) {
+    this._removeLayer(clone[i]);
+  }
+
   return this._reload();
 };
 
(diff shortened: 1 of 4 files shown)


src/api/v4/client.js
```diff
@@ -163,7 +163,13 @@ Client.prototype.removeLayer = function (layer) {
  * @api
  */
 Client.prototype.removeLayers = function (layers) {
-  layers.forEach(this._removeLayer, this);
+  var clone = layers.slice(0);
```
jesusbotella: I have been thinking about what we did yesterday here and I think there are two slightly _easier_ solutions.

The one I think is better is to keep doing `.forEach` but using the cloned array to iterate, like:
```js
var layersToRemove = layers.slice(0);
layersToRemove.forEach(this._removeLayer, this);
```

And you will avoid the mutations on the original array passed on the parameters because of the cloned array. I think we tested this one but don't know why it didn't worked 🤔, I have just tested it on the browser and it works.

And the other one, avoids cloning the array, but it is riskier than the previous one:
```js
var length = layers.length;

for (var i = 0; i < length; i++) {
  this._removeLayer(layers[0]);
}
```

I prefer the first one because of obvious reasons. The thing that we did yesterday is not bad but I think it can be improved :)

ivanmalagon: Changing the code to your first suggestion. :+1:

jesusbotella: 💪

PR #2121, opened 2018-05-18
Fix legends paddings / margins

rubenmoya:
Related to: https://github.com/CartoDB/support/issues/1521

This PR changes how the paddings in legends work so perfect-scrollbar doesn't get activated when it shouldn't.

### DIFF
--- a/themes/scss/map/_overlays.scss
+++ b/themes/scss/map/_overlays.scss
@@ -228,7 +228,11 @@ $buttonColor: #636D72;
 $maxLegendContainerHeight: 300px;
 
 .CDB-LayerLegends {
-  margin-bottom: 12px;
+  margin-top: 0;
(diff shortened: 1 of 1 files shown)


themes/scss/map/_overlays.scss
```diff
@@ -228,7 +228,11 @@ $buttonColor: #636D72;
 $maxLegendContainerHeight: 300px;
 
 .CDB-LayerLegends {
-  margin-bottom: 12px;
+  margin-top: 0;
+
+  + .CDB-LayerLegends {
```
ivanmalagon: The same selector?

rubenmoya: Yes, so it's only applied when there is more than 1 legend

PR #2130, opened 2018-06-05
Fix Gmaps 3.32

ivanmalagon:
Related issue: https://github.com/CartoDB/carto.js/issues/2067

This PR upgrades the wax cache to record the number of calls to `getTile`.

Starting with Google Maps v3.32 (with a new renderer), Gmaps provides a smoother zoom. A side effect is that they call `getTile` several times for the same tile, releasing them as well. We have a cache in wax that removes the `` from the DOM, so we were removing a required tile.

This is the sequence during a pinch zoom in mobile:
```
a: start zooming.
b: getTile(32)
c: stop zooming.
d: getTile(32)
e: releaseTile(32) (the one required on b) -> we remove it from the DOM.
f: the tile disappear.
```

Now we track the number of calls to `getTile`, à la reference counting, and we only remove the `` from the DOM if there are no active calls.

### DIFF
--- a/vendor/wax.cartodb.js
+++ b/vendor/wax.cartodb.js
@@ -3329,20 +3329,27 @@ wax.g.connector = function(options) {
 // Get a tile element from a coordinate, zoom level, and an ownerDocument.
 wax.g.connector.prototype.getTile = function(coord, zoom, ownerDocument) {
     var key = zoom + '/' + coord.x + '/' + coord.y;
+    var img = null;
     if (!this.cache[key]) {
-      this.cache[key] = {
-        img: new Image(256, 256),
-        calls: 1
-      };
-      this.cache[key].img.src = this.getTileUrl(coord, zoom);
-      this.cache[key].img.setAttribute('gTileKey', key);
-      this.cache[key].img.setAttribute("style","opacity: "+this.opacity+"; filter: alpha(opacity="+(this.opacity*100)+");");
-      this.cache[key].img.onerror = function() { alert('mierdon'); img.style.display = 'none'; };
+      this.cache[key] = new Image(256, 256);
+      this.cache[key].src = this.getTileUrl(coord, zoom);
+      this.cache[key].setAttribute('gTileKey', key);
+      this.cache[key].setAttribute("style","opacity: "+this.opacity+"; filter: alpha(opacity="+(this.opacity*100)+");");
+      this.cache[key].onerror = function() { img.style.display = 'none'; };
+      this.cache[key].setAttribute('calls', '1');
     } else {
-      ++this.cache[key].calls;
+      try {
+        var calls = parseInt(this.cache[key].getAttribute('calls'), 10);
+        calls++;
+        this.cache[key].setAttribute('calls', calls.toString());
+      } catch (exc) {
+        // Swallow
+      }
+      
     }
+    img = this.cache[key];
 
-    return this.cache[key].img;
+    return img;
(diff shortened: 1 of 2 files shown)


vendor/wax.cartodb.js
```diff
@@ -3330,22 +3330,41 @@ wax.g.connector = function(options) {
...
-        this.cache[key].setAttribute("style","opacity: "+this.opacity+"; filter: alpha(opacity="+(this.opacity*100)+");");
-        this.cache[key].onerror = function() { img.style.display = 'none'; };
+      this.cache[key] = {
+        img: new Image(256, 256),
+        calls: 1
+      };
+      this.cache[key].img.src = this.getTileUrl(coord, zoom);
+      this.cache[key].img.setAttribute('gTileKey', key);
+      this.cache[key].img.setAttribute("style","opacity: "+this.opacity+"; filter: alpha(opacity="+(this.opacity*100)+");");
+      this.cache[key].img.onerror = function() { alert('mierdon'); img.style.display = 'none'; };
```
VictorVelarde: We don't probably want this alert anymore! :)

vendor/wax.cartodb.js
```diff
@@ -3329,23 +3329,58 @@ wax.g.connector = function(options) {
...
     var key = zoom + '/' + coord.x + '/' + coord.y;
+    var img = null;
     if (!this.cache[key]) {
-        var img = this.cache[key] = new Image(256, 256);
-        this.cache[key].src = this.getTileUrl(coord, zoom);
-        this.cache[key].setAttribute('gTileKey', key);
-        this.cache[key].setAttribute("style","opacity: "+this.opacity+"; filter: alpha(opacity="+(this.opacity*100)+");");
-        this.cache[key].onerror = function() { img.style.display = 'none'; };
+      this.cache[key] = new Image(256, 256);
+      img = this.cache[key];
```
VictorVelarde: As we have the reference inside the `img` variable maybe we could replace the next `this.cache[key]` ocurrences with it

vendor/wax.cartodb.js
```diff
@@ -3329,23 +3329,58 @@ wax.g.connector = function(options) {
...
+      this.cache[key].setAttribute("style","opacity: "+this.opacity+"; filter: alpha(opacity="+(this.opacity*100)+");");
+      this.cache[key].onerror = function() { img.style.display = 'none'; };
+      this.cache[key].setAttribute('calls', '1');
+    } else {
+      try {
+        var calls = parseInt(this.cache[key].getAttribute('calls'), 10);
+        calls++;
+        this.cache[key].setAttribute('calls', calls.toString());
+      } catch (exc) {
+        // Swallow
```
VictorVelarde: Why do we need this empty catch?

vendor/wax.cartodb.js
ivanmalagon: I prefer being explicit on things added to the cache but I get your point.
`img` is only used inside here for the `onerror` callback. If it wasn't for it I wouldn't create `var img` at all.

vendor/wax.cartodb.js
ivanmalagon: I'm protecting us from a bad parse. I'm not sure if `catch` is optional in Javascript but I prefer being explicit on `we are swallowing this exception because we don't want to break the code but if it fails we can go on`.

This pattern is clearly my old C# developer coming back, anyway.

PR #2137, opened 2018-06-12
Improving structure of contents

inigomedina:
As part of the second development cycle of the dev center, we are improving contents, creating new guides, especially for transversal themes, and updating old ones.


inigomedina:
I add @andrewbt as a reviewer, as he always helps a lot not only from the English side but especially from the user perspective. He is also working on the second development cycle of the dev center, and thus, is pretty aware of the changes we are implementing in terms of contents.

@ivanmalagon, we have more changes like these, and we want to publish them in an iterative manner. Could we merge the contents when the review is done, or do you want to lead that part?

ivanmalagon:
It depends. In this case I appreciate you asking for a review because you changed the structure. In the end it was only some added files and renaming, but it's good to take a look.

If you only change / update some text a review on our side it's not needed. If you make a change in file structure or it's something of some magnitude, ping us.

inigomedina:
Nice, man. 👍

docs/guides/01-overview.md
```diff
@@ -0,0 +1,173 @@
...
+
+```html
+      <!DOCTYPE html>
+<html>
+  <head>
+    <title>Single layer | CARTO</title>
+    <meta name="viewport" content="initial-scale=1.0">
+    <meta charset="utf-8">
+    <!-- Include Leaflet -->
+    <script src="https://unpkg.com/leaflet@1.3.1/dist/leaflet.js"></script>
```
andrewbt: @ivanmalagon do we have a cartocdn.com link for leaflet? Could we get one? Seems odd and off-brand to do these examples with unpkg...

docs/guides/01-overview.md
```diff
@@ -0,0 +1,173 @@
...
+
+### Loading the CARTO.js library
+
+To load the Maps JavaScript API, use a script tag like the one in the following example:
+
+```html
+<script src="https://libs.cartocdn.com/carto.js/v4.0.8/carto.min.js"></script>
+```
+
+The URL contained in the script tag is the location of a JavaScript file that loads all of the symbols and definitions you need for using the CARTO.js library. This script tag is required. We are using the minified version of the library.
```
andrewbt: @inigomedina do you mean "symbols" like ES6 symbols https://www.keithcirkel.co.uk/metaprogramming-in-es6-symbols/ ? CARTO.js is using those? Just want to make sure this is accurate - I hadn't heard the term before.

docs/guides/01-overview.md
```diff
@@ -0,0 +1,173 @@
...
+
+#### HTTPS or HTTP
+We think security on the web is pretty important, and recommend using HTTPS whenever possible. As part of our efforts to make the web more secure, we've made all of the CARTO components available over HTTPS. Using HTTPS encryption makes your site more secure, and more resistant to snooping or tampering.
+
+We recommend loading the CARTO.js library over HTTPS using the <script> tag provided above.
+
+#### Libraries
+
+When loading the CARTO.js library via the URL you may optionally load additional libraries through use of the libraries URL parameter. Libraries are modules of code that provide additional functionality to the main Maps JavaScript API but are not loaded unless you specifically request them.
```
andrewbt: @inigomedina This doesn't make sense to me. What is the "libraries URL parameter"? Do we have a more concrete example of an "additional library" a user would load?

docs/guides/02-get-api-key.md
```diff
@@ -0,0 +1,60 @@
...
+  - On the API key page, configure it giving a name and the APIs and Datasets you want to use. 
+
+For more information on using the CARTO Platform, see our [fundamentals]({{site.fundamental_docs}}/).
+
+
+#### Troubleshooting authorization issues
+
+If your API key is malformed or supplies an invalid username, the CARTO.js library returns an HTTP 403 (Forbidden) error.
+
+Alternatively, CARTO Enterprise customers have access to enterprise-level support through CARTO’s support representatives available at enterprise-support@carto.com.
```
andrewbt: What's the "alternative" here? Did we miss a paragraph?

docs/guides/02-get-api-key.md
```diff
@@ -0,0 +1,60 @@
...
+  - On the API key page, configure it giving a name and the APIs and Datasets you want to use. 
+
+For more information on using the CARTO Platform, see our [fundamentals]({{site.fundamental_docs}}/).
+
+
+#### Troubleshooting authorization issues
+
+If your API key is malformed or you supply an invalid username, the CARTO.js library returns an HTTP 403 (Forbidden) error.
+
+Alternatively, CARTO Enterprise customers have access to enterprise-level support through CARTO’s support representatives available at enterprise-support@carto.com.
```
andrewbt: @inigomedina what's "alternatively" here? Did we skip a paragraph?

docs/guides/03-quickstart.md
```diff
@@ -1,6 +1,8 @@
 ## Quickstart Guide
 
-Welcome to the CARTO.js documentation. CARTO.js is a JavaScript library that enables you to create custom location intelligence applications that leverage the power of the **[CARTO Engine](https://carto.com/pricing/engine/)** ecosystem.
+CARTO.js lets you create custom location intelligence applications that leverage the power of the **[CARTO Engine](https://carto.com/pricing/engine/)** ecosystem.
+
+This document details the CARTO.js v4. To update your v3 apps, please consult the [Upgrade Considerations]({{site.cartojs_docs}}/guides/upgrade-considerations/).
```
andrewbt: remove "the":
"This document details CARTO.js v4."

docs/guides/01-overview.md
ivanmalagon: We don't use Symbols. We can leave it as:

`...that loads all thee code you need for using...`

docs/guides/01-overview.md
ivanmalagon: We don't and I think we shouldn't. It implies maintenance in our part, whereas `unpkg` works using the `npm` registry and if something happens with `npm` or the `leaflet`package they'll fix it.

docs/guides/01-overview.md
ivanmalagon: I'm sure this is inspired by Google Maps API, right?

They have that functionality but we don't, so we should get rid of this paragraph.

docs/guides/01-overview.md
inigomedina: 👍

docs/guides/01-overview.md
inigomedina: 👍

docs/guides/01-overview.md
inigomedina: 👍

docs/guides/03-quickstart.md
inigomedina: 👍

docs/guides/01-overview.md
andrewbt: From https://unpkg.com/#/about

"unpkg is a free, best-effort service and cannot provide any uptime or support guarantees....The goal of unpkg is to provide a hassle-free CDN for npm package authors. It's also a great resource for people creating demos and instructional material. However, if you rely on it to serve files that are crucial to your business, you should probably pay for a host with well-supported infrastructure and uptime guarantees."

Maybe this is a larger issue that should be discussed elsewhere (what is the best place?), but since our customers do use CARTO and supporting libraries in business-crucial scenarios, we should think about either putting these files in a CDN we control, or at the very least documenting with a "disclaimer" of sorts the possible disadvantage of using unpkg in production. Customers will look to the Developer Center as the guide for what they should do in production, and unpkg seems to recommend against using themselves for that.

inigomedina: Good points, @andrewbt. I understand that we manage a different policy for CARTO resources (CDN policy) and external resources (it depends a lot on the specific resource). I agree with you that since users are expecting for ready-for-production content, we should add a "disclaimer" explaining that management of external resources.

I think we could add this to the `quickstart` guide as a general note. At the same time, the comment in the code `Include Leaflet` could be extended with the same disclaimer, so that users can read it in the same place the resources are being called. Even maybe a note on the `support` section, as an FAQ, would be worthy.

What do you think?

ivanmalagon: It's a larger issue, yes.

We can start the conversation in [docs repo](https://github.com/CartoDB/docs) since this will be reflected mostly in docs and engage @rochoa to provide the eagle-eye vision on this particular issue on where to host external libraries.

ivanmalagon:
I've merged this PR but it won't be published to docs until another PR with code gets merged.

We did a change to avoid differences between docs and code and we need to publish a tag in order to reflect changes in docs.

Since we're in active development of v4.1 this isn't a problem. We're going to merge code these days.

docs/guides/01-overview.md
rochoa: @andrewbt, what Leaflet versions do our customers request to use?

rochoa: @andrewbt, the _latest stable release_ term is something that changes over time, so we will need to start hosting every single release. Most serious developers will bundle, manage, and host their JS. As you say, unpkg is what Leaflet website recommends, and for demo purposes, it should be enough.

We have had our internal discussions about using unpkg for CARTO VL demos. However, as we have our `cartocdn` URLs, we opted to only use cartocdn to, precisely, avoid what you mentioned: "_unpkg is a free, best-effort service and cannot provide any uptime or support guarantees_".

Anyway, as @inigomedina proposes, something to consider, decide, fix, and document properly.

PR #2143, opened 2018-06-15
Fix margins on legends

jesusbotella:
This PR fixes margins and spacing on legends, specially on Bubble legend.

The problem was that the legend was taking more space than what it has assigned due to absolute positioning. So the margins were collapsed, appearing misaligned.

### Acceptance
See https://github.com/CartoDB/support/issues/1566

Related to https://github.com/CartoDB/support/issues/1566

### DIFF
--- a/themes/scss/map/_overlays.scss
+++ b/themes/scss/map/_overlays.scss
@@ -322,7 +322,7 @@ $maxLegendContainerHeight: 300px;
   right: 0;
   width: 100%;
   height: 100px;
-  margin: 16px 0 10px 0;
+  margin: 16px 0 10px;
 }
 
 .Bubble-circle {
--- a/themes/scss/map/_overlays.scss
+++ b/themes/scss/map/_overlays.scss
@@ -318,11 +318,11 @@ $maxLegendContainerHeight: 300px;
 
 .Bubble-container {
   position: relative;
-  top: 7px;
+  top: 0;
   right: 0;
   width: 100%;
   height: 100px;
-  margin: 10px 0;
+  margin: 16px 0 10px 0;
 }
 
 .Bubble-circle {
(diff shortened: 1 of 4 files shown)


themes/scss/map/_overlays.scss
```diff
@@ -230,7 +230,7 @@ $maxLegendContainerHeight: 300px;
 .CDB-LayerLegends {
   margin-top: 0;
 
-  + .CDB-LayerLegends {
+  &:not(:empty) + &:not(:empty) {
```
rubenmoya: brain-explode

jesusbotella: 💥

arianaescobar:
The margin error is fixed, but I found some other weird behaviors about the legends but unrelated to the margins. Should I open another issue or share them here?

jesusbotella:
@arianaescobar Can you open a new issue with the things you found, please? :)

arianaescobar:
Sure!

arianaescobar:
I'm just going to share one of those here because it might be related.

screen shot 2018-06-18 at 15 33 42

In the embed, if I collapse the legends, an extra margin appears on the bottom.

jesusbotella:
Yes, I have just checked the last change of that behaviour and [it was 2 months ago](https://github.com/CartoDB/cartodb/blob/47a880503bbaa2084f1e897fe1afb260d803005d/app/assets/stylesheets/deep-insights/themes/scss/map/_embed.scss#L97). So, I can do the change but seems like it was done on purpose 🤔

arianaescobar:
"🤔" indeed

jesusbotella:
I don't get your last comment 😂. Do you want me to change that margin?

ramiroaznar:
The RT has dealt with lots of bug-legends, look at support or cartodb repos and you will find the culprit.

arianaescobar:
@jesusbotella Yes, change it please 🙂 it should be the same margin on each side.

jesusbotella:
@arianaescobar I have just noticed that the margin change needs to be done in Builder, so I am gonna open a new issue for that. I'll link it in here once it is opened.

PR #2148, opened 2018-06-21
Update jQuery to version 3.3.1 and Mustache to version 2.3.0

rubenmoya:
Related to: https://github.com/CartoDB/development/issues/31

### Description

This PR updates `jQuery` and `Mustache` dependencies to avoid vulnerabilities. It also removes `vendor/mustache.js` since it wasn't used.


rubenmoya:
This is after running `update-internal-deps` in Builder:

screen shot 2018-07-04 at 09 30 09

We're going to have different versions of `jquery` and `mustache`, are we ok with it?

ivanmalagon:
I'm OK with it as long as Builder behaves and the acceptance passes.

The bundle size is huge anyway but how much do they add to the bundle size?

It's funny because there's no sign of mustache warning on Builder repo. I guess it uses a more modern version than carto.js with no vulnerabilities. Jquery warning is there so it's a matter of time to tackle it in Builder and have the same versions.

PR #2151, opened 2018-06-22
add hexagon aggregation example

oriolbx:
Add Leaflet and Gmaps example of point aggregation with hexagon aggregation

### DIFF
--- a/examples/examples.json
+++ b/examples/examples.json
@@ -176,6 +176,14 @@
             "leaflet": "public/misc/error-handling-leaflet.html",
             "gmaps": "public/misc/error-handling-gmaps.html"
           }
+        },
+        {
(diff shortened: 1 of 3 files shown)


oriolbx:
done @ivanmalagon :)

oriolbx:
Ok, this is not on the rest of examples, that means that we would need to modify all existing examples with comments for each block of code, right?

ivanmalagon:
I've merged this PR but it won't be published to docs until another PR with code gets merged.

We did a change to avoid differences between docs and code and we need to publish a tag in order to reflect changes in docs.

Since we're in active development of v4.1 this isn't a problem. We're going to merge code these days.

oriolbx:
Great! Thanks for the info! :)

PR #2152, opened 2018-06-26
Allow multiple carto clients using Google Maps

rubenmoya:
Related to: https://github.com/CartoDB/carto.js/issues/2132

### Description
This PR allows to use multiple carto clients when using Google Maps.

We had a few problems:
- We were assuming we only had one `GMapsCartoDBLayerGroupView`, and always using `setAt(0, layer)` which was overriding other layer groups added previously by other clients.
- We were adding `GoogleMapsMapType` to the `map.overlayMapTypes` array, but checking for `GMapsCartoDBLayerGroupView` instances, so we never found them.
- We were calling `map.overlayMapTypes.push` and telling our users to do the same, which lead to duplicated map types.
- We were calling `map.overlayMapTypes` (in the example) for each layer, which works with Leaflet, but not with google maps, we should document this somehow, we just have to call it once for each carto client / layer group.

I solved it by adding an internal `id` to both, `GoogleMapsMapType` and `GMapsCartoDBLayerGroupView`, and check that `id` instead of using `indexOf(this)`.

It seems to be working, although `map.overlayMapTypes` can contain two different things depending on where it's being used (builder or public api).

I've also checked Leaflet, but since it works differently we don't have that problem.


src/geo/gmaps/gmaps-cartodb-layer-group-view.js
```diff
@@ -285,11 +293,12 @@ _.extend(
     },
 
     _refreshView: function () {
-      var overlayIndex = this.gmapsMap.overlayMapTypes.getArray().indexOf(this);
+      var overlays = this.gmapsMap.overlayMapTypes;
+      var overlayIndex = this._getOverlayIndex();
+
       if (overlayIndex >= 0) {
-        this.gmapsMap.overlayMapTypes.removeAt(overlayIndex);
+        overlays.setAt(overlayIndex, overlays.getAt(overlayIndex));
```
ivanmalagon: This seems a bit strange since we're setting the overlay at the same exact place. I guess that `setAt` triggers something internally in Google Maps. If not, this has no much sense.

rubenmoya: It's basically a reload, before we were doing `remoteAt 0` -> `setAt 0`, but just doing `setAt` works fine.

rjimenezda:
I've tested the acceptance example as well as builder with gmaps and it seems like all's good in da hood

PR #2153, opened 2018-06-26
Fix interactivity when only 'cartodb_id' is selected

rubenmoya:
Related to: https://github.com/CartoDB/carto.js/issues/2089

### Description

This PR changes how we check if a layer is interactive or not.

The problem is that we were checking `this.getInteractiveColumnNames().length > 1` because we always add `cartodb_id` by default, so when adding it manually the check was false.


jesusbotella:
I've tested this PR in local and Builder and seems like it works like a charm.

I've tested that you can edit, select, and the popups are shown successfully. ✨

PR #2154, opened 2018-06-26
Fix removeDataview

rubenmoya:
Related to: https://github.com/CartoDB/carto.js/issues/2119

### Description
This PR calls `dataview.disable()` in `removeDataview`.

### Acceptance
- Use the example provided in the issue
- Open the map and check that moving the map calls the api
- Click in the _Remove widget_ button
- Check that moving the map doesn't call the api


jesusbotella:
I have made the acceptance for this PR, and works as expected. The `dataChanged` event is not fire after dataview is removed from the client.

I found that we were not returning a Promise when the dataview is not found in the client, so that's what I changed in the last commit, to follow the API specification.

Issue #2142, opened 2018-06-14
Histogram basic changes


ivanmalagon:
### Context
Right now, in CARTO.js v4.0 the only parameter available is `bins` to set the desired number of bins. There are cases where it'd be great to allow the user to control the range of the histogram: to get a more detailed histogram, for instance.
Internally, we already have a `start` and `end` parameters. We can expose them in the public API.

### Tasks
- Expose `start` and `end` parameters.
- Specs.
- Add documentation.
- Modify examples.

rubenmoya:
Added in version `4.0.18`

PR #2160, opened 2018-07-02
Add 'start' and 'end' values to histogram public API

rubenmoya:
Related to: #2142

### Description

This PR adds `start` and `end` values to `carto.dataview.Histogram`


src/api/v4/dataview/histogram/index.js
```diff
@@ -12,6 +12,8 @@ var parseHistogramData = require('./parse-data.js');
  * @param {string} column - The column name to get the data
  * @param {object} [options]
  * @param {number} [options.bins=10] - Number of bins to aggregate the data range into
+ * @param {number} [options.start] - Lower limit of the data range
```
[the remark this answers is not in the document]
rubenmoya: 👍 I'll make both of them optional and add more information

src/dataviews/histogram-dataview-model.js
```diff
@@ -71,6 +71,14 @@ module.exports = DataviewModelBase.extend({
     if (attrs && (attrs.min || attrs.max)) {
       this.filter && this.filter.setRange(this.get('min'), this.get('max'));
     }
+
+    if (this.get('customStart')) {
```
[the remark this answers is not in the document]
rubenmoya: I need a custom property, or a flag to skip [this](https://github.com/CartoDB/carto.js/blob/master/src/dataviews/histogram-dataview-model.js#L338), otherwise when fetching the totals the `start` and `end` values will be overridden. I tried using a boolean, but it was harder to work with it that work with `customStart` and `customEnd`.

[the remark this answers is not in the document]
rubenmoya: It should be completely transparent, since those properties are added through the public api, and without them, the code should work as it did before, but yeah, we should be careful.

rubenmoya:
I get what you say, I'll try to add `start` and `end` to the `histogram-data-model`!

rubenmoya:
While developing this we found a problem in Windshaft related to the `start` and `end` parameters, since in Builder we use the map instantiation to filter values we never realized this was wrong.

Blocking until https://github.com/CartoDB/Windshaft/issues/634 is fixed.

ivanmalagon:
As stated in https://github.com/CartoDB/Windshaft/issues/634#issuecomment-402684574 both `start` and `end` parameters should be defined for the histogram to behave.

[We're already doing it in the internal dataview](https://github.com/CartoDB/carto.js/blob/master/src/dataviews/histogram-dataview-model.js#L45) We only send them if both are defined.

Please modify the code to reflect that:
- Histogram validation.
- Remove `setStart` and `setEnd` and create a `setStartEnd` to modify both properties. Validate as well.
- Make sure that the datamodel only send both values.
- Reflect all this in the documentation.

src/api/v4/dataview/histogram/index.js
```diff
@@ -121,18 +161,31 @@ Histogram.prototype._validateBins = function (bins) {
   }
 };
 
+Histogram.prototype._validateStartEnd = function (start, end) {
+  const values = [start, end];
+  const notBothNumber = _.some(values, _.isFinite) && !_.every(values, _.isFinite);
```
jesusbotella: Shouldn't we check that both numbers are defined and valid, or both are null?

I think we shouldn't allow a single `null` value in this properties as per the comments I read.

src/dataviews/histogram-dataview-model.js
```diff
@@ -401,7 +413,10 @@ module.exports = DataviewModelBase.extend({
 
   _onTotalsError: function (model, error) {
     var parsedError = error && this._parseError(error);
-    this._triggerStatusError(parsedError);
+
+    if (parsedError && parsedError.message !== 'abort') {
```
jesusbotella: Why do we check if error is distinct from `abort`?

src/dataviews/histogram-dataview/histogram-data-model.js
```diff
@@ -138,6 +147,18 @@ module.exports = Model.extend({
     this.fetch();
   },
 
+  getStart: function () {
```
jesusbotella: Are `getStart` and `getEnd` methods useful?

Seems like we're not using them anywhere, and we always return the values the user set in histogram dataview instead of the _real_ values.

src/dataviews/histogram-dataview-model.js
rubenmoya: Because when you send the same request twice one of them gets aborted and it was showing an alert here: https://github.com/CartoDB/carto.js/blob/6c0c79f0987fd770bccf785380a667600ddbdde0/examples/public/dataviews/histogram.html#L76

We're doing that check in more places in the code, although with the debounce we merged I'm not sure if this will keep happening.

src/dataviews/histogram-dataview/histogram-data-model.js
rubenmoya: It's the same case as the bins, we return whatever the user sets 🤷‍♂️

jesusbotella: Yes, but you're doing that in https://github.com/CartoDB/carto.js/pull/2160/files/21e1d1e97a06083f70d83f0cf00227dbdee8a308#diff-26d0be025c5b2c0c647fb0f9b7f25e76R130 instead of here.

Here you are returning the values in the histogram data, rather than the ones that the user sets.

But my point is that I couldn't find any part of the code using these methods 🤷‍♂️

src/dataviews/histogram-dataview-model.js
jesusbotella: Can we check that it is not happening anymore to remove that if? 🙏

rubenmoya: Still happening

src/dataviews/histogram-dataview-model.js
```diff
@@ -366,7 +369,19 @@ module.exports = DataviewModelBase.extend({
...
+    const start = this.get('start');
+    const end = this.get('end');
+
+    const startEndChanged = helper.hasChangedSomeOf(['start', 'end'], this.changed);
+    const startEndValid = _.isFinite(start) && _.isFinite(end);
+    const hasDifferentValues = this._totals.get('start') !== start || this._totals.get('end') !== end;
+
+    if (startEndChanged && startEndValid && hasDifferentValues) {
+      this._totals.set({ start, end });
+      this._totals.refresh();
```
ivanmalagon: Why the `refresh` call?

I'm worried about the code below, where there are more `set` to `this._totals`. If they really don't have any conflict, it'd be better to encapsulate all the `startEnd` logic in a function. We're mixing here the new behaviour with the existing one. It's hard to read.

test/spec/api/v4/dataview/histogram.spec.js
```diff
@@ -223,6 +243,65 @@ describe('api/v4/dataview/histogram', function () {
...
+      dataview.on('endChanged', endChangedSpy);
+
+      expect(endChangedSpy).not.toHaveBeenCalled();
+      dataview.$setEngine(createEngine());
+      dataview.setStartEnd(20, 30);
+
+      expect(endChangedSpy).toHaveBeenCalledWith(30);
+    });
+  });
+
```
ivanmalagon: What happens when you set both values to `null`? It should reset the properties and make a call with no `start` and `end`. Can you add a spec for that?

examples/public/dataviews/histogram.html
```diff
@@ -71,9 +80,13 @@ <h1>Histogram widget</h1>
...
         const column = document.getElementById('column').value;
-        const bins = document.getElementById('bins').value;
+        const bins = parseInt(document.getElementById('bins').value);
+        const start = parseInt(document.getElementById('start').value) || null;
+        const end = parseInt(document.getElementById('end').value) || null;
+
         histogramDataview.setColumn(column);
-        histogramDataview.setBins(parseInt(bins));
+        histogramDataview.setStartEnd(start, end);
+        histogramDataview.setBins(bins);
```
ivanmalagon: Add an `alert` for validation errors. It'd be helpful for users.

src/dataviews/histogram-dataview-model.js
rubenmoya: The refresh call is basically to avoid handling the `change` event in the `data-model` because we would have to duplicate the same logic for the `onFieldsChanged` just to call refresh, so I thought it was easier just to do it here.

src/dataviews/histogram-dataview-model.js
```diff
@@ -208,7 +208,9 @@ module.exports = DataviewModelBase.extend({
   _onColumnChanged: function () {
     this._totals.set({
       column_type: this.get('column_type'),
-      column: this.get('column')
+      column: this.get('column'),
+      start: null,
```
jesusbotella: You didn't take into account the public API while doing these changes. You are going to mislead CARTO.js developers. 😂

We need to override `.setColumn` to reset start and end values in carto.dataviews.Histogram, because if we don't do that we are going to return the wrong values when invoking `.getStart` and `.getEnd`.

ivanmalagon: Good catch! 👀

ivanmalagon:
The error when switching columns in Builder is fixed. Anyway, we'll need to another round of Public API acceptance because of the Botella's comment.

src/dataviews/histogram-dataview-model.js
jesusbotella: I don't want developers to say `pero que ha pachao` 😂

rubenmoya: The problem is we don't have any way to know if the changes are because of the public API or the internal one.

If we reset `start` and `end`, we break the workflow in the public API, because users won't expect the start and end to change, if we don't reset it we break it in builder...

I think it makes sense the `start` and `end` values are reset since probably they won't have that much sense with the new column, maybe we can just document it in the public API?

jesusbotella: Yes, we do have a way 🤔 Do you want me to tell you about it, or you want to discover it by yourself? 🕵️

src/api/v4/dataview/histogram/index.js
```diff
@@ -90,6 +90,12 @@ Histogram.prototype.getData = function () {
   return null;
 };
 
+Histogram.prototype.setColumn = function (column) {
```
jesusbotella: ✨

PR #2170, opened 2018-07-11
Add 'visible' to layer options

rubenmoya:
Related to: https://github.com/CartoDB/carto.js/issues/2004

### Description
This PR adds `visible` to the layer options so people can create layers that are hidden by default.

It also improves the examples a little bit.

### DIFF
--- a/src/api/v4/layer/layer.js
+++ b/src/api/v4/layer/layer.js
@@ -30,6 +30,7 @@ var metadataParser = require('./metadata/parser');
  * @param {carto.style.CartoCSS} style - A CartoCSS object with the layer styling
  * @param {object} [options]
  * @param {Array<string>} [options.featureClickColumns=[]] - Columns that will be available for `featureClick` events
+ * @param {boolean} [options.visible=true] - A boolean value indicating the layer's visibility
  * @param {Array<string>} [options.featureOverColumns=[]] - Columns that will be available for `featureOver` events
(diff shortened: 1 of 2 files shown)


src/api/v4/layer/layer.js
```diff
@@ -63,26 +81,33 @@ var metadataParser = require('./metadata/parser');
...
-function Layer (source, style, options) {
+function Layer (source, style, options = {}) {
   Base.apply(this, arguments);
 
-  options = options || {};
-
   _checkSource(source);
   _checkStyle(style);
 
+  options = _.defaults(options, {
```
jesusbotella: Why did you change the default assignment to `_.defaults`? 🤔

jesusbotella: I know that the `visible` check _might not be_ as easy as the other ones, but you can take the chance to make a simple type check, so that you only allow boolean values.

Something like that:
```js
this._visible = _.isBoolean(options.visible) ? options.visible : true;
```

rubenmoya: It was easier to handle the default value of `visible` being it a boolean, since I can't do `falsy || true` and I didn't want to add a bunch of conditionals, what do you recommend?

jesusbotella: I recommend what I posted on top of your comment 😂

rubenmoya: you were faster pressing enter 😂

jesusbotella:
Working like a charm 🚀

PR #2173, opened 2018-07-11
Improve reload

Jesus89:
I realized that our requests system can be improved since we are still aborting prev requests, for example, when `reload` is called while waiting for a server response.

The following diagrams show the different states:

<br>

- **State 1**: No debounce, abort prev requests (Old)

<br>

reload-1

<br>

- **State 2**: Debounce: global stack, abort prev requests (Current)

<br>

reload-2

<br>

- **State 3**: Debounce: local stack, no abort prev requests (Proposed)

<br>

reload-3

<br>

I would like to know if this is OK for our Windshaft servers cc @rochoa

### DIFF
--- a/src/engine.js
+++ b/src/engine.js
@@ -178,50 +178,50 @@ Engine.prototype.reload = function (options) {
       this._performReload(batchOptions, stackCalls);
     }.bind(this);
     clearTimeout(this._timeout);
-    this._timeout = setTimeout(later, RELOAD_DEBOUNCE_TIME_IN_MILIS);
+    if (this.DISABLE_DEBOUNCE) {
+      later();
+    } else {
+      this._timeout = setTimeout(later, RELOAD_DEBOUNCE_TIME_IN_MILIS);
+    }
   }.bind(this));
 };
 
 Engine.prototype._performReload = function (options, stackCalls) {
-  return new Promise(function (resolve, reject) {
-    // Build Windshaft options callbacks
-    var windshaftOptions = this._buildWindshaftOptions(options,
-      // Windshaft success callback
-      function (serverResponse) {
-        this._onReloadSuccess(serverResponse, options.sourceId, options.forceFetch);
-        // Resolve stacked callbacks and promises
-        stackCalls.forEach(function (call) {
-          call.success && call.success();
-          call.resolve();
-        });
-        resolve();
-      }.bind(this),
-      // Windshaft error callback
-      function (errors) {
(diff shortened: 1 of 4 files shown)


alonsogarciapablo:
In State 2, I would expect the output to be Resolve 2, 3 if we were using my idea of Debounce (eg: [ Underscore's Debounce](https://underscorejs.org/#debounce)). If 1 and 2 are triggered within `RELOAD_DEBOUNCE_TIME_IN_MILIS` I would expect 1 to be aborted.

What's the meaning of the red cross in the server timeline?

Jesus89:
Black arrows are calls to reload function, blue arrows are requests to the server and red crosses represents a request abort, right.

In State 1 there was no debounce so requests were always aborted with the problems of limits and not resolving properly the promises.

In State 2 we implemented a debounce system but stacking the calls' information: Promise/resolve-reject and Callbacks/success-error. So when there are multiple calls of `reload` close enough they are batched and only one request is done, to avoid the max requests issue, and when this call is finished it resolves the promises and callbacks of 1 and 2.

```js
dataview1.setOperation('sum').then(...) // 1
dataview2.setOperation('sum').then(...) // 2
```

State 2 is the current state of the code, that was implemented to fix https://github.com/CartoDB/carto.js/issues/2140.

This PR adds an optimization for that. Since we are batching and stacking calls there should be no need to continue aborting calls because this only happens in a very specific case: when a request is made while another request is waiting for a response and this slows down the response time of the first request.

So, in State 3 there is the same debounce system, stacking calls to be properly resolved, but any call after a request won't block the first one but will create another request.

Hope it is clear enough :smile: :stuck_out_tongue:

Jesus89:
Deployed to be tested in staging: https://carto-staging.com/login
- user: improve-reload
- pass: improve-reload123

alonsogarciapablo:
Thanks @Jesus89. Clearer now.

> red crosses represents a request abort, right.

In State 2, I would expect request 1 to be aborted (eg: not be sent to the server) in favor of request 2 and request 3 to succeed (not sure why that one is being aborted).

Jesus89:
Yes, in this case, aborted means that the HTTP request is aborted, but the request is sent. Using debounce we merge debounced calls to do only one call to the server (the merge is only for the `forceFetch` option).

In State 2, request 3 aborts prev request 2 and that's what we want to change with this PR to have finally State 3.

alonsogarciapablo:
Thanks for your explanation @Jesus89. Understood. Makes total sense. 👍 👍 👍

PR #2177, opened 2018-07-13
Improve examples

rubenmoya:


jesusbotella:
I have tested all the examples and they're working like they were before, so 🎉

PR #2179, opened 2018-07-16
Fix popups/featureClick positions when scrolled

rjimenezda:
Related issue: https://github.com/CartoDB/support/issues/1659

So this has been broken since we introduced zera, but probably never seen it because we almost never have maps that are not full screen (iframes work just fine). I've attached an example that would be broken with the previous implementation if you scroll vertically or horizontally. Feel free to try it out on master.

I assume there's some difference between what we're getting with zera and what we used to get with wax. The thing is, there's apparently a Leaflet method that retrieves the containerPoint from the layerPoint, it seems to work perfectly.

### DIFF
--- a/examples/acceptance/displaced-popups.html
+++ b/examples/acceptance/displaced-popups.html
@@ -1,0 +1,109 @@
+<html>
+  <head>
+    <title>Refactoria</title>
+    <meta name="viewport" content="initial-scale=1.0">
+    <meta charset="utf-8">
+    <link href="https://fonts.googleapis.com/css?family=Montserrat:400,600,700|Open+Sans:300,400,600" rel="stylesheet">
+    
+    <!-- Include Leaflet -->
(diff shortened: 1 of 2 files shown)


rjimenezda:
@ivanmalagon @IagoLast :point_up:

ivanmalagon:
Regarding the code, nothing to say there. It makes sense to use the native Leaflet method for it.

I've tried with `master` distribution and with this fix and it works as expected. In that sense, :bandera:

Do you have any clue about what's different between what wax sent and zera? I guess not because wax is totally wiped out and it's not easy to compare both versions. Perhaps, @IagoLast has an idea of what's the root cause of this.

rjimenezda:
I have the old version around as well to compare.

It's interesting, wax fires two events (pointerup and click), and it looks like it doesn't use Leaflet's event API for that at all, which zera does. Instead, it adds listeners to the DOM element and calculates where it lands on the map, gets the grid tile, etc etc. Iago probably knows this better.

PR #2189, opened 2018-08-29
Filters improvements

jesusbotella:
This PR fixes a bug and introduces several improvements in filters described in https://github.com/CartoDB/carto.js/issues/2186.

### Allow empty filters combination
There were a problem when combining several filters, when having one of them empty. So, now the problem is gone. You can:

```js
const roomTypeFilter = new carto.filter.Category('room_type', { in: getSelectedRoomTypes() });
const emptyFilter = new carto.filter.Category('price', { });

const source = new carto.source.SQL('SELECT * FROM table');
source.addFilters([ roomTypeFilter, emptyFilter ]);
```

### Reset filters conditions
Filters now have `.resetFilters()` method to remove all filter conditions and leave it empty.
```js
const roomTypeFilter = new carto.filter.Category('room_type', { in: getSelectedRoomTypes() });
roomTypeFilter.resetFilters();
```

### Subqueries for IN and NOT IN filters in Category filter
From now on, we're going to allow subqueries as a parameter for IN and NOT IN conditions. You only need to set `query` parameter in the filter value, like this:
```js
const roomTypeFilter = new carto.filter.Category('room_type', { in: { query: 'SELECT distinct(name) FROM room_types'} });
```

### DIFF
--- a/src/api/v4/filter/base-sql.js
+++ b/src/api/v4/filter/base-sql.js
@@ -53,7 +53,7 @@ class SQLBase extends Base {
    * @param {object} filters - The object containing all the new filters to apply.
    */
   setFilters (filters) {
-    if (!filters || !_.isObject(filters) || _.isEmpty(filters)) {
+    if (!filters || !_.isObject(filters)) {
       return;
     }
 
@@ -61,6 +61,13 @@ class SQLBase extends Base {
     this._filters = filters;
 
     this.trigger('change:filters', filters);
+  }
+
+  /**
+   * Remove all conditions from current filter
+   */
+  resetFilters () {
+    this.setFilters({});
   }
 
   $getSQL () {
--- a/src/api/v4/filter/category.js
+++ b/src/api/v4/filter/category.js
@@ -47,6 +47,7 @@ const ALLOWED_FILTERS = Object.freeze(Object.keys(CATEGORY_COMPARISON_OPERATORS)
  * airbnbDataset.addFilter(roomTypeFilter);
  *
  * @class Category
+ * @extends carto.filter.SQLBase
  * @memberof carto.filter
  * @api
  */
@@ -93,7 +94,7 @@ class Category extends SQLBase {
    */
 
   /**
-   * Set filter conditions, overriding all the previous ones.
+   * Remove all conditions from current filter
    *
    * @memberof Category
    * @method resetFilters
--- a/src/api/v4/filter/category.js
+++ b/src/api/v4/filter/category.js
@@ -20,9 +20,9 @@ const ALLOWED_FILTERS = Object.freeze(Object.keys(CATEGORY_COMPARISON_OPERATORS)
  *
  * @param {string} column - The column which the filter will be performed against
  * @param {object} filters - The filters you want to apply to the table rows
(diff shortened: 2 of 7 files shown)


src/api/v4/filter/base-sql.js
```diff
@@ -63,6 +63,13 @@ class SQLBase extends Base {
     this.trigger('change:filters', filters);
   }
 
+  /**
```
Jesus89: Document this for the public API

src/api/v4/filter/base-sql.js
```diff
@@ -53,7 +53,7 @@ class SQLBase extends Base {
    * @param {object} filters - The object containing all the new filters to apply.
    */
   setFilters (filters) {
```
Jesus89: Document this for the public API

jesusbotella:
Yes, I put the parameter but I forgot to put the method's description  :(

Jesus89:
OK, just be careful with the text `Set filter conditions, overriding all the previous ones.` -> `*Reset* filter conditions, overriding all the previous ones.`.

I see that these methods (`set`, `setFilter` and `resetFilter`) are implemented in a common abstract class `BaseSQL` that is not exposed in the API. This is legit but breaks the relation between `filter.Category|Range` and `filter.Base` class. In this case, I would add **@extends carto.filter.Base** to `filter.Category` and `filter.Range` classes.

Finally, I noticed that `filter.FiltersCollection` has the message `This object should not be used directly.` so in the documentation, it should include the tag **@abstract**, as the `filter.Base` class, and also the **@extends carto.filter.Base** tag.

jesusbotella:
Shouldn't we add `@extends carto.filter.BaseSQL` instead of `@extends carto.filter.Base` in Category and Range filters? They extend from BaseSQL, not from Base.

Jesus89:
Yes, that should be the natural option, but I thought this class was hidden intentionally because the documentation for `set` and `setFilters` is different for each child `Category` and `Range`.

At this point, I'm not sure about exposing `BaseSQL` to the use or simply bypass it by using `Base`. I'll try to explain it.

```
Base -> SQLBase -> Category
                -> Range
     -> FiltersCollection -> AND
                          -> OR
     -> BoundingBox
     ...
```

The point of having an abstract class, like `FiltersCollection`, is to define a common set of functions, for example, for `AND` and `OR`. In this case, the documentation for those methods is in `FiltersCollection`.

In the case of `SQLBase`, the methods `set`, `setFilter` and `resetFilter` live in the class, but are documented in the children because of the attributes. So I think in this case exposing the `SQLBase` class does not add value to the final user and `Base` should be used instead.

Jesus89:
BUT, if you finally want to expose `SQLBase` class, it should contains tags **@abstract**, and **@api**.

These are the two options (I hope I'm not too fuzzy :stuck_out_tongue:)

jesusbotella:
Yes, that's fair enough.
I think it is better not to expose SQLBase class. I don't think anybody wants to implement that class on their own. :)

jesusbotella:
There's an error that needs to be fixed in Developer Center due to filters.in parameter types. But as soon as it's ready, this will be merged.

PR #2191, opened 2018-08-31
Review changelog

Jesus89:
Related to: https://github.com/CartoDB/carto.js/issues/2190

- Review CHANGELOG
- Review and rename CHANGELOG_INTERNAL (prev DEVLOG)
- Add CHANGELOG and LICENSE to the public package
- Update LICENSE date and format


Jesus89:
I'm not sure about this tool. We talk at the meeting that DEVLOG.md was not a good name for the document so we would change it to CHANGELOG_INTERNAL.md and only with the content of pre-releases because the autogenerated output was not useful enough and had a lot of noise.

Also, if you check this line you can see that this command is not being used because errors were found in the tool: https://github.com/CartoDB/carto.js/blob/master/package.json#L166. So I would remove it.

@jesusbotella @rjimenezda @VictorVelarde

jesusbotella:
I didn't know that there were a command to autogenerate the DEVLOG.md, but if it's broken there's no point to maintain that. We need to take a decision, whether we fix it or we remove it. I would remove it and do it manually.

I think the changes proposed are good, but we need to clearly state how to use the workflow.

VictorVelarde:
I would also go for the manual process

rjimenezda:
If it's not used and hasn't been in a long time, it's pointless imho.

The release procedure is already quite manual, so I don't mind, but I would add the check to github that either file has been changed in a PR.

Jesus89:
OK, so I'll clean this part.

> I would add the check to github that either file has been changed in a PR.

I have no permissions to add that test configuration in the repository so we must remember no change it for now when it is necessary.

jesusbotella:
We can adopt conventional changelog if necessary :)

PR #2192, opened 2018-09-05
Allow subqueries parameter in Range filters and more Category Filters

jesusbotella:
In a previous PR we introduced subquery parameter in NOT IN and IN category filters. I noticed that it can be passed in to another filters such as GT, GTE, LT, etc... So this PR add that features :)

### DIFF
--- a/src/api/v4/filter/category.js
+++ b/src/api/v4/filter/category.js
@@ -3,8 +3,8 @@ const SQLBase = require('./base-sql');
 const CATEGORY_COMPARISON_OPERATORS = {
   in: { parameters: [{ name: 'in', allowedTypes: ['Array', 'String', 'Object'] }] },
   notIn: { parameters: [{ name: 'notIn', allowedTypes: ['Array', 'String', 'Object'] }] },
-  eq: { parameters: [{ name: 'eq', allowedTypes: ['String', 'Number', 'Date'] }] },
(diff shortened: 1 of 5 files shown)


src/api/v4/filter/category.js
```diff
@@ -67,8 +69,8 @@ class Category extends SQLBase {
     return {
       in: '<% if (value) { %><%= column %> IN (<%= value.query || value %>)<% } else { %>true = false<% } %>',
       notIn: '<% if (value) { %><%= column %> NOT IN (<%= value.query || value %>)<% } %>',
-      eq: '<%= column %> = <%= value %>',
-      notEq: '<%= column %> != <%= value %>',
+      eq: '<%= column %> = <%= value.query ? "(" + value.query + ")" : value %>',
```
Jesus89: Why not `(<%= value.query || value %>)` as in `in`,`notIn`?

jesusbotella: I wouldn't like to include an unnecessary parenthesis in every query.
I can inject an `queryValueWithParenthesis` in the template and change it to `<%= queryValueWithParenthesis || value %>` if you think it's better :)

jesusbotella: Yep, I didn't think about it too much 😂

PR #2199, opened 2018-09-18
2197 allow ip as server url

### LINKED ISSUE #2197: CartoError: serverUrl is not a valid URL.
I've installed Carto On-Premises on a server inside closed network. When I create new carto.Client with serverUrl which is IP-Address like 192.168.0.XX, then uncaught error come out as above title.

IagoLast:
fixes #2197

### DIFF
--- a/src/api/v4/client.js
+++ b/src/api/v4/client.js
@@ -506,8 +506,9 @@ function _checkUsername (username) {
 }
 
 function _checkServerUrl (serverUrl, username) {
+  var ipRegex = /^(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)/;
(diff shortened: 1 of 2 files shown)


src/api/v4/client.js
```diff
@@ -506,8 +506,9 @@ function _checkUsername (username) {
 }
 
 function _checkServerUrl (serverUrl, username) {
+  var ipRegex = /^(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)/;
```
ivanmalagon: Add the protocol as well.

And I'm sure this regex can be simplified with a group and a quantifier.

test/spec/api/v4/client.spec.js
```diff
@@ -33,6 +33,46 @@ describe('api/v4/client', function () {
...
+      expect(function () {
+        client = new carto.Client({
+          apiKey: '84fdbd587e4a942510270a48e843b4c1baa11e18',
+          username: 'cartojs-test',
+          serverUrl: '192.168.0.1'
+        });
+      }).toThrow();
+    });
+
+    it('should throw when serverURL is an invalid ip adress with no /user/{username}', function () {
```
ivanmalagon: This spec and the next have the description switched.

PR #2210, opened 2018-12-05
Fix versioning and tags in guides

csubira:
Related to: https://github.com/CartoDB/carto.js/issues/2182

It includes fixes to two issues:

1. On the reference page. Since @Jesus89 already made a PR including the string replace task for the reference page, the `Loading the library` section would be as follows:

```
<!-- CDN: load the latest CARTO.js version -->
<script src="https://libs.cartocdn.com/carto.js/%CURRENT_VERSION%/carto.min.js"></script>
<!-- CDN: load a specific CARTO.js version-->
<script src="https://libs.cartocdn.com/carto.js/%VERSION%/carto.min.js"></script>
```

`%CURRENT_VERSION%` will be replaced by the newest tag (specified in https://github.com/CartoDB/developers/blob/master/_app/_data/technologies.yml#L153) 
`%VERSION%` will remain the same to show where the version should be specified to load a different one.

2. Replacing `4.0` with `v4` in guides

These changes will be live when a new tag is ready and updated in the developer center
[here](https://github.com/CartoDB/developers/blob/master/_app/_data/technologies.yml#L134)  and [here](https://github.com/CartoDB/developers/blob/master/grunt-tasks/config.js#L14 ) since the fetching of latest tag is not automatic.

### DIFF
--- a/docs/guides/01-overview.md
+++ b/docs/guides/01-overview.md
@@ -13,7 +13,7 @@ This conceptual documentation is designed to let you quickly start exploring and
 The easiest way to start learning about the CARTO.js library is to see a simple example. The following web page displays a map adding a layer over it.
 
 ```html
-      <!DOCTYPE html>
+<!DOCTYPE html>
 <html>
   <head>
     <title>Single layer | CARTO</title>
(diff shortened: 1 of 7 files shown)


csubira:
The outcome can be checked in [here](http://2182-update-cartojs-guides.developers.carto-staging.com/developers/carto-js/guides/quickstart/) (Dev center page staging link)

PR #2212, opened 2018-12-17
Fix wrong bound selection in histogram last bucket

csubira:
Fixes:
https://github.com/CartoDB/support/issues/1871

Useful comments found in an older closed issue:
https://github.com/CartoDB/carto.js/issues/926


ramiroaznar:
Let me know if you want me to check in Builder and give you the acceptance.

csubira:
Let's see first is @VictorVelarde can have a look and make a CR and then yes, it will be very helpful if you could test it in Builder!

VictorVelarde:
The fix LGTM @csubira, and it seems like there is no 'one duplicated bucket issue', so I think we can move on. I have just added 2 specs and made a little refactor to an if, and that's it.

Feel free to merge it

ramiroaznar:
@csobier @VictorVelarde next time ask me for checking the change on staging!

VictorVelarde:
I will Ramiro. Indeed I'll ask you for acceptance after merging another carto.js PR and releasing a new version

PR #2214, opened 2018-12-27
Add new tomtom geocoder

VictorVelarde:
tomtom-search

### Related to https://github.com/CartoDB/carto.js/issues/2213

Objectives:

- [x] Add new tomtom geocoder provider
- [x] Connect it as default to the 'search bar'
- [x] (optional...) improve the provider selection, to change more easily in the future (hopefully).

This process is very similar to what we did in: https://github.com/CartoDB/carto.js/pull/2020

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+### Fixed
+- Moved 'browserify-shim' to devDependencies, fixing a potential problem with npm shrinkwrap.
+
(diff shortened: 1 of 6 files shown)


VictorVelarde:
Instead of removing the previous geocoder provider, the new tomtom is selected by default for the Search. 

But, in the future, the Search could receive another service as parameter with `geocoderService` (if properly configured, currently just 'mapbox' or 'tomtom' are valid):

```
var search = new Search({
      model: this.map,
      mapView: this.mapView,
      geocoderService: 'mapbox', // <<<
      token: 'a_valid_mapbox_api_key'
    });
```

That parameter could be controlled by cartodb backend in the future

PR #2218, opened 2019-01-14
Fix filter "between" using dates

elenatorro:
Fixes Support issue https://github.com/CartoDB/support/issues/1898

# Context
_(copied from the issue)_

The range filter can use dates as follows, using `lte` & `gte` filters:

```js
let mi = new Date("2014-01-01T00:00:00.00Z");
let ma = new Date("2014-12-31T23:59:59.00Z");
let filtreData = new Carto.filter.Range('dat_data', { lte: ma, gte: mi});
```

However, when it uses the `between` filter like the next block of code:

```js
let mi = new Date("2014-01-01T00:00:00.00Z");
let ma = new Date("2014-12-31T23:59:59.00Z");
let filtroData = new Carto.filter.Range('dat_data', {between:{max: ma, min: mi}});
```

It gets the next error message: `message: "syntax error at or near "Jan""`:

screenshot 2019-01-04 at 10 40 02

File:

[between_test.html.zip](https://github.com/CartoDB/support/files/2726507/between_test.html.zip)

### Steps to Reproduce

1. Create a JS map using a dataset with date or timestamp fields
2. Apply the between filter
3. The map returns an error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+### Fixed
+- Filters: parse dates properly in the `between` filter.
+
 ## 4.1.9 - 2019-01-09
 
 ### Added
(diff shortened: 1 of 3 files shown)


elenatorro:
Thanks, @jesusbotella! I didn't thought about doing that check, it makes everything much simpler and makes much more sense ✨

Issue #2223, opened 2019-01-16
Enable the selection of the search box provider


ivanmalagon:
### Context

We've recently changed our search box provider to TomTom instead of Mapbox. See tickets:
- https://github.com/CartoDB/carto.js/issues/2213
- https://github.com/CartoDB/carto.js/pull/2214/files

The "problem" is that [even if the code seems to select the provider using an option](https://github.com/CartoDB/carto.js/blob/master/src/geo/ui/search/search.js#L70), in the reality we don't set that option in any place.

We need to address it somehow, to enable the configuration of the provider. This is a likely scenario in OnPremises, where a customer can have the chance of using their own credentials.

I don't a clear idea on how to tackle this in a simple way but one of the options is to provide a hierarchy on providers and then look into the credentials that backend sets. If TomTom key exists, use TomTom. If not, mapbox... and so on.

```
<script>
    var mapzenApiKey = 'mapzen-NYwhatever';
    var mapboxApiKey = 'pk.eyJ1IjoiY2FyWadusLadusFadus';
    var tomtomApiKey = 'supercalifragilisticoespialidoso';
  </script>
```

If you have any better idea please let's discuss it here.

cc @javitonino 

cc @ramiroaznar @jesusbotella  (Response Team)

jesusbotella:
User model has a property named `geocoder_provider` which sets the provider that you might want to use, which is not being used right now in Builder because it defaults to tomtom.

So why instead of relying on some window variables, we use that property to determine the geocoder and then look for the API Key on the window scope and if it's not there throw an error?.

javitonino:
We cannot use the `geocoder_provider` for that. That is the configuration of the user, which may be an unsupported provider for search (e.g: `heremaps`). It has to be some global config.

Carto.js wise, we probably want to expose a provider parameter. Builder-side we probably want to add an option to `app_config.yml` to select that (or use some order if multiple keys are provided).

jesusbotella:
Yes, it might be much easier using that injected property rather than trying to determine which provider to use from CARTO.js

jesusbotella:
I've noticed that we have some geocoder configuration options in `app_config.yml` already. So my proposal is, why don't we add a property to select which geocoder is the one to use and inject the configuration into JS?

The [app_config.yml](https://github.com/CartoDB/cartodb/blob/master/config/app_config.yml.sample#L201-L226) configuration is:
```yml
  geocoder:
    #force_batch: true
    #disable_cache: true
[...]

PR #2224, opened 2019-01-18
Enable search box geocoder provider selection

jesusbotella:
This PR allow us to select the geocoding provider for the search box overlay in Builder by injecting `geocodingConfiguration` variable containing the needed configuration into window scope.

Mirror PR from `cartodb`: https://github.com/CartoDB/cartodb/pull/14622
Related to #2223.

### DIFF
--- a/src/geo/ui/search/search.js
+++ b/src/geo/ui/search/search.js
@@ -15,6 +15,11 @@ var GEOCODERS = {
   'mapbox': mapboxGeocoder
 };
 
+var GEOCODERS_WINDOW_API_KEYS = {
+  'tomtom': 'tomtomApiKey',
+  'mapbox': 'mapboxApiKey'
+};
+
 /**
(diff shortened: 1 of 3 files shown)


src/geo/ui/search/search.js
```diff
@@ -67,11 +67,13 @@ var Search = View.extend({
   },
 
   _initializeGeocoder: function () {
-    this.geocoderService = this.options.geocoderService || DEFAULT_GEOCODER;
+    const injectedGeocoderConfig = this._getGeocodingInfoFromConfig();
```
ivanmalagon: The only thing that tingles my spidey sense is the mix of configurations.

I can get the provider from the injected config but the token can be undefined and then  I go directly for the window property... those combinations.

That won't break the code but if a configuration is wrong perhaps it makes it harder to debug.

PR #2235, opened 2019-08-02
Change attribution character and toggle widget based on container

alejandraarri:
This PR is to change the attribution widget character because it can be confusing as suggested in 
https://github.com/CartoDB/cartodb/issues/14914.

We are also making the widget to appear condensed when the map width is small enough to prevent if from taking too much space, and appear expanded in bigger map sizes so the attributions appear clearly visible.

### DIFF
--- a/CHANGELOG_INTERNAL.md
+++ b/CHANGELOG_INTERNAL.md
@@ -4,6 +4,9 @@ This file contains all the changes in the **internal bundle** (used by Builder).
 
 All the changes that affects the public bundle should be released as *patch* or *minor* and be included in the main Changelog.
 
+## 4.1.12-0 - 2019-08-02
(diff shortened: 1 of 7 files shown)


CHANGELOG_INTERNAL.md
```diff
@@ -4,6 +4,9 @@ This file contains all the changes in the **internal bundle** (used by Builder).
 
 All the changes that affects the public bundle should be released as *patch* or *minor* and be included in the main Changelog.
 
+## 4.1.12-0 - 2019-08-02
+- Change attribution character and toggle widget based on container [#2235](https://github.com/CartoDB/carto.js/pull/2235)
```
jesusbotella: I think it would be `4.1.11-1` as we haven't released any `4.1.11` public version

PR #2241, opened 2019-12-02
Replace uglify with terser to fix internal build

VictorVelarde:
Replace uglify, as internal build to get `cartodb.js` from `cartodb.uncompressed` was broken.


csubira:
Implemented in PR https://github.com/CartoDB/carto.js/pull/224

PR #2248, opened 2020-01-31
Return a new instance of source filters array

jesusbotella:
This PR modifies `getFilters()` method return value, given that there might be some incorrect behaviours when reusing that original array.

@oriolbx has found a bug that when trying to remove filters from the source with `source.getFilters()` didn't work properly because we were modifying the same array.


Mmoncadaisla:
@VictorVelarde this issue recently came up in Support and it'd be great if this fix could be applied, do you think it'd be feasible to implement this? cc @pablomoniz

PR #2249, opened 2020-02-27
Update basemaps attribution to match requirements from the data provider

skgsergio:
We need to update our basemaps attribution line. I think this is the place to change it in `carto.js` :)


skgsergio:
Closing as per https://github.com/CartoDB/cartodb-platform/issues/6575

PR #2261, opened 2021-10-20
Check attributions text before loading them

moicalcob:
## Resources
[Shortcut story](https://app.shortcut.com/cartoteam/story/178449/notify-about-the-limit-of-4096-named-maps-in-builder)

## Context
Datasets' attributions are not working fine when loading a map. Sometimes the attributions are duplicated, or an empty space is being shown.

## Changes
I've checked that the attributions array doesn't contain duplicated elements. I've also checked with a regular expression that the string that is going to be rendered is clean.

### DIFF
--- a/src/geo/ui/attribution/attribution-view.js
+++ b/src/geo/ui/attribution/attribution-view.js
@@ -31,12 +31,11 @@ module.exports = View.extend({
   },
 
   render: function () {
+    var emptyStringRegExp = new RegExp(/[A-z]/g);
     var attributionsSet = [...new Set(this.map.get('attribution'))];
     var attributions = _.compact(attributionsSet)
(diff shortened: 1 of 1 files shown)


VictorVelarde (approved):
LGTM. I'll try to make ~a release~ a tag so it can be used in prod from Builder (it needs another PR in cartoDB, just to update internal version link)

VictorVelarde:
v4.2.2-2 tag created

VictorVelarde:
@moicalcob, now you should be able to create a PR in cartodb to use this tag. Change would be something like:

image

Remember also to bump assets in package.json
👋🏻

moicalcob:
That's nice! Thanks so much!! 😄

PR #2262, opened 2021-11-02
Replace git:// with https:// in package.json files

nhorvath:
Github is removing git:// access because it's insecure.
https://github.blog/2021-09-01-improving-git-protocol-security-github/

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6465,8 +6465,8 @@
       }
     },
     "grunt-contrib-watch": {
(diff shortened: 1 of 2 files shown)


Radik24:
Hi, can we merge it? it seems to be big problem right now. Thanks

padawannn:
> Hi, can we merge it? it seems to be big problem right now. Thanks

https://github.com/CartoDB/carto.js/pull/2268

In the 4.2.2 version should be resolved

